PR #3, opened 2022-02-07
Python script composable environments

kgerheiser:
@climbfuji @mark-a-potts @rhoneyager

Thoughts on this direction? (Warning code not fully tested). In @climbfuji's instructions we were running `rsync` to copy the env files, but this formalizes that into a Python script.

My idea is to have a Python setup script that creates the environment for you by copying the necessary config files into a directory that you then run Spack on. It takes the pre-configured site name (hera, orion, cheyenne, etc), an app name (ufs, jedi), and a name for the directory. This is similar to hpc-stack where you specify a config and a stack.

This keep everything pure Spack once the env is created, and just setups the initial directory.

See the new layout here: https://github.com/NOAA-EMC/spack-stack/tree/python-scipt-composable-env

# Some changes
* Re-named `envs` to `configs`
* Changed `<site>/config.yaml` to `site.yaml` to distinguish it from the built-in `config.yaml`.
* Created `setup.sh` that is sourced at startup, which sources the Spack `setup-env.sh` and sets `SPACK_STACK_DIR`.
* Created `create-env.py` to initialize a directory with a pre-configured Spack environment

For example:

```
source setup.sh
./create-env.py --site hera --app ufs --name ufs-env
```

```
spack-stack % ./create-env.py -h
usage: create-env.py [-h] [--site SITE] [--app APP] [--name NAME]

Create a pre-configured Spack env folder.

optional arguments:
  -h, --help   show this help message and exit
  --site SITE  Pre-configured platform to setup env for (hera, orion,
               cheyenne, etc).
  --app APP    Which app env to build (ufs, jedi).
  --name NAME  Optional name for env dir. Defaults to site name.
```

It copies "common configs" (located in common/", config.yaml, packages.yaml, modules.yaml), and site.yaml into `envs/<name>`.

Then, you go into the directory and run `spack env create -d .` and everything is in place.

The repos.yaml is referenced by the absolute path of `SPACK_STACK_DIR` to keep a consistent package build (i.e. a new package is added), but this could adjusted.

https://github.com/NOAA-EMC/spack-stack/blob/0f55fcc8c2e26b1f19ebd0a26ea8682e0bda2230/configs/apps/jedi/spack.yaml#L8


configs/apps/jedi/spack.yaml
```diff
@@ -2,28 +2,13 @@ spack:
...
-      root: install
-      projections:
-        all: ${PACKAGE}-${VERSION}
-    module_roots:
-      lmod: install/modulefiles
-
-  # Probably don't need that if we install envs
-  definitions:
-    - packages:
+ specs:
```
climbfuji: Do we need those specs at all?

configs/apps/ufs/spack.yaml
```diff
@@ -0,0 +1,63 @@
...
+  concretization: separately
+  view: false
+  include:
+  - config.yaml
+  - packages.yaml
+  - modules.yaml
+  - ${SPACK_STACK_DIR}/configs/repos/repos.yaml
+  - site.yaml
+
+ specs:
```
climbfuji: Same here.

create-env.py
```diff
@@ -0,0 +1,41 @@
...
+
+import argparse
+import os
+import sys
+import shutil
+
+parser = argparse.ArgumentParser(description='Create a pre-configured Spack env folder.')
+parser.add_argument('--site', type=str, help = 'Pre-configured platform to setup env for (hera, orion, cheyenne, etc).')
+parser.add_argument('--app', type=str, help = 'Which app env to build (ufs, jedi).')
+parser.add_argument('--name', type=str, help = 'Optional name for env dir. Defaults to site name.')
```
climbfuji: The default is not configured yet. 

I am wondering if it should be a combination of app and site. That would have the disadvantage that a lot of code would be duplicated, but the advantage that potential mismatches between package specs wouldn't be a problem.

We don't have that problem right now, it's probably best to just use the site name as default as you suggested.

climbfuji (changes requested):
The `create_env.py` script is a great idea.

Not sure about `setup.py`, it replaces to basic commands that would help the user to understand what is happening with one script to run - but I am open to keep it if we think it makes sense.

Can you please update the `.gitmodules` file and submodule pointer so that it points to the EMC spack fork? Thanks!

configs/apps/ufs/spack.yaml
kgerheiser: This could be `ufs-env` (from the `package.py`) to encapsulate the spec, and you could remove all these specs. 

Though, my preference is to specify the specs in a `spack.yaml`. I think the `*-env` packages are kind of a hacky solution to what environments offer.

```
    depends_on('eccodes', type=('build', 'run'))
     depends_on('py-setuptools', type=('build', 'run'))
     depends_on('py-cython', type=('build', 'run'))
     depends_on('py-numpy', type=('build', 'run'))
     depends_on('py-proj', type=('run'))
```
vs. 

```
specs:
- eccodes
- py-setuptools
- py-numpy
- py-proj
```

climbfuji: So this means that each env is specific to an application? By doing `spack install` you would get whatever is needed for the app you requested, but you wouldn't be able install the missing packages for another app in a convenient way?

create-env.py
kgerheiser: Yeah, I had similar thoughts. I think there's room for some specific configuration info in the site name like `hera.ufs.intel`.

configs/apps/ufs/spack.yaml
kgerheiser: Also good points. Each env would be specific to an application, yes. Though, I imagined some sort of combined env that contains 'everything' (kind of like `stack_noaa` from hpc-stack) along with trimmed down envs tailored to specific apps.

kgerheiser:
> Not sure about setup.py, it replaces to basic commands that would help the user to understand what is happening with one script to run - but I am open to keep it if we think it makes sense.


I can take or leave `setup.sh` it seemed convenient to not have to type `source spack/share/spack/setup-env.sh` each time. And I wanted an env var to point to the top-level dir.

> Can you please update the .gitmodules file and submodule pointer so that it points to the EMC spack fork? Thanks!

It already points to the EMC fork (https://github.com/NOAA-EMC/spack/tree/86465d55244585512419221f77f136f0ca72b28a)

climbfuji:
> > Not sure about setup.py, it replaces to basic commands that would help the user to understand what is happening with one script to run - but I am open to keep it if we think it makes sense.
> 
> I can take or leave `setup.sh` it seemed convenient to not have to type `source spack/share/spack/setup-env.sh` each time. And I wanted an env var to point to the top-level dir.

Yes, we need that env var. Let's keep the script for now.

> 
> > Can you please update the .gitmodules file and submodule pointer so that it points to the EMC spack fork? Thanks!
> 
> It already points to the EMC fork (https://github.com/NOAA-EMC/spack/tree/86465d55244585512419221f77f136f0ca72b28a)

Perfect, thanks!

climbfuji:
I need to add more packages as I work on the various JEDI applications. Therefore it would be good if we could merge the PR soon so that I don't need to make changes based on my old branch and create merge conflicts. I'll test your PR in a toy environment on my Mac now.

kgerheiser:
Hmm, the gcc build on Ubuntu is failing because of a duplicate install prefix for HDF5. I don't know why it's building two.

climbfuji:
> Hmm, the gcc build on Ubuntu is failing because of a duplicate install prefix for HDF5. I don't know why it's building two.

If you can replicate this on a local machine/AWS instance, then you can look at the `spack.lock` file. It should have both hashes for `hdf5` listed somewhere as dependency.

climbfuji:
I need to come back to the bigger picture, replacing the environment spack packages with separate spack environments. The benefits of using `ufs-weather-model-env`, `fv3-bundle-env` etc are (1) that you can install one and then add the missing packages by installing the other on top of it in the same environment, and (2) that you can create a module file with autoloading of dependencies and then simply do `module load ufs-weather-model-env/1.2.3` to get a tagged combination of libraries that are known to work with specific versions of the ufs-weather-model. I would like to retain those two benefits. If that can be done with the solution presented here, then that's ok, but I don't see how.

kgerheiser:
That's fair enough. I'll replace the individual spec list with the `-env` packages in `spack.yaml`.

climbfuji:
> That's fair enough. I'll replace the individual spec list with the `-env` packages in `spack.yaml`.

Thanks, Kyle. Do you want me to try building on Ubuntu to see if I can figure out the hdf5 version conflict?

kgerheiser:
Sure, try it out. On my Mac (Intel, gcc-9) I'm getting the same error with the `lz4` package.

climbfuji:
> Sure, try it out. On my Mac (Intel, gcc-9) I'm getting the same error with the `lz4` package.

Did you run `spack concretize` on the entire set of packages? The output should contain multiple `lz4` with the same version number but different specs.

kgerheiser:
Check out the `concretize` and `install` steps here. I only see one version of HDF5 in the concretization. All the variants look the same to me, and are constrained by `packages.yaml`, so I don't see how there can be two.

https://github.com/NOAA-EMC/spack-stack/runs/5104711119?check_suite_focus=true

climbfuji:
The concretizer must be getting confused. Check out the attached image ...

Screen Shot 2022-02-08 at 10 23 09 AM

kgerheiser:
How did you generate spec1 and spec2?

climbfuji:
I didn’t, it’s from the travis ci log you pointed me to.

kgerheiser:
Thoughts on merging this @climbfuji? I have changed the environments to use the bundle packages, and the test build should finish up in a bit.

configs/apps/fv3/spack.yaml
```diff
@@ -0,0 +1,12 @@
...
+  view: false
+  include:
+  - config.yaml
+  - packages.yaml
+  - modules.yaml
+  - site.yaml
+  - ${SPACK_STACK_DIR}/configs/repos/repos.yaml
+
+  specs:
+  - fv3-bundle-env
```
climbfuji, line 12: There is a bit of confusion around the meaning of these environments. `fv3-bundle-env` is an application that uses JEDI with the FV3 dycore (and for now also with NASA GEOS). `ufs-bundle-env` is an application that uses JEDI with the entire UFS, not just the dycore.

kgerheiser: Yes, I didn't really understand what these bundles were. What do you think a good name would be? Just name it `fv3-bundle` and `ufs-bundle`? Or `jedi-fv3-bundle`...?

climbfuji: I think the names of the bundles are ok, but we should call the apps `fv3-jedi` or `jedi-fv3` etc. Let's not worry about this now, we can rename them in a follow-up PR. There are many more of these: JEDI with soca, um, mpas, ...

configs/apps/ufs/spack.yaml
```diff
@@ -0,0 +1,12 @@
...
+  view: false
+  include:
+  - config.yaml
+  - packages.yaml
+  - modules.yaml
+  - site.yaml
+  - ${SPACK_STACK_DIR}/configs/repos/repos.yaml
+
+  specs:
+  - ufs-bundle-env
```
climbfuji, line 12: This will install a lot more than what you need to build the ufs (eckit, fckit, atlas, ecbuild, ...) but ok for now, we can fine tune later.

configs/common/packages.yaml
```diff
@@ -139,3 +139,7 @@
       version: [4.7.0]
     udunits:
       version: [2.2.28]
+    openmpi:
```
climbfuji, line 142: So my hope was that we could entirely avoid building MPI libraries as part of the spack stack and always require the following three things to be present (1) compiler (2) MPI library (3) Python interpreter. But if that's not a reasonable goal, I can be convinced.

climbfuji (approved):
Let's merge this now and address questions and issues as they come up in this early stage of development.

PR #13, opened 2022-02-09
Initial readme

kgerheiser:
Create initial README

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,41 @@
+# spack-stack
+
+Spack-stack makes it easy to build JEDI and UFS application environments using Spack, and comes pre-configured for many systems.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -0,0 +1,40 @@
...
+
+# Creates a pre-configured Spack environment in envs/<app>.<site>
+# Copies site-specific, application-specific, and common config files into the environment file
+./create-env.py --site hera --app jedi-fv3
+
+cd envs/jedi-fv3
+
+# Create a Spack environment in this directory and activate it
+spack env create -d .
+spack env activate .
```
climbfuji: There is an option `-p` that can be used the modify the prompt with the env name (in this case the name of the directory `jedi-fv3`. But currently it messes up the prompt on some systems.

climbfuji: Do you want to mention it here?

kgerheiser: I didn't know about that, but sure.

PR #19, opened 2022-02-09
Various minor updates to packages, site configs, Python scripts, README.md, ...

climbfuji:
A bunch of minor updates:
- several updates to package specs
- - better use of base-env
- updates to macOS site config
- let Python scripts output what they are doing
- update spack submodule pointer
- update top-level `README.md`
- python bindings for buffer

Installed on my macos with clang+gfortran in the jedi-ufs-bundle-env.

Requires https://github.com/NOAA-EMC/spack/pull/9.


README.md
```diff
@@ -18,23 +18,22 @@ cd spack-stack
...
+./create-env.py -h
 
 # Creates a pre-configured Spack environment in envs/<app>.<site>
 # Copies site-specific, application-specific, and common config files into the environment directory
 ./create-env.py --site hera --app jedi-fv3 --name jedi-fv3.hera
 
-cd envs/jedi-fv3
-
-# Create a Spack environment in this directory and activate it
-spack env create -d .
```
climbfuji: This step is not needed.

configs/sites/macos/site.yaml
```diff
@@ -127,8 +115,8 @@ packages:
     version: [release-stable]
     variants: +64bit +enable_quad_precision +gfs_phys ~openmp +pic
 
-  libbacktrace:
-    version: [2020-02-19]
+  #libbacktrace:
```
climbfuji, line 112: `spack env activate` fails with an error message complaining about the version number. I don't think it matters to pin the version here.

README.md
kgerheiser: Didn't know that. The fewer the better I suppose.

climbfuji:
Hang on, let's make this a draft until I got the nceplibs-bufr Python interface in there.

kgerheiser:
I like the changes

kgerheiser:
Why does site need to go first? I thought it was the opposite and should go last so that it has the final say, but maybe my understanding of the config order is wrong.

climbfuji:
I created a separate PR for this - https://github.com/NOAA-EMC/spack-stack/pull/20

I don't know why, but I left instructions that allow you to see that yourself. The way how spack pieces together configs is pretty confusing.

climbfuji:
Took me some time to figure this out in my original branches.

kgerheiser (approved):

PR #20, opened 2022-02-10
Order matters! Change order of including yamls in spack.yaml

climbfuji:
`site.yaml` needs to go first, otherwise site-specific overrides of packages and compiler configs will be ignored.

How to reproduce: use current order on macOS and do `git config blame packages` - you'll see that fms is built with OpenMP, although the macOS site config says no OpenMP.


kgerheiser:
I thought it was the opposite and should come last to take precedence.

kgerheiser:
https://spack.readthedocs.io/en/latest/environments.html#included-configurations

` Included configs listed earlier will have higher precedence, as the included configs are applied in reverse order.`

climbfuji:
> https://spack.readthedocs.io/en/latest/environments.html#included-configurations
> 
> ` Included configs listed earlier will have higher precedence, as the included configs are applied in reverse order.`

Sometimes it would help reading the documentation ... thanks for doing that on my behalf ;-)

PR #22, opened 2022-02-11
GitHub action improvements

### LINKED ISSUE #10: Improve Github Actions
Current Action runs using GCC on Ubuntu 20.04. Expand this to Intel compilers, macOS, etc.

kgerheiser:
Just wanted to get this setup to prove that everything can build on across different compilers/os.

I created a generic composite action that builds a spack-stack env. Then, I have each OS-specific workflow call that to avoid duplicating code.

Everything successfully builds using ubuntu-gcc (gcc-9), ubuntu-intel, macos-clang, macos-gcc (gcc-10). 

See results here: https://github.com/kgerheiser/spack-stack/actions

Fix #10


kgerheiser:
@climbfuji ready for this to be reviewed. I think it'll be good to have some automatic testing that you can run. The builds take forever so I have the Action set to run manually (the option appears after being merged to the main branch). Everything builds across Ubuntu/macOS/Intel/Clang/GNU.

.github/actions/setup-spack-stack/action.yaml
```diff
@@ -0,0 +1,126 @@
...
+        fi
+      elif [ "$RUNNER_OS" == "macOS" ]; then
+        echo ""
+      fi
+
+  - name: setup-spack-env
+    shell: bash
+    run: |
+      cd ${{ inputs.path }}
+      if [[ "${{ inputs.compiler }}" == "intel" ]]; then
```
climbfuji, line 58: Is this still needed? In line 48, the sourcing of `/opt/intel/oneapi/setvars.sh` is added to `.bash_profile`.

.github/actions/setup-spack-stack/action.yaml
```diff
@@ -0,0 +1,126 @@
...
+  - name: build-env
+    shell: bash
+    run: |
+      cd ${{ inputs.path }}
+      source setup.sh
+      spack env activate ${{ inputs.path }}/envs/${{ inputs.app }}
+      if [[ "${{ inputs.compiler }}" == "intel" ]]; then
+        source /opt/intel/oneapi/setvars.sh
+        # Pass --dirty to fix Intel compiler environment
+        spack install --dirty
```
climbfuji: We don't want users (and us) to do that when we install the stack on our machines. I was able to use Intel compilers (classic and oneAPI) on the HPCs without the option. Do we need to fine tune the spack compiler config to get around this?

climbfuji (commented):
This looks great, thanks a lot for setting this up. I have only two questions below ...

.github/actions/setup-spack-stack/action.yaml
kgerheiser: Yes, the Action doesn't have an option to run Bash with `-l`. So, steps within the Action have to do it manually.

.github/actions/setup-spack-stack/action.yaml
kgerheiser: It's a quirk with Intel compilers on Ubuntu that I could never get around. ifort/MPI don't work because of `libimf.so`. I've tried setting various environment variables through `compilers.yaml`, etc.

climbfuji: I had the same problem on Cheyenne and Orion. The way around this was to set/extend the `LD_LIBRARY_PATH` environment variable in the compiler config, see the two site configs.

climbfuji: Let's merge this now and see if we can avoid the `--dirty` flag for Intel in the future.

climbfuji (approved):

kgerheiser:
Intel build is still broken FYI, I was testing out a way to add `libimf` to compilers.yaml LD_LIBRARY_PATH through the command line.

I'll fix it in a follow up

climbfuji:
Thanks, no problem. I hit the merge button too early, sorry.

PR #23, opened 2022-02-13
Change py-bufr to bufr with +python variant

kgerheiser:
Related to https://github.com/NOAA-EMC/spack/pull/10

### DIFF
--- a/configs/common/packages.yaml
+++ b/configs/common/packages.yaml
@@ -143,9 +143,9 @@
     py-scipy:
       # Does this version work with all the other packages listed here? 1.5.3 worked ...
       version: [1.7.3]
-    py-bufr:
+    bufr:
       version: [11.5.0]
-      variants: +enable_python
+      variants: +python
     diffutils:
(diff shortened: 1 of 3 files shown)


climbfuji:
The problem I had was that it didn't want to find pip when the name was "bufr". But maybe the change to use "+python" does the same trick that the name "py-xyz" does. I used "enable_python" because that's how it was called in the nceplibs-bufr `CMakeLists.txt`. Using `python` is much better anyway.

kgerheiser:
Hmm, I see the error. Other (non-Python) packages (fenics, cmor, mxnet) do the same thing, so it should be possible.

kgerheiser:
Got it to work, just had to add `depends_on('py-pip')`

kgerheiser:
@climbfuji can we merge this?

climbfuji (approved):

climbfuji:
Yes! Sorry, I lost track of this one.

PR #25, opened 2022-02-13
Split site configs

### LINKED ISSUE #24: Split site.yaml into individual config files
Instead of using a singular `site.yaml`, I think we should keep the individual configs (`packages.yaml`, `compilers.yaml`, etc) in the repository.

This makes it easier to generate configs for each system because Spack already outputs these files instead of having to concatenate them together manually.

`create-env.py` can create the `site.yaml` from the config files automatically.

kgerheiser:
Fix #24 

Split `site.yaml` into individual configs. This makes it easier to generate configs for machines because Spack creates these individual configs already without having to manually put them together.

Then, `create-env.py` concatenates them into `site.yaml` when the environment is generated.


kgerheiser:
@climbfuji and also this PR

climbfuji (approved):
We'll have to add instructions for the user-generated sites to make this lua module trick work (compiler used to compile stack cannot be the core compiler).

PR #32, opened 2022-02-17
Add wgrib2

### LINKED ISSUE #29: Add wgrib2 package
Wgrib2 exists in Spack with the CMake fork, but the official release of wgrib2 should be used.

kgerheiser:
Fix #29

### DIFF
--- a/configs/repos/ufs/packages/wgrib2/package.py
+++ b/configs/repos/ufs/packages/wgrib2/package.py
@@ -117,4 +117,5 @@ class Wgrib2(MakefilePackage):
             install('lib/libwgrib2.a', prefix.lib)
             install('lib/wgrib2api.mod', prefix.include)
             install('lib/wgrib2lowapi.mod', prefix.include)
+
         
--- a/configs/repos/ufs/packages/wgrib2/package.py
+++ b/configs/repos/ufs/packages/wgrib2/package.py
@@ -74,10 +74,11 @@ class Wgrib2(MakefilePackage):
         'wmo_validation': 'USE_WMO_VALIDATION',
         'ipolates': 'USE_IPOLATES',
         'disable_timezone': 'DISABLE_TIMEZONE',
-        'disable_alarm': 'DISABLE_ALARM'
+        'disable_alarm': 'DISABLE_ALARM',
+        'fortran_api': 'MAKE_FTN_API'
     }
            
-
+    # Disable parallel build
     parallel = False
 
     def url_for_version(self, version):
--- a/configs/repos/ufs/packages/wgrib2/package.py
+++ b/configs/repos/ufs/packages/wgrib2/package.py
@@ -72,7 +72,9 @@ class Wgrib2(MakefilePackage):
         'jasper': 'USE_JASPER',
         'openmp': 'USE_OPENMP',
         'wmo_validation': 'USE_WMO_VALIDATION',
-        'ipolates': 'USE_IPOLATES'
+        'ipolates': 'USE_IPOLATES',
+        'disable_timezone': 'DISABLE_TIMEZONE',
+        'disable_alarm': 'DISABLE_ALARM'
     }
            
 
--- a/configs/repos/ufs/packages/wgrib2/package.py
+++ b/configs/repos/ufs/packages/wgrib2/package.py
@@ -21,6 +21,8 @@
 # ----------------------------------------------------------------------------
 
 import os
+import sys
+import re
 
 from spack import *
 
@@ -71,6 +73,8 @@ class Wgrib2(MakefilePackage):
     variant('openmp', default=True, description='OpenMP parallelization')
     variant('wmo_validation', default=False, description='WMO validation')
 
+    conflicts('+openmp', when='%apple-clang')
+
     variant_map = {
         'netcdf3': 'USE_NETCDF3',
         'netcdf4': 'USE_NETCDF4',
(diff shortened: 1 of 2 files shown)


configs/repos/ufs/packages/wgrib2/package.py
```diff
@@ -0,0 +1,129 @@
...
+        return url.format(version)
+
+    def edit(self, spec, prefix):
+        makefile = FileFilter('makefile')
+
+        for variant_name, makefile_option in self.variant_map.items():
+            value = int(spec.variants[variant_name].value)
+            makefile.filter(r'^%s=.*' % makefile_option, '{}={}'.format(makefile_option, value))
+
+    def build(self, spec, prefix):
```
climbfuji, line 149: I see you are installing everything as part of the build. I had to make a similar choice for a new package `shumlib`, but in this case it was possible to "install" during the build phase into a subdirectory "build" and then in the real install phase move things over using `install_tree`.

I thought this would reduce the risk of a half broken installation. For example, if building `wgrib2` works, but the fortran api fails, will the wgrib2 binary already be in place - or will spack roll back the install command? Do you know?

kgerheiser: The thing is that wgrib2 is built in two phases for the executable and the library, with a `make clean` in between which deletes the executable. I think I could work around that by saving the executable before cleaning.

edwardhartnett:
We are trying to move away from wgrib2 in operations. Who needs to use it?

The reason we are trying to move away from it is because of all it's weird build problems, which create a tremendous amount of extra work and trouble...

kgerheiser:
A lot of people use wgrib2 for command-line manipulation of grib2.

It's used operationally by various applications.

UFS_UTILS uses the library portion, and though it's moving to g2 we still have to be able to build it.

edwardhartnett:
Yes, I know, it's used in a lot of places. THat makes it hard. So we should discourage future use, if we can...

kgerheiser:
We still need the ability to install it. And this `package.py` abstracts away some of that complexity.

configs/repos/ufs/packages/wgrib2/package.py
kgerheiser: Fixed. I move the products into a temporary directory and use `install_tree` to install.

climbfuji:
I tried to install this from your branch on my Mac with llvm-clang on Monterey and got the following error:
```
/Users/heinzell/work/jedi-stack/spack-stack-new-joint/spack-stack-develop-20220215/spack/lib/spack/env/clang/gfortran   splat.f90
Undefined symbols for architecture x86_64:
  "_main", referenced from:
     implicit entry/start for main executable
ld: symbol(s) not found for architecture x86_64
collect2: error: ld returned 1 exit status
make[1]: *** [splat.o] Error 1
```
Too late to figure this out today. Could be one of these pseudo case-sensitive filesystem issues.

kgerheiser:
Closing until #36 is complete

kgerheiser:
And I'll give it a try with Clang

climbfuji:
Reopening because #36 is not going to happen for now.

climbfuji:
I picked this up again today. It's easy to get around the `-c` flag, but a really weird one is that adding the "usual" gfortran-10+ flag `-fallow-argument-mismatch` seems to be ignored:
```
gfortran -fallow-argument-mismatch gdswzd_c.f90
gdswzd_c.f90:233:48:

  233 |              CROT,SROT,XLON,XLAT,YLON,YLAT,AREA)
      |                                                1
Error: Type mismatch in argument 'fill' at (1); passed REAL(8) to REAL(4)
```
Maybe I've been staring at my screen for too many hours today ...

kgerheiser:
Fixed with a bit of a hack to find the gfortran libdir, using `gfortran -###`, when using Clang.

```
            if self.spec.compiler.name in ['apple-clang', 'clang']:
                if 'gfortran' in self.compiler.fc:
                    output = Executable(self.compiler.fc)('-###', output=str, error=str)
                    libdir = re.search('--libdir=(.+?) ', output).group(1)
                    flags.append('-L{}'.format(libdir))
```

Tested with apple-clang and clang

kgerheiser:
Ready for review @climbfuji

climbfuji:
@kgerheiser I got an error building wgrib2 in the libaec step:
```
checking for gcc... /Users/heinzell/scratch/spack-stack-package-wgrib2-20220309/spack/lib/spack/env/clang/clang
checking whether the C compiler works... no
configure: error: in `/Users/heinzell/scratch/spack-stack-package-wgrib2-20220309/cache/build_stage/spack-stage-wgrib2-2.0.8-as3xj7ysvzf424mkz2eeiwywk777khc2/spack-src/libaec-1.0.2':
configure: error: C compiler cannot create executables
See `config.log' for more details
make: *** [/Users/heinzell/scratch/spack-stack-package-wgrib2-20220309/cache/build_stage/spack-stage-wgrib2-2.0.8-as3xj7ysvzf424mkz2eeiwywk777khc2/spack-src/lib/libaec.a] Error 77
==> Error: ProcessError: Command exited with status 2:
    'make'
```
The issue is an unsupported compiler flag I think:
```
configure:2906: checking whether the C compiler works
configure:2928: /Users/heinzell/scratch/spack-stack-package-wgrib2-20220309/spack/lib/spack/env/clang/clang -I/Users/heinzell/scratch/spack-stack-package-wgrib2-20220309/cache/build_stage/spack-stage-wgrib2-2.0.8-as3xj7ysvzf424mkz2eeiwywk777khc2/spack-src/include -Wmissing-prototypes -Wold-style-definition -Werror=format-security --fast-math -O3 -DGFORTRAN -fopenmp -I/Users/heinzell/scratch/spack-stack-package-wgrib2-20220309/cache/build_stage/spack-stage-wgrib2-2.0.8-as3xj7ysvzf424mkz2eeiwywk777khc2/spack-src/jasper-1.900.1/src/libjasper/include -I/usr/include    conftest.c  >&5
clang-13: error: unsupported option '--fast-math'; did you mean '-ffast-math'?
```
It's in the top-level `makefile`, line 286:
```
ifeq (${COMP_SYS},gnu_linux)
   wCPPFLAGS+=  -Wmissing-prototypes -Wold-style-definition -Werror=format-security --fast-math -O3
```
If you look a few lines above (240+), you'll see the correct flag in what seems to be the "old configuration method":
```
# C compile and load commmands
# wCPPFLAGS has the directory of the includes 
# wLDFLAGS has the directory/name of the library

ifeq ($(findstring opencc,$(notdir $(CC))),opencc)
   wCPPFLAGS+=-O3   -ffast-math -opencc
   netcdf3CPPFLAGS+=-O3   -ffast-math -opencc
   netcdf4CPPFLAGS+=-O3   -ffast-math -opencc
   hdf5CFLAGS+=-O1   -opencc
endif
```

climbfuji:
I may have an update for your PR ready in a bit, unless I hit more errors further in the build process.

climbfuji (approved):

PR #38, opened 2022-02-22
Add jedi-um app, jedi-all app and many other updates

### LINKED ISSUE #5: Debug version of ESMF
UFS uses two a release and debug build of ESMF. I think this can be accomplished with variant in the ESMF package.py and using the suffix option in modulefiles.

### LINKED ISSUE #15: Need to capture dependency of mapl on ESMF?
mapl depends on the ESMF version. In hpc-stack, we capture this by append the ESMF version to the mapl version. This is, however, somewhat inconsistent, because we don't do this for any other package that depends on some package in the stack (i.e. we don't add the netcdf version to any of the packages, even though many of the packages depend on it).

Should we try to be more consistent and drop the ESMF dependency information from the mapl install/module name? Simply roll out an entirely new env if a new mapl/esmf combination is required?

Or are ESMF and map outliers that change a lot more often than netcdf such that it makes sense to add this hierarchy information? We should be able to do this with spack.

### LINKED ISSUE #34: HDF5 1.10.6 Fortran modules are installed in subdirectories of "include"
For `hdf5@1.10.6 +fortran +shared`, identical sets of Fortran modules are installed in subdirectories `shared` and `static` of the `include` directory. These are not found by cmake and they are also not part of the default `CPATH` in the lua module files.

Upgrading to hdf5@1.12.1 fixes the problem.

Since there are other good reasons for using version 1.12.1 (better performance as noted recently by @edwardhartnett), and since the default version in the existing jedi-stack is 1.12.0, this upgrade is not a problem from the JEDI side.

The ufs-weather-model currently uses 1.10.6 (https://github.com/ufs-community/ufs-weather-model/blob/develop/modulefiles/ufs_common). @kgerheiser @edwardhartnett do you think we can update to 1.12.1 for the UFS?

### LINKED ISSUE #37: HDF5 1.10.6 version issues
For the jedi-fv3 environment, the stack installs HDF5 version 1.10.6, but ioda requires HDF5 1.12.0 or greater to run successfully (v1.12.1 preferred).

HDF5 1.12.0 refactored several functions related to chunking and dataspace selections that greatly improved performance when reading and writing large files. IODA also uses several of the new APIs in a few parts of the code (and should error out in ctest when these functions are not available).

https://www.hdfgroup.org/2020/03/release-of-hdf5-1-12-0-newsletter-172/

### LINKED ISSUE #46: Setting up compiler toolchains
Many of the JEDI packages are written in C++ and require support for newer C++ standards (currently `c++-14`, soon hopefully `c++-17`). To compile these packages with the current jedi-stack on systems where the default `libstdc++` is too old, we set up toolchains as follows (e.g. cheyenne):
```
# C++-14 compliant compiler settings
# set / export these variables when building for Intel compiler(s)
if [[ "$JEDI_COMPILER" =~ .*"intel"* ]]; then
    export CXXFLAGS="-gxx-name=/glade/u/apps/ch/opt/gnu/9.1.0/bin/g++ -Wl,-rpath,/glade/u/apps/ch/opt/gnu/9.1.0/lib64"
    export LDFLAGS="-gxx-name=/glade/u/apps/ch/opt/gnu/9.1.0/bin/g++ -Wl,-rpath,/glade/u/apps/ch/opt/gnu/9.1.0/lib64"
fi
```
This works within the limited scope of jedi-stack, but within spack and as I we are adding more software, doing the same in the compiler config breaks a lot of packages. I think I found a way to configure this in spack so that it works for all packages (currently building):
```
  - compiler:
      spec: intel@2021.2.0
      paths:
        cc: /glade/u/apps/opt/intel/2021.2/compiler/latest/linux/bin/intel64/icc
        cxx: /glade/u/apps/opt/intel/2021.2/compiler/latest/linux/bin/intel64/icpc
        f77: /glade/u/apps/opt/intel/2021.2/compiler/latest/linux/bin/intel64/ifort
        fc: /glade/u/apps/opt/intel/2021.2/compiler/latest/linux/bin/intel64/ifort
      flags: {}
      operating_system: sles12
      target: x86_64
      modules:
      - intel/2021.2
      environment:
        prepend_path:
          LD_LIBRARY_PATH: '/glade/u/apps/opt/intel/2021.2/compiler/2021.2.0/linux/compiler/lib/intel64_lin:/glade/u/apps/ch/opt/gnu/10.1.0/lib64'
          CXXFLAGS: '-std=c++14 -gxx-name=/glade/u/apps/ch/opt/gnu/10.1.0/bin/g++ -Wl,-rpath,/glade/u/apps/ch/opt/gnu/10.1.0/lib64'
      extra_rpaths: []
```
Note 1. I don't think `-std=c++14` is required, but it's currently building with that. Will try again later without it.
Note 2. For the reason to have the additional `LD_LIBRARY_PATH` `/glade/u/apps/opt/intel/2021.2/compiler/2021.2.0/linux/compiler/lib/intel64_lin` see #28
Note 3. Right now I also have
```
        set:
          I_MPI_ROOT: '/glade/u/apps/opt/intel/2021.2/mpi/latest'
```
in there to avoid these nasty `I_MPI_SUBSTITUTE_INSTALL_DIR` errors in various places. This can hopefully be removed, because we don't want any MPI dependency in the compiler config!

### LINKED ISSUE #47: Add netcdf-cxx4 to the stack
JEDI uses netcdf-cxx4, which is not yet in the spack-stack environments.

climbfuji:
This PR is an accumulation fo work over the last few weeks and contains many changes and updates:
- Add two new applications, jedi-um and jedi-tools
- Add a new github workflow for macOS with llvm-clang, in addition to the existing apple-clang workflow
- Complete list of applications to build in github workflows
- Enable the use of either lmod or tcl modules
    - Both work in exactly the same way, including the compiler and MPI hierarchy
        - The only difference is: with tcl modulefiles, one has to rerun `meta_modules/setup_meta_modules.py` **EVERY TIME** after `spack module tcl refresh` (this is not necessary for lmod, but it doesn't hurt, so maybe we just make it part of the default documentation)
    - This required small changes in `meta_modules/setup_meta_modules.py`
    - As of now, there is a substantial amount of duplication between the spack lmod and tcl module config - check if this can be combined into one
- Append `-debug` to modulefiles when package is built with `+debug`
- Append esmf version info to mapl modulefiles
- Fix Cheyenne site config (compiler and package configs) - **THIS IS STILL BEING WORKED ON AND WILL BE FINALIZED IN A FOLLOW-UP PR`
- Update macOS example config
- Add site config for gaea (which uses tcl modulefiles) - works great, no need to run the `init_lmod.py` script anymore
- Change `type=('build', 'run')` to `type='run'` for all our virtual packages - nothing gets built
- Add new packages required by jedi-um application: shumlib, faux, trans
- Add a jedi-base-env that sits between base-env and the individual JEDI applications
- Rename ufs-weather-model-bundle-env to ufs-weather-model-env because it's not really a bundle (in the way bundles are defined in ecbuild)
- Temporarily comment out `wgrib2` in nceplibs-bundle
- Orion site config tested

Still todo (in this or a follow-up PR)
- Update `README.md` in all `config/sites/*` directories
- Look into renaming nceplibs-bundle because it's not really a bundle (in the way bundles are defined in ecbuild)
- Check if the duplication between tcl and lmod modulefile config can be avoided
- Fix/update hera site config
- Test cheyenne site config with other compiler versions, fix netcdf build errors

Fixes https://github.com/NOAA-EMC/spack-stack/issues/47
Fixes https://github.com/NOAA-EMC/spack-stack/issues/46
Fixes https://github.com/NOAA-EMC/spack-stack/issues/37
Fixes https://github.com/NOAA-EMC/spack-stack/issues/34
Fixes https://github.com/NOAA-EMC/spack-stack/issues/15
Fixes https://github.com/NOAA-EMC/spack-stack/issues/5


.github/workflows/macos-llvm-clang.yaml
```diff
@@ -23,6 +23,6 @@ jobs:
         uses: ./.github/actions/setup-spack-stack
         with:
           app: ${{ matrix.app }}
-          compiler: apple-clang
+          compiler: clang
```
climbfuji, line 26: Note to self. Do we need to add to PATH and LD_LIBRARY_PATH to find homebrew's llvm-clang?

configs/common/modules.yaml
```diff
@@ -1,9 +1,83 @@
 modules:
   default:
     arch_folder: false
-    enable:
-    - lmod
+    tcl:
+      # Core compiler is a dummy that is not used to build the stack.
```
climbfuji, line 5: Note. Almost the entire 'tcl' section is a repetition of the 'lmod' section. Can we use 'all' for the common elements?

configs/repos/ufs/packages/nceplibs-bundle/package.py
```diff
@@ -53,6 +53,6 @@ class NceplibsBundle(BundlePackage):
     depends_on('w3emc')
     depends_on('w3nco')
     depends_on('wrf-io')
-    depends_on('wgrib2')
+    #depends_on('wgrib2')
```
climbfuji, line 56:
```
  Fetch: 1.37s.  Build: 0.42s.  Total: 1.79s.
[+] /glade/work/heinzell/jedi-stack/spack-stack-develop-20220215/envs/cheyenne-intel-2021.2.0-all/install/intel/2021.2.0/shumlib-macos_clang_port-qtrwavv
==> Error: Unable to parse extension from https://www.ftp.cpc.ncep.noaa.gov/wd51we/wgrib2/wgrib2.tgz.v2.0.8-cmake-v6.

If this URL is for a tarball but does not include the file extension
in the name, you can explicitly declare it with the following syntax:

    version('1.2.3', 'hash', extension='tar.gz')

If this URL is for a download like a .jar or .whl that does not need
to be expanded, or an uncompressed installation script, you can tell
Spack not to expand it with the following syntax:

    version('1.2.3', 'hash', expand=False)
```

climbfuji:
@kgerheiser I am opening this up for review, I could be working for many more weeks and make the PR even bigger. A few last tests are running on Cheyenne, but I think we should get these changes checked in and test/fix hera, orion, and cheyenne with other compilers in a follow-up PR. After all, this repo is still in early development and not used anywhere for development and production.

.gitmodules
```diff
@@ -2,5 +2,7 @@
   path = spack
   #url = https://github.com/spack/spack
   #branch = develop
-  url = https://github.com/NOAA-EMC/spack.git
-  branch = jcsda_emc_spack_stack
+  #url = https://github.com/NOAA-EMC/spack.git
+  #branch = jcsda_emc_spack_stack
+  url = https://github.com/climbfuji/spack
```
kgerheiser: Do we want to change the submodule here?

configs/common/modules.yaml
```diff
@@ -1,9 +1,83 @@
...
+          ^esmf@8.2.0~debug: 'esmf-8.2.0'
+          ^esmf@8.2.0+debug: 'esmf-8.2.0-debug'
+          +debug^esmf~debug: 'debug-esmf-8.2.0'
+          +debug^esmf+debug: 'debug-esmf-8.2.0-debug'
+
     lmod:
+      # Core compiler is a dummy that is not used to build the stack.
+      # Hopefully nobody is trying to build software with gcc-4.6 in 2022 ...
+      core_compilers::
+      - gcc@4.6
```
kgerheiser, line 80: Can `core_compilers` be set this for all modules even if a machine doesn't have a spec for gcc@4.6?

configs/repos/jedi/packages/faux/package.py
```diff
@@ -0,0 +1,39 @@
...
+    # DH* TODO UPDATE
+    homepage = "https://github.com/JCSDA-internal/faux"
+    git = "https://github.com/JCSDA-internal/faux.git"
+
+    maintainers = ['climbfuji']
+
+    version('develop', branch='develop', no_cache=True, preferred=True)
+    #version('bugfix-compile-failures', branch='bugfix/fix_compile_failures', no_cache=True)
+
+    variant('enable_mpi',   default=True, description='Use MPI?')
```
kgerheiser: Just a general comment that the preferred variant names are `+mpi`, `+openmp`, `+fckit`. 

https://spack.readthedocs.io/en/latest/packaging_guide.html#variant-names

configs/sites/hera/packages.yaml
```diff
@@ -1,30 +1,31 @@
...
     - spec: openmpi@3.1.4%gcc@9.2.0~cuda+cxx+cxx_exceptions~java~memchecker+pmi+pmix~sqlite3+static~thread_multiple~wrapper-rpath fabrics=mxm schedulers=slurm
       modules:
       - openmpi/3.1.4
+  python:
+    buildable: False
+    externals:
+    - spec: python@3.9.7+bz2+ctypes+dbm+ensurepip+lzma+nis+pyexpat+pythoncmd~readline+sqlite3+ssl+tix+tkinter+uuid+zlib
+      modules:
+      - intelpython/2022.1.2
+      prefix: /apps/oneapi/intelpython/python3.9
```
kgerheiser: I thought we wanted to use a different Python?

kgerheiser (changes requested):

configs/common/modules.yaml
climbfuji: Yes, it's completely ignored.

configs/repos/jedi/packages/faux/package.py
climbfuji: Yes, I know. The old problem with ECMWF packages using ecbuild. We could, of course, translate this using a variant map. Maybe that's a good idea.

configs/sites/hera/packages.yaml
climbfuji: Yes, hera is still in the works  (see "still todo section"). Let me push what I have so far.

.gitmodules
climbfuji: Temporarily, I need to create a PR for our internal spack repo. I am cherry-picking the commits that I submit to the official spack repo in parallel to avoid merge conflicts down the road.

configs/repos/jedi/packages/faux/package.py
kgerheiser: No need for any special logic with a CMakePackage. It's built into Spack.

```
self.define_from_variant('BUILD_WITH_FLAP', 'flap'),
self.define_from_variant('BUILD_WITH_PFLOGGER', 'pflogger'),
self.define_from_variant('ESMA_USE_GFE_NAMESPACE', 'esma_gfe_namespace'),
self.define_from_variant('BUILD_SHARED_MAPL', 'shared'),
```

climbfuji: That's great. I'll make those changes in the afternoon and retest.

configs/sites/hera/packages.yaml
climbfuji: Ok, I pushed my updates.

configs/apps/all/spack.yaml
```diff
@@ -9,7 +9,6 @@ spack:
     - ${SPACK_STACK_DIR}/configs/repos/repos.yaml
 
   specs:
-    - jedi-fv3-bundle-env
-    - jedi-ufs-bundle-env
-    - ufs-weather-model-bundle-env
+    - jedi-env
+    - ufs-weather-model-env
     - nceplibs-bundle
```
climbfuji, line 14: @kgerheiser Is nceplibs-bundle redundant, or does it contain ncep libraries that are not built as part of ufs-weather-model-env?

configs/repos/jedi/packages/trans/package.py
```diff
@@ -0,0 +1,74 @@
...
+    version('develop', branch='develop', no_cache=True, preferred=True)
+    #version('bugfix-compile-failures', branch='bugfix/fix_compile_failures', no_cache=True)
+
+    variant('enable_mpi', default=True, description='Use MPI?')
+    variant('enable_omp', default=True, description='Use OpenMP?')
+
+    #variant('enable_double_precision', default=True, description='Enable double precision?')
+    #variant('enable_single_precision', default=True, description='Enable single precision?')
+
+    variant('enable_mkl',  default=True, description='Use MKL?')
```
climbfuji: Note to self. Turn off MKL for now (then can remove the special package config for macos) to avoid problems with multiple specs of large parts of the software stack when building all applications, as observed on gaea and orion.

climbfuji:
@kgerheiser This is ready to merge. Any changes to fix netcdf on Cheyenne will need to happen in a follow-up PR.

.gitmodules
```diff
@@ -1,6 +1,6 @@
 [submodule "spack"]
   path = spack
-  #url = https://github.com/spack/spack
-  #branch = develop
+  url = https://github.com/spack/spack
```
kgerheiser: Is this correct? Why two URLs for the submodule?

configs/apps/all/spack.yaml
kgerheiser: It builds things that ufs-weather-model doesn't necessarily use

configs/common/modules.yaml
kgerheiser: Yes, I think you can do that.

configs/repos/common/packages/base-env/package.py
```diff
@@ -21,19 +20,19 @@ class BaseEnv(BundlePackage):
...
-    depends_on('parallelio', type=('build', 'run'))
-    depends_on('nccmp', type=('build', 'run'))
+    depends_on('zlib', type='run')
+    depends_on('szip', type='run')
+    depends_on('hdf5', type='run')
+    depends_on('netcdf-c', type='run')
+    depends_on('netcdf-fortran', type='run')
+    depends_on('parallel-netcdf', type='run')
+    depends_on('parallelio', type='run')
+    depends_on('nccmp', type='run')
```
kgerheiser, line 35: Are these not build dependencies?

configs/repos/ufs/packages/nceplibs-bundle/package.py
kgerheiser: This will be fixed with #32

configs/sites/gaea/modules.yaml
```diff
@@ -0,0 +1,8 @@
+modules:
+  default:
+    enable::
+    - tcl
+    #tcl:
+    #  projections:
+    #    git: 'Core/{name}/{version}'
+    #    git-lfs: 'Core//{name}/{version}'
```
kgerheiser: Not a big deal, but could you fix the newline thing?

configs/sites/hera/packages.yaml
```diff
@@ -1,30 +1,31 @@
...
     - spec: openmpi@3.1.4%gcc@9.2.0~cuda+cxx+cxx_exceptions~java~memchecker+pmi+pmix~sqlite3+static~thread_multiple~wrapper-rpath fabrics=mxm schedulers=slurm
       modules:
       - openmpi/3.1.4
+  python:
+    buildable: False
+    externals:
+    - spec: python@3.9.7
+      modules:
+      - miniconda/3.9.7
+      prefix: /scratch1/BMC/gsd-hpcs/Dom.Heinzeller/spack-stack/miniconda-3.9.7
```
kgerheiser, line 27: Nice to not have to deal with Intel Python here

configs/sites/macos/README.md
```diff
@@ -32,35 +32,17 @@ brew install git-lfs@3.0.2
...
-# don't use but which we also don't want to have on the system
-# brew install eccodes@2.24.2
-```
 - Activate the `lua` module environment:
 ```
 source /usr/local/opt/lmod/init/profile
 ```
 
-## Building your own spack stack
+This environment enables working with spack and building new software environents, as well as loading modules that are created by spack for building JEDI and UFS software.
```
kgerheiser: Typo `environents`

kgerheiser (changes requested):

configs/common/modules.yaml
climbfuji: Ok, this will be a follow-up PR - see https://github.com/NOAA-EMC/spack-stack/issues/50

.github/workflows/macos-llvm-clang.yaml
climbfuji: Will figure out when trying the CI tests.

configs/repos/ufs/packages/nceplibs-bundle/package.py
climbfuji: Ok, when #32 gets merged after this PR, let's put it back in.

configs/repos/jedi/packages/faux/package.py
climbfuji: Done.

.gitmodules
climbfuji: No, that is wrong. Thanks for catching this!

configs/repos/common/packages/base-env/package.py
climbfuji: The thing is, none of these "envs" builds anything. The build dependencies are in each of the packages that get installed. For the envs, you only need the "run" dependency so that the modules get loaded when the env module gets loaded.

configs/sites/gaea/modules.yaml
climbfuji: Fixed.

kgerheiser (approved):

climbfuji:
Thanks, @kgerheiser . I am fixing a last inconsistency in the FFTW/MKL/LAPACK configuration, once that's done we can finally merge this. Thanks for your review and your patience.

climbfuji:
> Thanks, @kgerheiser . I am fixing a last inconsistency in the FFTW/MKL/LAPACK configuration, once that's done we can finally merge this. Thanks for your review and your patience.

I think I got those inconsistencies fixed. Will build the stack once more on gaea, if that works then I'll merge.

Issue #16, opened 2022-02-09
Boostrap Clingo on Hera


kgerheiser:
Hera (and other RDHPCS platforms) block outgoing connections by default. This means Spack can't download the Clingo binary and it has to be built from source.

It also turns out there was another bug which prevented bootstrapping from source  because Spack was detecting Hera as a Cray system. 

Applying the fix here allowed me to build Clingo: https://github.com/spack/spack/pull/28726

This requires a GCC compile in your compilers.yaml that supports C++14 (> GCC 4.0).

Then, I ran `spack bootstrap untrust github-actions` 

Then, the firewall prevented me from doing more, but I think a Spack mirror would resolve that issue.

kgerheiser:
Can download Clingo now. The Spack mirror is still necessary for many external packages, but it's better than building Clingo.

climbfuji:
> Can download Clingo now. The Spack mirror is still necessary for many external packages, but it's better than building Clingo.

Mirror not necessary on hera for now - I could build fv3-bundle-env, ufs-bundle-env, um-bundle-env without using a mirror.

kgerheiser:
Wow, I'm surprised nothing in our package list was behind the firewall.

climbfuji:
Hera update PR coming in a bit.

PR #43, opened 2022-02-23
Add ability to create mirror using Github Actions

### LINKED ISSUE #26: Set cache/stage dir in config.yaml
I think the Spack staging/cache directories should be set to reside within spack-stack by default with `${SPACK_STACK_DIR}/cache`. Mirrors and package sources are difficult to find because they are placed in some opaque temp directory.

```
  # The build stage can be purged with `spack clean --stage` and
  # `spack clean -a`, so it is important that the specified directory uniquely
  # identifies Spack staging to avoid accidentally wiping out non-Spack work.
  build_stage:
    ${SPACK_STACK_DIR}/cache/build_stage

  # Directory in which to run tests and store test results.
  # Tests will be stored in directories named by date/time and package
  # name/hash.
  test_stage: ${SPACK_STACK_DIR}/cache/test_stage

  # Cache directory for already downloaded source tarballs and archived
  # repositories. This can be purged with `spack clean --downloads`.
  source_cache: ${SPACK_STACK_DIR}/cache/source_cache


  # Cache directory for miscellaneous files, like the package index.
  # This can be purged with `spack clean --misc-cache`
  misc_cache: ${SPACK_STACK_DIR}/cache/misc_cache
```

### LINKED ISSUE #27: Add Github Action that automatically creates a Spack mirror
It would be useful to have a Spack mirror automatically created from a Github Action, so downloaded so that firewalled systems (e.g. Jet, Hera) can access these packages, rather than relying on others to create their own mirrors.

Still have to work on the details, but I think this should be possible by generating the mirror and then uploading it as an artifact. Maybe have it as a manual action, or whenever there's a merge to develop?

### LINKED ISSUE #40: Specify providers for blas, lapack, fftw
We should specify the providers for blas, lapack, fftw in the site config (or provide instructions for doing so for the default configs). intel-oneapi-mkl or intel-mkl work fine on most HPCs that have Intel compilers installed, but on macOS I had lots of trouble with intel-oneapi-mkl and building Python packages. I ended up removing mkl and using this instead:
```
      blas:: [openblas]
      lapack:: [openblas]
      fftw-api:: [fftw]
```

kgerheiser:
Create a workflow that creates a Spack mirror using a newly created "all" app that combines all current bundles into one environment. I figure this will also be useful to have in case you want to build everything. The workflow is run manually and then the artifact can be downloaded which contains all the individual package sources.

The mirror is placed in `cache/source_cache` (#26)

This can be used in Spack like `spack mirror add my_mirror file://$HOME/spack-mirror`

An example artifact can be seen here: https://github.com/kgerheiser/spack-stack/actions/runs/1885201334

Along the way I added the providers for #40 so to be sure the correct packages are downloaded.

Some of the patches downloaded for the mirror have question marks from links in their name which upload-artifact does not allow, so I delete them before uploading. None of them are relevant to the versions that we use.

Fix #40
Fix #27 
Fix #26

And working towards #16

Also disables the automatic (and failing) Intel build to be consistent with the rest of the workflow that are set to run manually.

### DIFF
--- a/.github/workflows/create-spack-mirror.yaml
+++ b/.github/workflows/create-spack-mirror.yaml
@@ -1,0 +1,45 @@
+name: create-spack-mirror
+
+on:
+  workflow_dispatch:
+    inputs:
+      app:
+        type: string
+        description: "App to generate mirror for"
+        required: true
+        default: "all"
+      site:
+        type: string
+        description: "Site to create environment for"
+        required: true
+        default: "default"
+jobs:
+  create-spack-mirror:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+        with:
+          submodules: true
+
+      - name: create-env
+        run: |
+          source ./setup.sh
+          ./create-env.py --app ${{ github.event.inputs.app }}  --site ${{ github.event.inputs.site }}  --name all
+          spack env activate -d envs/all
+          spack compiler find
+          spack add re2c
+          spack add clingo
+          spack concretize
+          spack mirror create -a
+          # upload-artifact does not allow filenames that begin with '?'
+          rm cache/source_cache/autoconf/\?id=05972f49ee632cd98057a3caf82ebfb9574846da-eaa3f69
+          rm cache/source_cache/boost/attachment.cgi?id=356970-b6f6ce6
+          rm cache/source_cache/py-scipy/extern_decls.patch?id=711fe05025795e44b84233e065d240859ccae5bd-5433f60
+
+      - name: upload-mirror
+        uses: actions/upload-artifact@v2
+        with:
+          name: spack_mirror
+          path: cache/source_cache
--- a/configs/apps/all/spack.yaml
+++ b/configs/apps/all/spack.yaml
@@ -8,14 +8,7 @@ spack:
     - modules.yaml
     - ${SPACK_STACK_DIR}/configs/repos/repos.yaml
 
-  packages:
-    all:
-      compiler:: [gcc@9.4.0]
-    mpi:
-      buildable: False
-
   specs:
-    - re2c
     - jedi-fv3-bundle-env
     - jedi-ufs-bundle-env
     - ufs-weather-model-bundle-env
--- a/configs/apps/all/spack.yaml
+++ b/configs/apps/all/spack.yaml
@@ -1,0 +1,22 @@
+spack:
+  concretization: separately
+  view: false
+  include:
+    - site.yaml
+    - config.yaml
+    - packages.yaml
+    - modules.yaml
+    - ${SPACK_STACK_DIR}/configs/repos/repos.yaml
+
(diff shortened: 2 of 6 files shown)


configs/apps/all/spack.yaml
```diff
@@ -0,0 +1,15 @@
...
+  include:
+    - site.yaml
+    - config.yaml
+    - packages.yaml
+    - modules.yaml
+    - ${SPACK_STACK_DIR}/configs/repos/repos.yaml
+
+  specs:
+    - jedi-fv3-bundle-env
+    - jedi-ufs-bundle-env
```
climbfuji, line 13: Note to self. Need to add the jedi-um-bundle later.

configs/common/packages.yaml
```diff
@@ -1,5 +1,10 @@
 # Pin versions and specs when building packages
   packages:
+    all:
```
climbfuji, line 3: Note. We may want to use MKL on HPCs when using the Intel compilers (i.e. when looking for the best performing code). Maybe it's possible to use single colons here so that we can overwrite it in the site configs, if needed?

climbfuji (approved):
This looks great! The "all" app is something I also wanted to add, just need to maintain it manually. I am not sure about hardcoding the linalg providers for all platforms, maybe we can overwrite this in the site configs when Intel compilers are used. The problem described in the issue so far only occurred on macOS.

configs/common/packages.yaml
kgerheiser: Yeah, that's smart. Probably want built-in MKL that comes with Intel.

.github/workflows/create-spack-mirror.yaml
```diff
@@ -0,0 +1,33 @@
...
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+        with:
+          submodules: true
+
+      - name: create-env
+        run: |
+          source ./setup.sh
+          ./create-env.py --app all --site default --name all
```
climbfuji: There is one issue here that we need to be aware of. The workflow runs for site `default`. Some sites, especially those behind firewalls, (may) have site-specific modifications of package versions. These won't be in the artifact.

kgerheiser: Added inputs to the workflow where you can specify the site and app. Defaults to `default`, and `all`.

configs/common/packages.yaml
kgerheiser: I removed the double colons. Depending on how we want to handle it we could list `intel-mkl` or `intel-oneapi-mkl` ahead of openblas and prefer those if they're available. But we can deal with that later.

climbfuji (commented):
I am ok to merge this now and work on the linalg providers afterwards. Would be good to run the github actions after this got merged and fix the errors that we are currently seeing (and any potential new ones).

kgerheiser:
Agreed

PR #59, opened 2022-03-14
Remove jedi-um-bundle-env from all

### LINKED ISSUE #57: JCSDA internal repos
@climbfuji is there something we can do with these internal repos? Like put them in their own bundle until if/when they become public.

kgerheiser:
jedi-um-bundle-env contains internal JCSDA repos

Fix #57

@climbfuji how is this for a workaround?


climbfuji (approved):
That's a workable solution.

PR #65, opened 2022-03-16
Bugfixes for building esmf and mapl in release and debug mode; fix mapl module names; updates for gaea and macos

### LINKED ISSUE #64: ESMF and MAPL only get built in debug mode, not in release mode; mapl module names are wrong
Currently ESMF and MAPL only get built in debug mode when installing the ufs-weather-model app (or the ufs-weather-model-env). This is because a single meta package cannot rely on two different versions of the same package.

Also, the module names (tcl, lmod) for mapl are off: `mapl/2.8.1-debug-esmf-8.2.0-debug-debug-esmf-8.2.0-debug` - this should be `mapl/2.8.1-debug-esmf-8.2.0-debug` for the debug version, and `mapl/2.8.1-esmf-8.2.0` for the release version.

Note that in order to have a consistent set of modules, mapl must be built in debug mode when esmf+debug is loaded, this is currently inconsistent in the ufs-weather-model / hpc-stack.

climbfuji:
This PR contains a few changes:
- Reorganize apps and meta packages, in particular introduce a ufs-weather-model-debug-env so that both the release and the debug versions of esmf and mapl get built
    - Note that in order to have a consistent set of modules, mapl must be built in debug mode when esmf+debug is loaded, this is currently inconsistent in the ufs-weather-model / hpc-stack
- Fix the mapl tcl/lmod module names
- Update the macos README.md with a known issue and workaround
- Update the gaea site config to use `CRAYPE_LINKTYPE='dynamic'`
- Clean up legacy comments the common `packages.yaml` now that Cheyenne and Gaea environments work without having to make changes to netcdf and other packages

Fixes https://github.com/NOAA-EMC/spack-stack/issues/64.


configs/repos/ufs/packages/ufs-weather-model-debug-env/package.py
```diff
@@ -0,0 +1,22 @@
...
+class UfsWeatherModelDebugEnv(BundlePackage):
+    """Development environment for ufs-weathermodel-bundle"""
+
+    homepage = "https://github.com/ufs-community/ufs-weather-model"
+    git      = "https://github.com/ufs-community/ufs-weather-model.git"
+
+    maintainers = ['kgerheiser', 'climbfuji']
+
+    version('main', branch='main')
+
```
climbfuji, line 20: @kgerheiser I leave it up to you whether you want to repeat what is in `ufs-weather-model-env` except the esmf-debug and mapl-debug lines or not. If the rest is repeated, a user can install only the `ufs-weather-model-debug-env` or the `ufs-weather-model-env`. If only esmf+debug and mapl+debug are listed here, then the user needs to install both to get the debug version. I prefer the latter, because most users just want everything to run the ufs-weather-model, in both debug and release mode.

kgerheiser: I like it like this

kgerheiser (approved):

PR #68, opened 2022-03-17
Update compilers on Orion with GCC path

### LINKED ISSUE #56: Static NetCDF being built on Orion because linker doesn't support shared libraries
I think we've seen this before, but I don't know why. When I attempt to build on Orion, NetCDF-Fortran fails because NetCDF-C is built statically.

When looking at the config.log for NetCDF it says `checking if libtool supports shared libraries... no`.

Same error as https://github.com/spack/spack/issues/29354

kgerheiser:
* Set PATH for GCC
* Lower the number of build jobs
* Remove rpath check, it should be possible to use rpath without raising exception in meta modules

Fix #56

### DIFF
--- a/configs/sites/orion/compilers.yaml
+++ b/configs/sites/orion/compilers.yaml
@@ -19,14 +19,15 @@ compilers:
       cxx: /apps/intel-2021.2/intel-2021.2/compiler/2021.2.0/linux/bin/intel64/icpc
       f77: /apps/intel-2021.2/intel-2021.2/compiler/2021.2.0/linux/bin/intel64/ifort
       fc: /apps/intel-2021.2/intel-2021.2/compiler/2021.2.0/linux/bin/intel64/ifort
-    flags:
-      cxxflags: '-gxx-name=/apps/gcc-10.2.0/gcc-10.2.0/bin/g++ -Wl,-rpath,/apps/gcc-10.2.0/gcc-10.2.0/lib64'
-      ldflags: '-gxx-name=/apps/gcc-10.2.0/gcc-10.2.0/bin/g++ -Wl,-rpath,/apps/gcc-10.2.0/gcc-10.2.0/lib64'
+    flags: {}
     operating_system: centos7
     target: x86_64
     modules:
     - intel/2021.2
-    environment: {}
+    environment:
+      prepend_path:
+        PATH: '/apps/gcc-10.2.0/gcc-10.2.0/bin'
+        LD_LIBRARY_PATH: '/apps/gcc-10.2.0/gcc-10.2.0/lib64:/apps/gcc-10.2.0/gcc-10.2.0/contrib/lib'
     extra_rpaths: []
 #- compiler:
 #    spec: oneapi@2021.2.0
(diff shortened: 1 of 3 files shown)


meta_modules/setup_meta_modules.py
```diff
@@ -220,10 +220,6 @@ def module_prereq_command(module_choice, module):
             substitutes['ENVVARS'] = substitutes['ENVVARS'].rstrip('\n')
             logging.debug("  ... ... ENVVARS  : {}".format(substitutes['ENVVARS']))
 
-        # Extra rpaths - not supported
```
climbfuji, line 223: I haven't tested this, but if there are problems down the road we can fix them there.

configs/sites/orion/compilers.yaml
```diff
@@ -19,14 +19,15 @@ compilers:
...
-      cxxflags: '-gxx-name=/apps/gcc-10.2.0/gcc-10.2.0/bin/g++ -Wl,-rpath,/apps/gcc-10.2.0/gcc-10.2.0/lib64'
-      ldflags: '-gxx-name=/apps/gcc-10.2.0/gcc-10.2.0/bin/g++ -Wl,-rpath,/apps/gcc-10.2.0/gcc-10.2.0/lib64'
+    flags: {}
     operating_system: centos7
     target: x86_64
     modules:
     - intel/2021.2
-    environment: {}
+    environment:
+      prepend_path:
```
climbfuji, line 28: Do you think it makes sense to add `CPATH` here?

climbfuji: Because this way CPATH makes it into the compiler meta module and other applications will be able to see the non-default gcc include files.

kgerheiser: I'm not familiar with CPATH, does LD_LIBRARY_PATH not go into the meta-module? I wasn't able to generate the meta modules because of the Cython thing.

climbfuji: `CPATH` specifies a list of directories to be searched as if specified with -I, but after any paths given with -I options on the command line. This environment variable is used regardless of which language is being preprocessed.

For example, the `crtm` module produced by spack has:
```
prepend_path("CPATH","/Users/heinzell/work/jedi-stack/spack-stack-new-joint/spack-stack-develop-20220215/envs/macos-clang-13.0.0-all-20220315/install/clang/13.0.0/crtm-2.3.0-fbk26m6/include")
```
because there are Fortran include files (`*.mod`) that some applications use.

I am not entirely sure if it is needed, but it seems safer to me to have the non-default gcc header/include files in the include search path when using/linking against code compiled with that version of gcc. I had added it to Cheyenne and it didn't hurt while building the spack-stack. I haven't tried compiling any downstream application on Cheyenne yet since we made the toolchain changes, though (but neither on orion).

kgerheiser: Sure, that sounds fine.

meta_modules/setup_meta_modules.py
kgerheiser: My reasoning is that anything that's built with an rpath is embedded into the executable/shared library directly. Anything that needed the extra rpath during the build already has it, and the meta module doesn't need to do anything about it

climbfuji: Yeah, that's how it should work - I am fine with this change.

configs/sites/orion/compilers.yaml
kgerheiser: Added CPATH

climbfuji (approved):

PR #73, opened 2022-03-17
Dynamically include common/ and site/ configs

### LINKED ISSUE #71: spack bug (?): concretizer modifying spack.yaml - related: better way to organize site config yamls?
Beta testers using the "default" site config found that the concretize step can corrupt the spack.yaml file. Specifically, it replaces variables followed by double colons (e.g. `compiler::`) with invalid quotes (`'compiler:':). If the corresponding section that sets the default compiler gets moved into `site.yaml` as it is the case for the pre-configured sites, this does not happen.

I would like to use this issue to revisit how the site configs. It is a bit confusing to me that the common config files are all copied as is, but the site configs get merged into one file. How about the site config files remain separate but are prefixed with `site_`? For example, one would then have:
```
config.yaml
install
modules.yaml
packages.yaml
site_compilers.yaml
site_config.yaml
site_modules.yaml
site_packages.yaml
spack.yaml
```

We would also need instructions how users using the "default" site config would configure it (move `spack external find` stuff from spack.yaml to the various `site_*.yaml` files?) before running the concretizer.

Looking for input and ideas ...

kgerheiser:
create-env.py will copy the copy the common/ and site/ configs and dynamically include them in the spack.yaml. 

Fix #71 

Now a spack.yaml looks like:

```
  include:
   - common/concretizer.yaml
   - common/config.yaml
   - common/packages.yaml
   - common/modules.yaml
   - site/config.yaml
   - site/packages.yaml
   - site/compilers.yaml
   - site/modules.yaml
   - ${SPACK_STACK_DIR}/configs/repos/repos.yaml
```

### DIFF
--- a/configs/apps/all/spack.yaml
+++ b/configs/apps/all/spack.yaml
@@ -1,7 +1,8 @@
 spack:
   concretization: separately
   view: false
-  include:
+  
+  @CONFIG_INCLUDES@
 
   specs:
     - jedi-fv3-bundle-env
--- a/configs/apps/all/spack.yaml
+++ b/configs/apps/all/spack.yaml
@@ -2,11 +2,6 @@ spack:
   concretization: separately
   view: false
   include:
-    - site.yaml
-    - config.yaml
-    - packages.yaml
-    - modules.yaml
-    - ${SPACK_STACK_DIR}/configs/repos/repos.yaml
 
   specs:
     - jedi-fv3-bundle-env
--- a/configs/apps/jedi-all/spack.yaml
+++ b/configs/apps/jedi-all/spack.yaml
@@ -1,7 +1,8 @@
 spack:
   concretization: separately
   view: false
-  include:
+
+  @CONFIG_INCLUDES@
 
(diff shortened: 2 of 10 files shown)


climbfuji (approved):
This is much better, thanks! I was wondering if we should use the same method as in the meta modules script. There would be a placeholder `@YAML_INCLUDES@` or so that gets replaced by an empty string (if there are no includes) or a formatted text. Not necessary, of course, but more consistent across the scripts.

Additionally, we use spack's `filter_file`  method I suppose (which the meta modules script isn't doing yet).

kgerheiser:
I took your advice and replace `@CONFIG_INCLUDES@`. I also allow `--site` to be optional, which defaults to `default` (and no `include:`), and generally cleaned up the logic and script.

kgerheiser:
Oh, forgot to include the repos.

kgerheiser:
Fixed. Tested with `--site default`, no site at all, `--exclude-common`, and regularly.

PR #75, opened 2022-03-17
Replace restricted access packages faux and trans with publicly available fiat and ectrans

### LINKED ISSUE #70: ECMWF libraries fiat and transi now publicly available
The last remaining restricted libraries from ECMWF are now publicly available: transi replaces trans, and fiat replaces faux.

https://github.com/ecmwf-ifs/fiat
https://github.com/ecmwf-ifs/ectrans

We need to make this change in our spack-stack repos and configuration files.

climbfuji:
ECMWF has released their packages fiat (replaces faux) and ectrans (replaces trans) to the public.

This PR replaces the two restricted packages with the public versions. atlas and the rest of the stack build fine with fiat and ectrans under the hood (on macOS and on hera).

Fixes #70.

### DIFF
--- a/configs/common/packages.yaml
+++ b/configs/common/packages.yaml
@@ -133,11 +133,11 @@
       variants: +fckit
     shumlib:
       version: [macos_clang_port]
-    faux:
+    fiat:
       version: [develop]
-    # Attention - when updating also check macos site config
-    trans:
+    ectrans:
       version: [develop]
+      variants: ~enable_mkl
     git-lfs:
       version: [2.11.0]
     openblas:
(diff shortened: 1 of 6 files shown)


configs/common/packages.yaml
```diff
@@ -133,11 +133,11 @@
...
     shumlib:
       version: [macos_clang_port]
-    faux:
+    fiat:
       version: [develop]
-    # Attention - when updating also check macos site config
-    trans:
+    ectrans:
       version: [develop]
+      variants: ~enable_mkl
```
climbfuji, line 140: Note. For now, build everthing with lapack and fftw. In a next step, I will add the option to use MKL instead. But do so consistently across all packages (as far as this is possible).

kgerheiser (approved):

Issue #78, opened 2022-03-18
add jedi-soca-bundle-env


shlyaeva:
I think it'd be good to have an option to build https://github.com/JCSDA-internal/soca with spack stack.

I started on trying to build soca (using `jedi-fv3-bundle-env` as a starting point), and here are a few things I discovered:
1. `jedi-soca-bundle-env` will need `nco` (some soca tests use `ncks`)

2. One of the first tests fails with:
>  3: ===============================================================================
3: Running test executable
3: ===============================================================================
3: dyld: Library not loaded: @rpath/libatlas_f.0.27.dylib
3:   Referenced from: /Users/annash/Documents/jedi/build/all-bundle/bin/soca_gridgen.x
3:   Reason: image not found
3: dyld: Library not loaded: @rpath/libatlas_f.0.27.dylib
3:   Referenced from: /Users/annash/Documents/jedi/build/all-bundle/bin/soca_gridgen.x
3:   Reason: image not found

Looks like Fortran `atlas` library isn't found? Any ideas how to fix this?

climbfuji:
@shlyaeva This is still under development, I have an environment that builds soca, tests are next ...

shlyaeva:
@climbfuji oh great, thank you!

kgerheiser:
Hmm, guessing you're running on a Mac? I think rpath is not supported on macOS.


"Control whether Spack embeds RPATH or RUNPATH attributes in ELF binaries so that they can find their dependencies. Has no effect on macOS. Two options are allowed:"

https://spack.readthedocs.io/en/latest/config_yaml.html#shared-linking

I  think we should set DYLD_LIBRARY_PATH in modules.yaml (https://github.com/NOAA-EMC/spack-stack/blob/56acc373505116037e493bb4c4d1863fff0e09b8/configs/common/modules.yaml#L153-L155)

shlyaeva:
> Hmm, guessing you're running on a Mac? I think rpath is not supported on macOS.

Yes, I am! Thank you for looking into it, I'll try with the updated develop!

climbfuji:
> I think it'd be good to have an option to build https://github.com/JCSDA-internal/soca with spack stack.
> 
> I started on trying to build soca (using `jedi-fv3-bundle-env` as a starting point), and here are a few things I discovered:
> 
> 1. `jedi-soca-bundle-env` will need `nco` (some soca tests use `ncks`)
> 2. One of the first tests fails with:
> 
[...]

PR #80, opened 2022-03-18
Set DYLD_LIBRARY_PATH in modules

kgerheiser:
macOS does not use LD_LIBRARY_PATH. We should also set DYLD_LIBRARY_PATH.

I think this will help with #78


climbfuji (approved):
Looks good, but let me note that I didn't have the problem described in #78 and I've been testing stuff for a while now.

kgerheiser:
Still a good thing to set, I think.

PR #81, opened 2022-03-18
Update install prefix to be relative to env

kgerheiser:
Use $env/install for install prefix. Prefix is being interpreted as relative to config.yaml in common/ and setup_meta_modules does not work.

See https://spack.readthedocs.io/en/latest/configuration.html#spack-specific-variables

### DIFF
--- a/meta_modules/setup_meta_modules.py
+++ b/meta_modules/setup_meta_modules.py
@@ -89,6 +89,11 @@ def module_prereq_command(module_choice, module):
         return ''
 
 def substitute_config_vars(config_str):
+    """ 
+    Substitute spack-specific and environment variables that may be present
+    in configuration files
+    See https://spack.readthedocs.io/en/latest/configuration.html#config-file-variables
+    """
     spack_vars = {'ENV': os.getenv('SPACK_ENV'),
                   'SPACK': os.getenv('SPACK_ROOT'),
                   'TEMPDIR': None,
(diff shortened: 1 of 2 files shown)


kgerheiser:
But with this, I get `Compiler module path /work/noaa/stmp/gkyle/spack-stack/envs/all.orion/$env/install/modulefiles/intel/2021.2.0 does not exist`

I think there needs to be a change in the meta module script.

climbfuji (approved):

climbfuji:
> But with this, I get `Compiler module path /work/noaa/stmp/gkyle/spack-stack/envs/all.orion/$env/install/modulefiles/intel/2021.2.0 does not exist`
> 
> I think there needs to be a change in the meta module script.

I see. See here: https://github.com/NOAA-EMC/spack-stack/blob/776621458593e5ea867b7cd5d0eb13d5526106d8/meta_modules/setup_meta_modules.py#L92 
env_dir is already known, so we need to substitute `$env` with this value.

kgerheiser:
You can use environment variables in Spack configuration files. Would it be helpful to use `${SPACK_ENV}` instead of `$env`?

climbfuji:
> You can use environment variables in Spack configuration files. Would it be helpful to use `${SPACK_ENV}` instead of `$env`?

I can work on this in a bit.

kgerheiser:
I've got a fix that substitutes correct values

climbfuji:
Great! Then I will not work on this and instead add nco ...

kgerheiser:
Think I came up with a pretty good and generic solution. Just need to test it out with setup_meta_modules.

meta_modules/setup_meta_modules.py
```diff
@@ -99,7 +124,8 @@ def module_prereq_command(module_choice, module):
 # Parse spack main config from environment
 logging.info("Parsing spack environment main config ...")
 main_config = spack.config.get('config')
-install_dir = main_config['install_tree']['root']
+install_dir = substitute_config_vars(main_config['install_tree']['root'])
```
climbfuji, line 132: Neat!

climbfuji: You may need to do the same for `module_dir` in line 172 ...

kgerheiser:
Tested it out and Orion and got the meta modules setup.

PR #82, opened 2022-03-18
Hera site config update

### LINKED ISSUE #48: Issues with Python packages requiring the installation of pkgconfig / poetry - use own miniconda Python where needed
This is a complicated issue around Python packages that require installing a newer/different version of pkconfig than the native Python installation provides. The issue in the authoritative spack repo is https://github.com/spack/spack/issues/29308.

In short, with the recent changes in spack to install using `pip` instead of `python setup.py build` etc, and an apparent bug in `pip`, it is required to have a full `poetry` (`py-poetry`) package installed.

If one tries to installs `py-poetry` in Spack, then this has a dependency on `py-crashtest`, which itself requires `py-poetry` to be installed ... a circular dependency!

The workaround is to use a Python installation that has `poetry` installed. This is easy if the Python installation on the HPC provides it (e.g. gaea), or if we use a basic miniconda installation on other HPCs and add poetry (as we do currently on hera).

Orion has the same problem, i.e. we need to provide a miniconda installation in the same way as on hera.

On macOS or Linux, it's easy, we can just use `pip` to install `poetry` outside of spack and everything works fine.

Note that one doesn't have to / cannot specify the dependency on poetry, having it in the basic Python installation is a workaround that does the magic behind the scenes.

climbfuji:
This PR updates the hera site config. This config has been tested to install fv3-bundle-env, um-bundle-env, ufs-bundle-env (which includes everything for ufs-weather-model).

Also: update a number of `README.md` files with notes regarding Python installation.

Fixes #48

### DIFF
--- a/configs/sites/hera/README.md
+++ b/configs/sites/hera/README.md
@@ -14,7 +14,6 @@ This needs to be done every time before installing packages with spack or before
 
 Note. Temporary location, this needs to be moved elsewhere.
 ```
-module purge
 module use /scratch1/BMC/gsd-hpcs/Dom.Heinzeller/spack-stack/modulefiles
 module load miniconda/3.9.7
 ````
--- a/configs/sites/hera/README.md
+++ b/configs/sites/hera/README.md
@@ -2,10 +2,19 @@
 
 ## General instructions/prerequisites
 
+## General instructions/prerequisites
+
+module purge
+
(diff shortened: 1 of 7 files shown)


climbfuji:
@kgerheiser please merge if ok with those changes, I'll be out for about an hour.

kgerheiser (approved):

PR #91, opened 2022-03-21
Bug fix in .github/actions/setup-spack-stack/action.yaml: add missing 'fi' ('endif')

climbfuji:
As the title says. This is an obvious bug fix that needs to be applied.


kgerheiser:
Nice, I was looking at this yesterday, but got distracted before fixing it.

PR #94, opened 2022-03-23
Update macOS site config to clarify the intended use

climbfuji:
- Rename macos site config to macos-monterey-llvm-clang-mpich-reference
- Add macos-monterey-gcc-openmpi-reference site config
- Clarify usage of these environments in their respective `README.md` files


kgerheiser (approved):
I think this is good. Is this building MPICH from scratch?

climbfuji:
> I think this is good. Is this building MPICH from scratch?

No, both setups use the MPI provider (mpich in one case, openmpi in the other case) from homebrew as detailed in the `README.md` files.

kgerheiser:
Right, I didn't see it because that file didn't change.

PR #97, opened 2022-03-24
Update setup_metamodules.py and fix CI tests.

climbfuji:
This PR update `setup_metamodules.py` and fixes the CI tests except for the very last step for the `ubuntu-intel-build` workflow:
```
Traceback (most recent call last):
[1056](https://github.com/climbfuji/spack-stack/runs/5699187611?check_suite_focus=true#step:3:1056)
  File "meta_modules/setup_meta_modules.py", line 395, in <module>
[1057](https://github.com/climbfuji/spack-stack/runs/5699187611?check_suite_focus=true#step:3:1057)
    raise Exception("Could not find a spack package for {}@{}:".format(mpi_name, mpi_version))
[1058](https://github.com/climbfuji/spack-stack/runs/5699187611?check_suite_focus=true#step:3:1058)
Exception: Could not find a spack package for intel-oneapi-mpi@2021.5.1:
[1059](https://github.com/climbfuji/spack-stack/runs/5699187611?check_suite_focus=true#step:3:1059)
Error: Process completed with exit code 1.
```
All other builds pass - see https://github.com/climbfuji/spack-stack/actions topmost five runs.

The main difference in `setup_metamodules.py` is that it now tries to cross-reference the `packages:all:compiler` and `packages:all:mpi` settings with the actual module tree. This allows it to determine which of the "suggestions" in the compiler and mpi list are actually in use, and it also allows specifying those suggestions without version numbers or with incomplete version numbers (e.g. `intel` for the `intel@2022.0.1` compiler, or `mpich@3` for the `mpich@3.4.2` mpi provider).

The limitations are that for mpi and python there must be packages configured in spack - this is why the `ubuntu-intel-build` fails in the last step, see my comments in changed files itself. I think that this should be fixed in a follow-up PR by adding adding a package `intel-oneapi-mpi` to the spack config. It would be nice if we could find a way to modify the compiler `LD_LIBRARY_PATH` and set `I_MPI_ROOT` via the spack command line (or sed, ...) to avoid the `--dirty` flag.

Note. Hide whitespace diffs to see what has changed in `setup_metamodules.py`.


.github/actions/setup-spack-stack/action.yaml
```diff
@@ -11,7 +11,7 @@ inputs:
   mpi:
     description: 'Which MPI flavor (openmpi, mpich, intel-oneapi-mpi)'
     required: false
-    default: 'mpich'
+    default: 'mpich@3'
```
climbfuji, line 14: Intentional to test half-way specified versions with `setup_metamodules.py`.

climbfuji:
@kgerheiser I have a draft PR here that updates the `setup_metamodules.py` script and "fixes" the CI tests. Except that we have the same Python poetry problem, which means that the envs that use ph-y5py all fail. See the latest five workflow runs here: https://github.com/climbfuji/spack-stack/actions

So either we turn off the jedi-{fv3,um,ufs} builds in CI, or we take out py-h5py in a smart way so that it's still there when you build it outside of CI, or we do the miniconda install trick in CI, too.

kgerheiser:
Can't we just install Poetry with `pip` directly with something like `python3 -m pip install poetry`? We don't even need our custom Miniconda.

.github/actions/setup-spack-stack/action.yaml
```diff
@@ -111,13 +117,14 @@ runs:
...
-        spack install
-      fi
+      # if [[ "${{ inputs.compiler }}" == "intel" ]]; then
+      #   source /opt/intel/oneapi/setvars.sh
+      #   # Pass --dirty to fix Intel compiler environment - Dom ??? I don't see --dirty ...
+      #   spack install
+      # else
+      #   spack install
+      # fi
+      spack install
```
kgerheiser: Intel libraries need to be added to LD_LIBRARY_PATH or use `--dirty`

kgerheiser:
Question: do you know how Spack knows that a Python package is already installed? Like when we install poetry with Conda, what prevents Spack from attempting to install it anyway? We don't set `buildable: False` or anything like that.

climbfuji:
> Question: do you know how Spack knows that a Python package is already installed? Like when we install poetry with Conda, what prevents Spack from attempting to install it anyway? We don't set `buildable: False` or anything like that.

I think that spack only installs py-poetry-core, it doesn't know about the Python libraries installed. As you know, this is just a dirty hack to fix the circular dependency of the latest pypi installer on poetry.

climbfuji:
@kgerheiser Getting there, macos-gcc CI tests passed for jedi-fv3. I will now run all apps for macos-gcc and the other workflows.

.github/actions/setup-spack-stack/action.yaml
```diff
@@ -111,10 +118,10 @@ runs:
       cd ${{ inputs.path }}
       source setup.sh
       spack env activate ${{ inputs.path }}/envs/${{ inputs.app }}
-      if [[ "${{ inputs.compiler }}" == "intel" ]]; then
+      if [[ "${{ inputs.compiler }}" == "intel"* ]]; then
         source /opt/intel/oneapi/setvars.sh
-        # Pass --dirty to fix Intel compiler environment
-        spack install
+        # Pass --dirty to fix Intel compiler environment to fix missing LD_LIBRARY_PATH
+        spack install --dirty
```
climbfuji, line 124: The disadvantage of this is that there is no package `intel-oneapi-mpi` configured in spack that needs to be loaded, because it's got all the environment settings in the dirty environment. This is why the `setup_metamodules.py` script is failing in the last step, it's simply not a "valid" configuration.

kgerheiser:
Looking at this now. What's the status of the builds? Everything works?

climbfuji:
> Looking at this now. What's the status of the builds? Everything works?

I updated the description of the PR - everything builds, the only thing that fails is the very last step in the `setup_metamodules.py` call for the Intel build for valid reasons, I think. We should fix the intel setup in a follow-up PR. Please check the PR description (sorry to make you read all of it again). Thanks!

kgerheiser (approved):

kgerheiser:
Shouldn't be too difficult to add the Intel config. Everything is installed in `/opt/intel/oneapi/mpi/<version>`.

And same for LMod. I can take what I added for hpc-stack and we can test the module with LMod.

I can come back with a PR for these.

kgerheiser:
And going to start a run of the CI now that it has been merged

climbfuji:
Let me try for one of the workflows, good suggestion.

PR #98, opened 2022-03-24
Add NOAA AWS cloud platform

### LINKED ISSUE #77: Port libraries to NOAA cloud
Working on porting Spack build to Azure, AWS, and Google Cloud in NOAA Parallel Works.

kgerheiser:
* Add NOAA AWS cloud platform
* Update NCO to 5.0.6
* Remove duplicate entry for py-pandas (I took 1.4.0 vs. 1.3.5, assuming newer is better)

### DIFF
--- a/configs/sites/noaa-aws/packages.yaml
+++ b/configs/sites/noaa-aws/packages.yaml
@@ -1,0 +1,187 @@
+packages:
+  all:
+    compiler:: [intel@2021.5.0]
+    providers:
+      mpi:: [intel-oneapi-mpi@2021.5.1]
+  mpi:
+    buildable: False
+  intel-oneapi-mpi:
+   externals:
+   - spec: intel-oneapi-mpi@2021.5.1%intel@2021.5.0
+     prefix: /contrib/spack-stack/apps/intel-oneapi
+     modules:
+     - impi/2022.1.2
+  autoconf:
+    externals:
+    - spec: autoconf@2.69
+      prefix: /usr
+  automake:
+    externals:
+    - spec: automake@1.13.4
+      prefix: /usr
+  bash:
+    externals:
+    - spec: bash@4.2.46
+      prefix: /usr
+  berkeley-db:
+    externals:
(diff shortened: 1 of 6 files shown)


configs/sites/noaa-aws/packages.yaml
```diff
@@ -0,0 +1,187 @@
...
+    externals:
+    - spec: lustre@2.10.8
+      prefix: /usr
+  m4:
+    externals:
+    - spec: m4@1.4.16
+      prefix: /usr
+  ncurses:
+    externals:
+    - spec: ncurses@6.3.20211021+termlib abi=6
```
climbfuji, line 124: I would keep the miniconda entry and remove the other one, to avoid that you get multiple versions of ncurses used in the stack.

configs/sites/noaa-aws/packages.yaml
```diff
@@ -0,0 +1,187 @@
...
+      prefix: /contrib/spack-stack/apps/miniconda/3.9.7
+    - spec: ncurses@6.0.20170212+termlib abi=6
+      prefix: /usr
+  openssh:
+    externals:
+    - spec: openssh@7.4p1
+      prefix: /usr
+  openssl:
+    externals:
+    - spec: openssl@1.1.1n
```
climbfuji: I would keep the miniconda entry and remove the other one, to avoid that you get multiple versions of ncurses used in the stack.

climbfuji (approved):
Two comments, but that doens't need to be changed.

configs/sites/noaa-aws/packages.yaml
kgerheiser: For OpenSSL I think it's better to use the system version. In case of any security vulnerability it's going to be updated by the OS, and maybe not my Python install.

climbfuji: Yes, makes sense. I kept the non miniconda versions in my site configs and dropped the miniconda ones.

PR #110, opened 2022-04-04
Add NOAA Cloud sites

### LINKED ISSUE #77: Port libraries to NOAA cloud

kgerheiser:
Add AWS, Azure, and Google Cloud platforms.

They all have an identical filesystem with identical compilers so the only difference between them is the external packages. Updated to the v2 images.

* Weird problem with libxml2 that fails to untar on Azure, so I set it to be external
* gftl-shared fails to download submodule (perhaps because of ancient 1.8 git), so use .tar.gz releases
* v2 image of AWS identifies as centos7 instead of amzn2.

Fix #77


climbfuji (approved):
Great, thanks for doing that!

PR #112, opened 2022-04-04
Switch to openmpi on macOS, update spack (merge spack authoritative updates as of 2022/04/04)

### LINKED ISSUE #28: Fix Intel build in Github Action
Error loading `libimf.so` on Ubuntu. Need to set `LD_LIBRARY_PATH` or `EXTRA_RPATHS` in compilers.yaml if trying to avoid using `---dirty`.

### LINKED ISSUE #109: gcc+mpich bugs with FMS and ufs-weather-model
From the release notes of FMS 2022.01 (https://github.com/NOAA-GFDL/FMS/releases/tag/2022.01):
> **Known Issues**
> The MPICH MPI implementation is unsupported when used alongside GCC 10 or 11 due to compilation issues with the mixed precision reals. MPICH can still be used to compile FMS with GCC 9 or earlier, or with other compilers.
> GCC 11.1.0 is unsupported due to compilation issues with select type. The issue appears to be resolved in later GCC releases

Also, from a recent PR made to the ufs-weather-model (https://github.com/ufs-community/ufs-weather-model/pull/1147):
> WARNING: We are currently using mpich MPI library with the gnu compilers on Hera and SGI MPT on Cheyenne. mpi_f08 module in mpich, when compiled with the current versions of the gnu compilers, has some issues and MPT does not provide mpi_f08 module at all. Which means this PR will require us to switch to OpenMPI, which will require hpc-stack to be rebuild on these two platforms. Do we want to do that? Do we want to make (working) mpi_f08 module a requirement for ufs-weather-model?

Further, the `homebrew` version of mpich on macOS doesn't support `mpi_f08`, but homebrew `open-mpi` does. I tried to compile `mpich` manually with `apple-clang@13.1.6`, and `./configure` said that Fortran 2008 is not supported.

To me, this means that we should switch to using `open-mpi` as mpi provider with `gcc`, `clang`, `apple-clang` to avoid these problems altogether.

climbfuji:
- Update submodule pointer for spack for the changes described in https://github.com/NOAA-EMC/spack/pull/50 (Update jcsda emc spack stack from authoritative 20220404).
- Switch macOS reference configs and CI tests from mpich to openmpi.
- Fix Intel CI tests - set `LD_LIBRARY_PATH` instead of using `--dirty` flag for installing

Fixes https://github.com/NOAA-EMC/spack-stack/issues/109
Fixes https://github.com/NOAA-EMC/spack-stack/issues/28

Tested on macOS Monterey with llvm clang and mpich, and also with llvm clang and openmpi.

All CI tests passed (see https://github.com/NOAA-EMC/spack-stack/actions).


.gitmodules
```diff
@@ -2,5 +2,7 @@
   path = spack
   #url = https://github.com/spack/spack
   #branch = develop
-  url = https://github.com/NOAA-EMC/spack.git
-  branch = jcsda_emc_spack_stack
+  #url = https://github.com/NOAA-EMC/spack.git
```
climbfuji: For CI testing only, will be reverted prior to merging

climbfuji: For CI testing only, will be reverted prior to merging

climbfuji:
@kgerheiser Ready for review.

.github/actions/setup-spack-stack/action.yaml
```diff
@@ -148,9 +156,32 @@ runs:
...
+        echo "        prefix: $HOME/mpi" >> ${SPACK_ENV}/spack.yaml
+      elif [[ "${{ inputs.mpi }}" == "openmpi"* ]]; then
+        openmpi_ver=${OPENMPI_VERSION}
+        echo "" >> ${SPACK_ENV}/spack.yaml
+        echo "  packages:" >> ${SPACK_ENV}/spack.yaml
+        echo "    openmpi:" >> ${SPACK_ENV}/spack.yaml
+        echo "      buildable: False" >> ${SPACK_ENV}/spack.yaml
+        echo "      externals:" >> ${SPACK_ENV}/spack.yaml
+        echo "      - spec: openmpi@${openmpi_ver}" >> ${SPACK_ENV}/spack.yaml
+        echo "        prefix: $HOME/mpi" >> ${SPACK_ENV}/spack.yaml
```
kgerheiser, line 176: Given the option of externally finding it, I would search for OpenMPI and MPICH instead of this.

.github/workflows/create-spack-mirror.yaml
```diff
@@ -33,10 +33,8 @@ jobs:
           spack add clingo
           spack concretize
           spack mirror create -a
-          # upload-artifact does not allow filenames that begin with '?'
-          rm -f cache/source_cache/autoconf/\?id=05972f49ee632cd98057a3caf82ebfb9574846da-eaa3f69
-          rm -f cache/source_cache/boost/attachment.cgi?id=356970-b6f6ce6
-          rm -f cache/source_cache/py-scipy/extern_decls.patch?id=711fe05025795e44b84233e065d240859ccae5bd-5433f60
+          # upload-artifact does not allow filenames that contain '?'
+          rm -vf `find cache/source_cache -name '*\?*'`
```
kgerheiser, line 37: Much better

.github/actions/setup-spack-stack/action.yaml
climbfuji: Good point. Now that I am using `spack external find --all`, it should be able to find it.

kgerheiser (approved):

.github/actions/setup-spack-stack/action.yaml
climbfuji: I'll try this in the next set of PRs.

PR #113, opened 2022-04-06
Re-usable MPI builds and other CI improvements

kgerheiser:
* Build MPI manually and use `spack external find` to use it. This saves on compilation time. Options are MPICH 3.4.3 and OpenMPI 4.1.3. I just took the latest stable versions of each.
* Add external entry for `intel-oneapi-mpi`. This is done manually with `cat` because you can't add an object array entry with `spack config add` 
* Fix macOS builds by using `spack config rm openssl` because while OpenSSL is installed, the development libraries aren't. Let Spack build them.
* Fix LLVM Clang not being used even though it was set. It's not in the default `PATH` have to point Spack to it in `(brew --prefix llvm)`. `apple-clang` was always being used instead. Also, LLVM Clang is installed by default, no need to install with Homebrew.
* Add `--fail-fast` to `spack install` so build failures are immediately noticed.

I will check back later, but the Ubuntu GCC build was successful and the other ones are chugging along.

### DIFF
--- a/.github/actions/setup-spack-stack/action.yaml
+++ b/.github/actions/setup-spack-stack/action.yaml
@@ -5,7 +5,7 @@ inputs:
     description: 'Which application env to build'
     required: true
   compiler:
-    description: 'Compiler (available options gcc@9, gcc@10, intel)'
+    description: 'Compiler (available options gcc@9, gcc@10, clang, apple-clang, intel)'
     required: true
     default: 'gcc@9'
   mpi:
@@ -44,7 +44,7 @@ runs:
             rm GPG-PUB-KEY-INTEL-SW-PRODUCTS.PUB
             echo "deb https://apt.repos.intel.com/oneapi all main" | sudo tee /etc/apt/sources.list.d/oneAPI.list
             sudo apt-get update
-            sudo apt-get install intel-oneapi-dev-utilities intel-oneapi-openmp intel-oneapi-compiler-fortran intel-oneapi-compiler-dpcpp-cpp-and-cpp-classic intel-oneapi-mpi-devel
+            sudo apt-get install intel-oneapi-dev-utilities intel-oneapi-openmp intel-oneapi-compiler-fortran intel-oneapi-compiler-dpcpp-cpp-and-cpp-classic
             echo "source /opt/intel/oneapi/setvars.sh" >> ~/.bash_profile
         fi
       elif [ "$RUNNER_OS" == "macOS" ]; then
@@ -59,7 +59,7 @@ runs:
     uses: actions/cache@v2
     with:
       path: ~/mpi
-      key: mpi-${{ inputs.mpi }}-${{ inputs.compiler }}.${{ runner.os }}1
+      key: mpi-${{ inputs.mpi }}-${{ inputs.compiler }}.${{ runner.os }}2
 
   # MPI takes a long time to build. Built it externally and cache it.
   - name: install-mpi
@@ -106,6 +106,8 @@ runs:
           ./configure --prefix=$HOME/mpi --enable-fortran --enable-cxx  --with-device=ch4:ofi
           make -j2
           make install
+        elif [[ "${{ inputs.mpi }}" == "intel-oneapi-mpi"* ]]; then
+           sudo apt install intel-oneapi-mpi-devel
         fi
 
   - name: setup-spack-env
--- a/.github/actions/setup-spack-stack/action.yaml
+++ b/.github/actions/setup-spack-stack/action.yaml
@@ -130,6 +130,20 @@ runs:
         spack compiler find
       fi
 
+      # No external find for intel-oneapi-mpi
+      # And no way to add object entry to list using "spack config add"
+      # Add this first so "spack config add packages:" will append to this entry
+      if [[ "${{ inputs.mpi }}" == "intel-oneapi-mpi" ]]; then
+        impi_ver=$(find /opt/intel/oneapi/mpi -maxdepth 1 -mindepth 1 -type d | xargs basename)
+        echo "" >> ${SPACK_ENV}/spack.yaml
+        echo "  packages:" >> ${SPACK_ENV}/spack.yaml
+        echo "    intel-oneapi-mpi:" >> ${SPACK_ENV}/spack.yaml
+        echo "      buildable: False" >> ${SPACK_ENV}/spack.yaml
+        echo "      externals:" >> ${SPACK_ENV}/spack.yaml
+        echo "      - spec: intel-oneapi-mpi@${impi_ver}" >> ${SPACK_ENV}/spack.yaml
+        echo "        prefix: /opt/intel/oneapi" >> ${SPACK_ENV}/spack.yaml
+      fi
+
       spack external find
 
   - name: update-bash-profile
@@ -148,20 +162,6 @@ runs:
       cd ${{ inputs.path }}
       source setup.sh
       spack env activate ${{ inputs.path }}/envs/${{ inputs.app }}
-
-      # No external find for intel-oneapi-mpi
-      # And no way to add object entry to list using "spack config add"
-      # Add this first so "spack config add packages:" will append to this entry
-      if [[ "${{ inputs.mpi }}" == "intel-oneapi-mpi" ]]; then
-        impi_ver=$(find /opt/intel/oneapi/mpi -maxdepth 1 -mindepth 1 -type d | xargs basename)
-        echo "" >> ${SPACK_ENV}/spack.yaml
-        echo "  packages:" >> ${SPACK_ENV}/spack.yaml
-        echo "    intel-oneapi-mpi:" >> ${SPACK_ENV}/spack.yaml
-        echo "      buildable: False" >> ${SPACK_ENV}/spack.yaml
-        echo "      externals:" >> ${SPACK_ENV}/spack.yaml
-        echo "      - spec: intel-oneapi-mpi@${impi_ver}" >> ${SPACK_ENV}/spack.yaml
-        echo "        prefix: /opt/intel/oneapi" >> ${SPACK_ENV}/spack.yaml
-      fi
 
       # Use external MPI to save compilation time
       spack config add "packages:mpi:buildable:False"
--- a/.github/actions/setup-spack-stack/action.yaml
+++ b/.github/actions/setup-spack-stack/action.yaml
@@ -169,7 +169,7 @@ runs:
 
       # Remove external OpenSSL entry
       # Error when wget attempts to build using external on macOS
-      spack config rm packages:openssl
+      spack config rm "packages:openssl"
 
       spack config add "packages:all:compiler:[${{ inputs.compiler }}]"
 
--- a/.github/actions/setup-spack-stack/action.yaml
+++ b/.github/actions/setup-spack-stack/action.yaml
@@ -70,7 +70,7 @@ runs:
     if: steps.cache-mpi.outputs.cache-hit != 'true'
     run: |
         echo "" >> ~/.bash_profile
-        if [[ "${ {inputs.compiler }}" == "gcc@9" ]]; then
+        if [[ "${{ inputs.compiler }}" == "gcc@9" ]]; then
           export CC=gcc-9
           export FC=gfortran-9
           export CXX=g++-9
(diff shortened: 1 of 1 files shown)


.github/actions/setup-spack-stack/action.yaml
```diff
@@ -44,32 +44,108 @@ runs:
...
+          export CXX=g++-9
+        elif [[ "${{ inputs.compiler }}" == "gcc@10"* ]]; then
+          export CC=gcc-10
+          export FC=gfortran-10
+          export CXX=g++-10
+          export FFLAGS="-fallow-argument-mismatch"
+        elif [[ "${{ inputs.compiler }}" == *"clang"* ]]; then
+          export CC=clang
+          export CXX=clang++
+          export FC=gfortran-10
```
kgerheiser, line 83: I just chose to use gcc-10 with Clang.

.github/actions/setup-spack-stack/action.yaml
```diff
@@ -44,32 +44,108 @@ runs:
...
+          export FFLAGS="-fallow-argument-mismatch"
+        fi
+
+        export MPICH_VERSION="3.4.3"
+        export OPENMPI_VERSION="4.1.3"
```
kgerheiser, line 93: These versions are hardcoded but could be passable with an optional input argument maybe?

.github/actions/setup-spack-stack/action.yaml
```diff
@@ -44,32 +44,108 @@ runs:
...
+      export "PATH=$HOME/mpi/bin:${PATH}"
+
+      # LLVM Clang not in PATH, search for it specifically
+      # Then, Fortran compilers are null, so set to gfortran
+      # Add minimal compilers just to prevent other compilers from being used by accident
+      if [[ "$RUNNER_OS" == "macOS" && ${{ inputs.compiler }} == "clang"* ]]; then
+        spack compiler add $(brew --prefix llvm)
+        sed -i".bak" "s|f77: null|f77: /usr/local/bin/gfortran-10|" ${SPACK_ENV}/spack.yaml
+        sed -i".bak" "s|fc: null|fc: /usr/local/bin/gfortran-10|" ${SPACK_ENV}/spack.yaml
+      elif [[ "${{ inputs.compiler }}" == "intel"* ]]; then
```
kgerheiser, line 138: Once again I set Clang to use gcc-10. Could query the location first.

.github/actions/setup-spack-stack/action.yaml
```diff
@@ -102,9 +186,9 @@ runs:
       spack config add "modules:default:tcl:whitelist:[${{ inputs.mpi }}]"
       spack config add "modules:default:lmod:whitelist:[${{ inputs.mpi }}]"
 
-      if [ "$RUNNER_OS" == "Linux" ]; then
+      if [[ "$RUNNER_OS" == "Linux" ]]; then
         echo ""
-      elif [ "$RUNNER_OS" == "macOS" ]; then
+      elif [[ "$RUNNER_OS" == "macOS" ]]; then
```
kgerheiser, line 201: I like double brackets in Bash for consistency

kgerheiser:
LLVM Clang build failed out, but seems to be an internal error where connection to the server was lost. Re-launching it.

kgerheiser:
@climbfuji I noticed you changed the matrix to just build the `all` app. I think we lose out on some efficiency gains by building everything monolithically. It also makes things clearer when just a single build fails. An idea I have is to build the base env  (and jedi-base) as a dependent job first and then chain the Spack installations together for efficiency.

It's also possible to generate a matrix dynamically without too much difficulty.

Also, I think we should add some automated launching of these builds now that we have them working. There are some clever things you can do like only launch when certain files have changed (i.e. don't have to wait hours to update a README).

.github/actions/setup-spack-stack/action.yaml
```diff
@@ -44,32 +44,108 @@ runs:
...
+          tar -xzf openmpi-${OPENMP_VERSION}.tar.gz
+          cd openmpi-${OPENMP_VERSION}
+          ./configure --prefix=$HOME/mpi --enable-mpi-fortran --enable-mpi-cxx
+          make -j2
+          make install
+        elif [[ "${{ inputs.mpi }}" == "mpich"* ]]; then
+          wget http://www.mpich.org/static/downloads/${MPICH_VERSION}/mpich-${MPICH_VERSION}.tar.gz
+          tar -xzf mpich-${MPICH_VERSION}.tar.gz
+          cd mpich-${MPICH_VERSION}
+          ./configure --prefix=$HOME/mpi --enable-fortran --enable-cxx  --with-device=ch4:ofi
```
kgerheiser, line 106: You have to set `ch4:ofi` or `ch4:ucx`. I don't know the difference between libfabric and ucx, so I chose this one. Probably inconsequential for a build like this.

.github/actions/setup-spack-stack/action.yaml
climbfuji: I think that's fine for now, we may want to create an issue to remind ourselves of these hardcoded versions.

.github/actions/setup-spack-stack/action.yaml
climbfuji: I had a quick look at a tech note that compares the two, https://permalink.lanl.gov/object/tr?what=info:lanl-repo/lareport/LA-UR-15-27176. `libfabric` seems to be older and the standard, and has more features than `ucx`, so I think your choice is good.

.github/actions/setup-spack-stack/action.yaml
climbfuji: If you look at my latest llvm compiler config, I also had to set `LD_LIBRARY_PATH` and `CPATH`, and possible also `RPATH` (following the equivalent instructions from homebrew to set `CPPFLAGS`, `LDFLAGS`).

.github/actions/setup-spack-stack/action.yaml
```diff
@@ -127,9 +212,9 @@ runs:
       if [[ "${{ inputs.compiler }}" == "intel"* ]]; then
         source /opt/intel/oneapi/setvars.sh
         # Pass --dirty to fix Intel compiler environment to fix missing LD_LIBRARY_PATH
-        spack install --dirty
+        spack install --dirty --fail-fast
```
climbfuji, line 225: I always wanted to add that and forgot to do it, thanks!

.github/actions/setup-spack-stack/action.yaml
climbfuji: How is that `install-mpi` step switching between native `apple-clang` and llvm `clang`?

climbfuji: Also, on macOS, you could simply use `brew install mpich` or `brew install openmp`, or does that not work?

kgerheiser: That's a good point. That needs to be corrected.

kgerheiser: You could install using brew, and same for Ubuntu with APT. But will it change compiler? Will compiler will Brew use out of gcc-9, gcc-10, Clang, and Apple-Clang? FIgure it's just better to build. It's cached, so it's built once and you forget about it.

kgerheiser:
LLVM Clang build fails now that I switched it from using apple-clang to build MPI. Spack fails to find MPICH 

`==> Warning: error detecting "mpich" from prefix /Users/runner/mpi/bin [invalid values for variant "netmod" in package "mpich": ['ofi%clang@13.0.1']`

And with `buildable: False` concretization fails

.github/actions/setup-spack-stack/action.yaml
climbfuji: Yes, I thought about that, too. In reality we either ask the user to install mpich/openmpi with homebrew, or build it with spack. But with homebrew you have the problem that mpich won't do mpi_f08, even if you use gcc down the road. So your approach is better.

kgerheiser:
This is really annoying. I don't see any reason why Spack should fail to pickup MPICH built with LLVM Clang. Why would MPICH pickup `ofi%clang@13.0.1` as the `netmod` variant? I run `mpichversion` from the runner and it shows the same thing as it does with other OS/compilers.

Ubuntu GCC (that works):

```
Setting environment variable SPACK_STACK_DIR to /home/runner/work/spack-stack/spack-stack
Sourcing spack environment /home/runner/work/spack-stack/spack-stack/spack/share/spack/setup-env.sh
Created environment jedi-ufs-all in /home/runner/work/spack-stack/spack-stack/envs/jedi-ufs-all
MPICH Version:    	3.4.3
MPICH Release date:	Thu Dec 16 11:20:57 CST 2021
MPICH Device:    	ch4:ofi
MPICH configure: 	--prefix=/home/runner/mpi --enable-fortran --enable-cxx --with-device=ch4:ofi
MPICH CC: 	gcc-9    -O2
MPICH CXX: 	g++-9   -O2
MPICH F77: 	gfortran-9   -O2
MPICH FC: 	gfortran-9   -O2
MPICH Custom Information:
```

LLVM Clang `mpichversion` (that fails):
```
MPICH Version:    	3.4.3
MPICH Release date:	Thu Dec 16 11:20:57 CST 2021
MPICH Device:    	ch4:ofi
MPICH configure: 	--prefix=/Users/runner/mpi --enable-fortran --enable-cxx --with-device=ch4:ofi
MPICH CC: 	/usr/local/opt/llvm/bin/clang    -O2
MPICH CXX: 	/usr/local/opt/llvm/bin/clang++   -O2
MPICH F77: 	gfortran-10 -fallow-argument-mismatch  -O2
MPICH FC: 	gfortran-10   -O2
MPICH Custom Information:
```

climbfuji:
> This is really annoying. I don't see any reason why Spack should fail to pickup MPICH built with LLVM Clang. Why would MPICH pickup `ofi%clang@13.0.1` as the `netmod` variant? I run `mpichversion` from the runner and it shows the same thing as it does with other OS/compilers.
> 
> Ubuntu GCC (that works):
> 
> ```
> Setting environment variable SPACK_STACK_DIR to /home/runner/work/spack-stack/spack-stack
> Sourcing spack environment /home/runner/work/spack-stack/spack-stack/spack/share/spack/setup-env.sh
> Created environment jedi-ufs-all in /home/runner/work/spack-stack/spack-stack/envs/jedi-ufs-all
> MPICH Version:    	3.4.3
> MPICH Release date:	Thu Dec 16 11:20:57 CST 2021
> MPICH Device:    	ch4:ofi
> MPICH configure: 	--prefix=/home/runner/mpi --enable-fortran --enable-cxx --with-device=ch4:ofi
> MPICH CC: 	gcc-9    -O2
> MPICH CXX: 	g++-9   -O2
> MPICH F77: 	gfortran-9   -O2
> MPICH FC: 	gfortran-9   -O2
> MPICH Custom Information:
> ```
> 
> LLVM Clang `mpichversion` (that fails):
> 
> ```
> MPICH Version:    	3.4.3
> MPICH Release date:	Thu Dec 16 11:20:57 CST 2021
> MPICH Device:    	ch4:ofi
> MPICH configure: 	--prefix=/Users/runner/mpi --enable-fortran --enable-cxx --with-device=ch4:ofi
> MPICH CC: 	/usr/local/opt/llvm/bin/clang    -O2
> MPICH CXX: 	/usr/local/opt/llvm/bin/clang++   -O2
> MPICH F77: 	gfortran-10 -fallow-argument-mismatch  -O2
> MPICH FC: 	gfortran-10   -O2
> MPICH Custom Information:
> ```

I can take a look at this later, don't worry about it. As long is it all builds to completion and we have an issue to remind us of this last missing bit, it's fine for me.

kgerheiser:
The LLVM build doesn't go until completion. It fails at finding MPICH, then because MPI is set to `buildable: False`, concretization fails.

climbfuji:
> The LLVM build doesn't go until completion. It fails at finding MPICH, then because MPI is set to `buildable: False`, concretization fails.

Ok, no problem either. I will try after this PR is merged. After all, I wanted to have the LLVM build ;-)

climbfuji (approved):

kgerheiser:
Ok, then everything else completes successfully.

kgerheiser:
There's some debug information still in the files. One way to figure out what may be happening is to probe the `determine_variants` method of MPICH's `package.py` and see why the `netmod` regex is picking up the compiler spec.

PR #114, opened 2022-04-07
Update Spack submodule to fix LLVM Clang workflow

kgerheiser:
Fixes Clang/MPICH issue in #113


climbfuji:
Aha. That's a good one. I tried last night with openmpi for curiosity, and had the same issue as you had with mpich.

climbfuji (approved):

kgerheiser:
I wouldn't expect to have the same issue with OpenMPI. It was an error with how the variant list was constructed in MPICH's package.py. There was a missing space.

kgerheiser:
You're right. The code is copy-pasted from each other and only MPICH was fixed.

PR #125, opened 2022-04-25
Fix wgrib2 build and re-enable

kgerheiser:
Fixes wgrib2 build error `cc1: error: '-Wformat-security' ignored without '-Wformat' [-Werror=format-security]`.

I see this error around the internet such as here: https://bugzilla.mozilla.org/show_bug.cgi?id=1388681

Build tested with Clang, Apple Clang, GCC, and Intel.

https://github.com/kgerheiser/spack-stack/actions/runs/2204985895

https://github.com/kgerheiser/spack-stack/actions/runs/2204985761

### DIFF
--- a/.github/actions/setup-spack-stack/action.yaml
+++ b/.github/actions/setup-spack-stack/action.yaml
@@ -82,6 +82,8 @@ runs:
         spack compiler add $(brew --prefix llvm)
         sed -i".bak" "s|f77: null|f77: /usr/local/bin/gfortran-10|" ${SPACK_ENV}/spack.yaml
         sed -i".bak" "s|fc: null|fc: /usr/local/bin/gfortran-10|" ${SPACK_ENV}/spack.yaml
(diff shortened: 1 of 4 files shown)


.github/actions/setup-spack-stack/action.yaml
```diff
@@ -219,6 +219,9 @@ runs:
       if [[ "$RUNNER_OS" == "Linux" ]]; then
         echo ""
       elif [[ "$RUNNER_OS" == "macOS" ]]; then
+        # Clang has trouble finding -lomp for OpenMP because of non-standard location
```
climbfuji, line 222: I think that it's good to be able to test non-OpenMP builds as well ... unless we want to make OpenMP support a hard requirement.

climbfuji (approved):
Looks good.

.github/actions/setup-spack-stack/action.yaml
kgerheiser: I agree

PR #128, opened 2022-04-28
Create crtm-fix package

kgerheiser:
Downloads and installs CRTM fix files. It's easier to have it as a package for the download and checksum capability rather than a variant within CRTM.

### DIFF
--- a/configs/repos/common/packages/crtm-fix/package.py
+++ b/configs/repos/common/packages/crtm-fix/package.py
@@ -15,7 +15,8 @@ class CrtmFix(Package):
 
     maintainers = ['kgerheiser']
 
-    version('2.4.0_emc', sha256='60e41f97e4b46bb84f252c9fc2e88cf765835aa94f9caa8e1d4dc3f57e3fbdf2')
+    version('2.4.0', sha256='7924285b8aa25b0b864643f03e7f281ca1816958e24c76aa9531b3ca902bf6e9')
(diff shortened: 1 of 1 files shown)


Hang-Lei-NOAA (approved):

Hang-Lei-NOAA:
Has to do so, since the crtm_emc version is still changing and the fix files has been different from the public release.

climbfuji (approved):

climbfuji:
Note: 2.4.1 was just released, together with new fix files: https://github.com/JCSDA-internal/crtm/tree/v2.4.1

kgerheiser:
Good to know. I have some changes on the CRTM side coming in as well.

kgerheiser:
@climbfuji how does JEDI/JCSDA handle CRTM? As the package is currently handled it is setup to use what is the `_emc` branches (I don't know how these differ from the non-EMC releases) for releases and we pull it from an EMC fork (https://github.com/NOAA-EMC/crtm). I'm thinking we use the JCSDA repo. And version wise I see some JEDI specific tags.

climbfuji:
> @climbfuji how does JEDI/JCSDA handle CRTM? As the package is currently handled it is setup to use what is the `_emc` branches (I don't know how these differ from the non-EMC releases) for releases and we pull it from an EMC fork (https://github.com/NOAA-EMC/crtm). I'm thinking we use the JCSDA repo. And version wise I see some JEDI specific tags.

We will be moving to the standard/official releases very soon, hopefully with 2.4.1. The special -jedi releases won't be needed in the future.

PR #131, opened 2022-05-03
Move spack-stack repos into spack submodule, use two-level namespace for openmpi and mpich, add package py-pyhdf, use only one macOS reference config

### LINKED ISSUE #127: Move spack-stack repos to builtin repo within Spack
I think we should combine our repos into a single. We only have about a dozen packages and I think it just makes it more confusing to find if a package is in common/ufs/jedi. Spack has several thousand in the builtin repo, I think we're fine with just a handful.

My proposal is to have two repos: `spack-stack` and `bundles`. All the bundle envs we have go in bundle and all the regular packages go in `spack-stack`.

### LINKED ISSUE #129: Add pyhdf package to JEDI virtual environments
Based on the discussion in https://github.com/JCSDA-internal/ioda-converters/issues/927, we need to add the `pyhdf` to some or all (tbd) JEDI virtual environments.

### LINKED ISSUE #132: Broken: CI tests using apple-clang/llvm-clang on macOS
Building `jedi-ewok-env` with `apple-clang` or (llvm) `clang` on macOS no longer works in the CI tests due to a problem with Apple's `m4` not recognizing certain flags. This used to work as recent as 10 days ago (see https://github.com/climbfuji/spack-stack/actions).

### LINKED ISSUE #135: Must build MPI with two-level namespaces on macOS to handle exceptions correctly
This problem was investigated by @srherbener, credits for identifying the issue and coming up with a solution for mpich go to him.

The problem is related to handling exceptions (unwinding the stack) when compiling with `clang` and `gfortran`. Here is the initial conversation:
```
I debugged the test_ioda_obsspace test failure (abort) and found that the wrong library got loaded for unwinding the stack after an exception was thrown. There is a function named _Unwind_Resume that gets called after an exception is thrown. The clang version of this function is in libunwind.dylib and the GNU version is in libgcc_s.dylib. These two versions are incompatible, the clang code gets linked with libunwind.dylib and the gfortran code gets linked with libgcc_s.dylib (which is what is supposed to happen), but during runtime the _Unwind_Resume function gets loaded from only libgcc_s.dylib . This is why the tests crash at the end of execution or is an exception is thrown during the test.
This can be remedied by using the two-level-namespace linker feature on the Mac. This feature puts the two _Unwind_Resume functions in separate namespaces effectively making them two different unique function names. I discovered that spack by default enables the flat-namespace linker feature on the Mac which is why the problem is exposed (ie, the two _Unwind_Resume functions have identical names). Look in the output from the spack install and you can see the -Wl,-flat_namespace linker option being used.
...
```

As it turns out, for `mpich` there is an option to compile with flat or non-flat (two-level) namespace, and spack has a variant defined. For `openmpi`, no such option or variant exists. However, I found that the two-level namespace option in `mpich` essentially does nothing else but replacing the linker flags
```
-Wl,-flat_namespace
```
with
```
-Wl,-commons,use_dylibs
```
when building `mpich` and when building applications with this `mpich`. I therefore implemented a similar variant in spack for openmpi and tested it successfully (it solves the same problems that the mpich two-level namespace variant solves).

The implementation is in https://github.com/NOAA-EMC/spack/pull/58 and the use of two-level namespaces in spack-stack in https://github.com/NOAA-EMC/spack-stack/pull/131.

climbfuji:
## Description

This PR has once again grown bigger than intended ... it:

- Finalizes the move of the `spack-stack` repos into the `spack` submodule
    - This requires adding the projections/name translations for `ecmwf-atlas` to `atlas` in the common module config
- Enables use of two-level namespace for `openmpi` and `mpich` (the former required making changes to the package config in `spack`, for the latter this was already implemented)
    - Also need to export `OMPI_MCA_rmaps_base_oversubscribe=1` in the `openmpi` module so that one can use more tasks than CPU cores (for unit testing)
- Fixes several problems with CI builds on Linux and macOS
    - Use a combination of package-level build parallelism and install-level build parallelism (see https://spack.readthedocs.io/en/latest/packaging_guide.html?highlight=parallel#parallel-builds) to speed up the Ubuntu Intel build
- Adds the existing `hdf` and new `py-pyhdf` packages
    - Requires setting the default provider for `jpeg` to `libjpeg-turbo` (already part of our spack-stack)
- Updates one of the macOS reference configs to `macos-monterey-apple-clang-openmpi-reference` (use `apple-clang` instead of (llvm) `clang`) and removes the macOS reference config for easier maintenance
    - This reference config now builds openmpi-4.1.3 built by `spack` with two-level namespace enabled
    - Note that I tested the same config with `mpich-3.4.2` built by `spack` with two-level namespace enabled
- Adds the capability to create containers with `spack`, all built into the existing scripts
    - `create-env.py` is renamed to `create.py environment ...` for creating local environments; Github unfortunately doesn't see this and thinks it is a new file and the old one got deleted (use a graphical diff tool locally to compare)
    - Containers are built using `create.py container ...`
    - One example is added: `configs/containers/docker-ubuntu-gcc-openmpi.yaml`
- Comments out `git-lfs` in the common package config, because pinning a version there isn't needed (we don't build `git-lfs`, but require it due to the dependencies on `go`/`go-bootstrap`, which are difficult to build) and having it in there breaks the container builds
- Updates and consolidates the documentation in `README.md`, `config/sites/README.md` and `config/common/README.md` into a single (top-level) `README.md`
    - Note: In my next PR, I will add the capability to create HTML (and possibly LaTeX/PDF) documentation using `sphinx`

## Dependencies
- waiting on https://github.com/NOAA-EMC/spack/pull/58
    - this updates the submodule pointer to include the changes from the previously merged PR https://github.com/NOAA-EMC/spack/pull/57 (Move spack-stack repos into Spack builtin)

## Issues

- fixes https://github.com/NOAA-EMC/spack-stack/issues/132
- fixes https://github.com/NOAA-EMC/spack-stack/issues/127
- fixes https://github.com/NOAA-EMC/spack-stack/issues/129
- fixes https://github.com/NOAA-EMC/spack-stack/issues/135

## Testing

- [x] CI tests all pass - see https://github.com/climbfuji/spack-stack/actions (Note: currently some still running, but they all passed yesterday)
- [x] macOS Monterey with `apple-clang@13.1.6` and `openmpi-4.1.3` built by `spack` for several applications (including full `ctests` for the `fv3-jedi` bundle)
- [x] macOS Monterey with `apple-clang@13.1.6` and `mpich-3.4.2` built by `spack` for several applications (including full `ctests` for the `fv3-jedi` bundle)
- [x] Container build of `docker-ubuntu-gcc-openmpi` config on macOS


.github/actions/setup-spack-stack/action.yaml
```diff
@@ -107,14 +113,34 @@ runs:
           tar -xzf openmpi-${OPENMPI_VERSION}.tar.gz
           cd openmpi-${OPENMPI_VERSION}
           # --with-hwloc=internal --with-libevent=internal : https://www.open-mpi.org/faq/?category=building#libevent-or-hwloc-errors-when-linking-fortran
-          ./configure --prefix=$HOME/mpi --enable-mpi-fortran --enable-mpi-cxx --with-hwloc=internal --with-libevent=internal
+          if [[ "$RUNNER_OS" == "macOS" ]]; then
+            ./configure --prefix=$HOME/mpi \
+                --enable-mpi-fortran --enable-mpi-cxx \
+                --with-hwloc=internal --with-libevent=internal \
+                --with-wrapper-ldflags="-Wl,-commons,use_dylibs" \
```
climbfuji, line 129: Note: These flags are equivalent to the two-level-namespace option for `mpich`.

kgerheiser:
This looks great. 

> Comments out git-lfs in the common package config, because it isn't needed (we don't build git-lfs but require it due to the dependencies on go/go-bootstrap, which are difficult to build) and having it in there breaks the container builds

JediBaseEnv has a dependency on git-lfs. Are you saying that it shouldn't have that?

.github/actions/setup-spack-stack/action.yaml
kgerheiser: Can you explain what the error is with two_level_namespace? I added the variant to the MPICH package myself, but it's just something I knew to always add to macOS MPICH flags. Something related to macOS linking and C++ and JEDI?

climbfuji:
> This looks great.
> 
> > Comments out git-lfs in the common package config, because it isn't needed (we don't build git-lfs but require it due to the dependencies on go/go-bootstrap, which are difficult to build) and having it in there breaks the container builds
> 
> JediBaseEnv has a dependency on git-lfs. Are you saying that it shouldn't have that?

Thanks for looking at this PR! Ready for "review" soon.

JediBaseEnv does have a dependency on `git-lfs`, yes, but we can't pin a version in `packages.yaml` (or have any entry in there), because of the way we piece together `configs/common/packages.yaml` and the docker config (`config/container/*`). It doesn't allow having two entries in the `packages` section (`spack containerize` ignores one of the two, which in the current order means the one from the container config).

We could fix that by stripping out the `packages:` section identifier and combine the common packages with the container packages in a different order, in which case the `containerize` command will ignore the entry in the common packages, i.e. it won't make a difference at all. Same for all regular sites, because we install it as an external package (`go` and `go-bootstrap` are really tricky to build).

.github/actions/setup-spack-stack/action.yaml
```diff
@@ -49,6 +55,15 @@ runs:
             sudo apt-get update
             sudo apt-get install intel-oneapi-dev-utilities intel-oneapi-openmp intel-oneapi-compiler-fortran intel-oneapi-compiler-dpcpp-cpp-and-cpp-classic intel-oneapi-mpi-devel
             echo "source /opt/intel/oneapi/setvars.sh" >> ~/.bash_profile
+            #
+            # DH*
+            # For now, install qt5 only for Intel to shorten the build time
+            # without having to rerun all other CI tests. Later, should consider
+            # using https://github.com/jurplel/install-qt-action for all runners
+            # and get rid of the macOS homebrew installation/package config
+            #sudo apt-get install qt5-default qttools5-dev-tools
```
climbfuji, line 64: Note. These are not yet used in spack, there was a problem during the concretization step when I added it manually as an external spack. I worked around the long build times using package-level plus install-level parallelism, see further down.

meta_modules/setup_meta_modules.py
```diff
@@ -650,3 +650,5 @@ def substitute_config_vars(config_str):
                 with open(python_module_file, 'w') as f:
                     f.write(module_content)
                 logging.info("  ... writing {}".format(python_module_file))
+
+logging.info("Metamodule generation completed successfully in {}".format(meta_module_dir))
```
climbfuji, line 654: This is nice to have, because one can copy & paste the `meta_module_dir` for the `module use` statement.

climbfuji:
@srherbener @mer-a-o FYI

.github/actions/setup-spack-stack/action.yaml
```diff
@@ -225,11 +284,26 @@ runs:
 
   - name: build-env
     shell: bash
+    env:
+      OMP_NUM_THREADS: 16
```
kgerheiser, line 288: Does this help? The runner only has 2 vcpus, and what uses OpenMP? Does Spack use it internally?

configs/common/modules.yaml
```diff
@@ -212,6 +226,10 @@ modules:
         suffixes:
           ^esmf@8.2.0~debug: 'esmf-8.2.0'
           ^esmf@8.2.0+debug: 'esmf-8.2.0-debug'
+      openmpi:
+        environment:
+          set:
+            'OMPI_MCA_rmaps_base_oversubscribe': '1'
```
kgerheiser, line 232: Nice, this was always one of the reasons I used MPICH. Having it encoded into the environment is good.

kgerheiser (approved):

climbfuji:
@kgerheiser Submodule pointer updated, `.gitmodules` reverted - ready for review and merging, if ok.

.github/actions/setup-spack-stack/action.yaml
climbfuji: OpenMP threads are used for the package-level build parallelism (`config:build_jobs`). It does help, though not a lot. At least it decreases the build time from 6hrs+ to 5hrs 25min with ubuntu/intel for jedi-ewok - good enough for the moment.

configs/common/modules.yaml
climbfuji: Problem with mpich on macOS is that once the ufs-weather-model update to use mpi_f08 is made, we can't use mpich on macOS anymore, because it doesn't build mpi_f08 on macOS. But openmpi does. So I am glad that there is a solution to (a) the oversubscribe problem and (b) the two-level namespace problem.

PR #138, opened 2022-05-09
Re-apply fix for 2.4.0_emc

kgerheiser:
I mistakenly thought this was changed on the ftp site, but turns out something else was changed.

Some files need to be manually updated because of an inconsistency in the fix files.

The file that is 'Little_Endian/amsua_metop-c_v2.SpcCoeff.bin' is the same file found in 2.3.0 as 'Big_Endian/amsua_metop-c.SpcCoeff.bin'.

And after this PR, this package should be moved into Spack like the others. This is the only here.


kgerheiser:
Doesn't really make sense to change the file and then delete it

PR #143, opened 2022-05-18
Feature/spack stack extension

kgerheiser:
Depends on https://github.com/NOAA-EMC/spack/pull/63

`./create environment/container` becomes `spack stack create env/container`. And instead of `--config` it's a positional argument as the first arg to `create container`.

The CI is running now and I was able to generate a valid container. The runs havemade it through the environment setup, so all seems well.

### DIFF
--- a/.spackstack
+++ b/.spackstack
@@ -1,1 +1,1 @@
-Hidden file for spack stack command to find spack-stack configs
+Let's spack stack command ffind top-level spack-stack directory for config location
--- a/.spackstack
+++ b/.spackstack
@@ -1,0 +1,1 @@
+Hidden file for spack stack command to find spack-stack configs
--- a/.spackstack
+++ b/.spackstack
@@ -1,1 +1,1 @@
-Let's spack stack command ffind top-level spack-stack directory for config location
+Lets spack stack command find top-level spack-stack directory for config location
--- a/README.md
+++ b/README.md
@@ -4,8 +4,8 @@ spack-stack is a collaborative effort between the NOAA Environmental Modeling Ce
 
 [spack](https://github.com/spack/spack) is a community-supported, multi-platform, Python-based package manager originally developed by the Lawrence Livermore National Laboratory (LLNL; https://computing.llnl.gov/projects/spack-hpc-package-manager). It is provided as a submodule so that a stable version can be referenced. [See the Spack Documentation for more information](https://spack.readthedocs.io/en/latest/)
 
-spack-stack is mainly a collection of Spack configuration files, but provides a few Python scripts to simplify the installation process:
-- `create.py` is provided to copy common, site-specific, and application-specific configuration files into a coherent Spack environment and to create container recipes
+spack-stack is mainly a collection of Spack configuration files, but provides a Spack extension to simplify the installation process:
+- `spack stack create` is provided to copy common, site-specific, and application-specific configuration files into a coherent Spack environment and to create container recipes
 - `meta_modules/setup_meta_modules.py` creates compiler, MPI and Python meta-modules for a convenient setup of a user environment using modules (lua and tcl)
 
 spack-stack is maintained by:
@@ -42,17 +42,17 @@ cd spack-stack
 source setup.sh
 
 # Basic usage of create.py
-./create.py -h
+spack stack create -h
 ```
 
 ### Create local environment
 ```
 # See a list of sites and apps
-./create.py environment -h
+spack stack create env -h
 
 # Create a pre-configured Spack environment in envs/<app>.<site>
 # (copies site-specific, application-specific, and common config files into the environment directory)
-./create.py environment --site hera --app jedi-fv3 --name jedi-fv3.hera
+spack stack create env --site hera --app jedi-fv3 --name jedi-fv3.hera
 
 # Activate the newly created environment
 # Optional: decorate the command line prompt using -p
@@ -81,13 +81,13 @@ spack module lmod refresh
 ### Create container
 ```
 # See a list of preconfigured containers
-./create.py container -h
+spack stack create container -h
 
 # Create container spack definition (spack.yaml) in directory envs/<spec>.<config>
-./create.py container --config docker-ubuntu-gcc-openmpi --spec esmf
+spack stack create container docker-ubuntu-gcc-openmpi --app ufs-weather-model
 
 # Descend into container environment directory
-cd envs/esmf.docker-ubuntu-gcc-openmpi
+cd envs/docker-ubuntu-gcc-openmpi.ufs-weather-model
 
 # Optionally edit config file
 emacs spack.yaml
@@ -102,7 +102,7 @@ docker run -it myimage
 ## Generating new site config
 Recommended: Start with an empty (default) site config. Then run `spack external find` to locate common external packages such as git, Perl, CMake, etc., and run `spack compiler find` to locate compilers in your path. Compilers or external packages with modules need to be added manually.
 ```
-./create.py environment --site default --app jedi-ufs --name jedi-ufs.mysite
+spack stack create env --site default --app jedi-ufs --name jedi-ufs.mysite
 
 # Descend into site config directory
 cd envs/jedi-ufs.mysite/site
--- a/configs/apps/empty/spack.yaml
+++ b/configs/apps/empty/spack.yaml
@@ -2,6 +2,6 @@ spack:
   concretization: separately
   view: false
   
-  
+  include: []
 
   specs: []
--- a/configs/apps/empty/spack.yaml
+++ b/configs/apps/empty/spack.yaml
@@ -2,6 +2,6 @@ spack:
   concretization: separately
   view: false
   
-  @CONFIG_INCLUDES@
+  
 
   specs: []
--- a/configs/apps/ufs-weather-model/spack.yaml
+++ b/configs/apps/ufs-weather-model/spack.yaml
@@ -1,9 +1,8 @@
-# Template spack.yaml for apps
 spack:
   concretization: separately
   view: false
 
-  @CONFIG_INCLUDES@
+  include: []
 
   specs:
     - ufs-weather-model-env
--- a/configs/containers/docker-ubuntu-gcc-openmpi.yaml
+++ b/configs/containers/docker-ubuntu-gcc-openmpi.yaml
@@ -22,7 +22,6 @@ spack:
       extra_rpaths: []
 
   # Basic package config from configs/common/packages.yaml
-  @PACKAGE_CONFIG@
   # Additional package config for container
     gcc:
       buildable: False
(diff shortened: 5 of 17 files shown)


configs/apps/ufs-weather-model/spack.yaml
```diff
@@ -1,9 +1,8 @@
-# Template spack.yaml for apps
 spack:
   concretization: separately
   view: false
 
-  @CONFIG_INCLUDES@
+  include: []
```
climbfuji, line 5: Why does only the ufs-weather-model have `include: []`, but none of the other apps?

configs/containers/docker-ubuntu-gcc-openmpi.yaml
```diff
@@ -114,5 +97,5 @@ spack:
 
     # Labels for the image
     labels:
-      app: "@SPEC@"
+      app: ""
```
climbfuji, line 100: Not sure if this label is needed at all (that came from the spack template for containers), but if it is, is there a way to set it to the spec?

climbfuji (commented):
This is grea, thanks also for updating the documentation! Just a couple of questions for my understanding.

configs/containers/docker-ubuntu-gcc-openmpi.yaml
kgerheiser: Yes, I meant to add this. Just a key insert.

configs/apps/ufs-weather-model/spack.yaml
kgerheiser: No good reason. Something I was playing with how the yaml keys were being inserted. If they key exists it affects the ordering of how new things are written, so for visual style I was trying things out.

configs/containers/docker-ubuntu-gcc-openmpi.yaml
kgerheiser: Fixed

configs/apps/ufs-weather-model/spack.yaml
kgerheiser: I added the include everywhere just so the includes placed near the top of the file.

climbfuji:
Great, I am going to test this on my Mac - interested to see the order of includes (and how spack overwrites settings).

kgerheiser:
That's a really good point that I should mention. The precedence from lowest to highest is base packages (i.e. `common/packages.yaml` unless overridden or excluded), then site, then what's in the `spack.yaml`. I think this is a logical way to handle it. Then you can have your base packages, and a special environment that explicitly asks for something different. by adding it to its `spack.yaml`.

climbfuji (commented):
A few questions from testing this out:
1. `spack stack env help` isn't up to date I think
2. If no site is specified it defaults to `default`, not `none`
3. Can we hide the unused/unnecessarily complex and confusing arguments for `spack stack env create`? (possibly the same for containers)
4. I noticed the site include dir is now called `site.default` (or `site.something-else`), no longer just `site`. While this is not a big problem, was this intentional? I liked the previous `site` for its simplicity.

climbfuji:
> That's a really good point that I should mention. The precedence from lowest to highest is base packages (i.e. `common/packages.yaml` unless overridden or excluded), then site, then what's in the `spack.yaml`. I think this is a logical way to handle it. Then you can have your base packages, and a special environment that explicitly asks for something different. by adding it to its `spack.yaml`.

Excellent. I checked this on my Mac, and packages such as `fms` that have an override of the OpenMP settings (`~openmp`) in the site config are configured correctly.

kgerheiser:
> spack stack env help isn't up to date I think

You're right

> If no site is specified it defaults to default, not none

What's the correct behavior here? It can either default to default, or default to none. I went back and forth.

> Can we hide the unused/unnecessarily complex and confusing arguments for spack stack env create? (possibly the same for containers)

Yes, I added some of these for some weird corner cases that are probably not necessary.

> I noticed the site include dir is now called site.default (or site.something-else), no longer just site. While this is not a big problem, was this intentional? I liked the previous site for its simplicity.

It was intentional, to make it more clear. I can just put it back to `site`.

climbfuji:
> > spack stack env help isn't up to date I think
> 
> You're right
> 
> > If no site is specified it defaults to default, not none
> 
> What's the correct behavior here? It can either default to default, or default to none. I went back and forth.

default is the correct behavior, but I think the help string or documentation isn't correct.

> 
> > Can we hide the unused/unnecessarily complex and confusing arguments for spack stack env create? (possibly the same for containers)
> 
> Yes, I added some of these for some weird corner cases that are probably not necessary.

Maybe we comment them out until we need them?

> 
> > I noticed the site include dir is now called site.default (or site.something-else), no longer just site. While this is not a big problem, was this intentional? I liked the previous site for its simplicity.
> 
> It was intentional, to make it more clear. I can just put it back to `site`.

No, that's fine, I just wanted to know.

kgerheiser:
Updated the help of container and env with a dynamic list of available options.

Removed some of the CLI args for simplicity, and combined the common argarse options.

Re-named site.<name>, to just site.

README.md
```diff
@@ -81,13 +81,13 @@ spack module lmod refresh
 ### Create container
 ```
 # See a list of preconfigured containers
-./create.py container -h
+spack stack create container -h
 
 # Create container spack definition (spack.yaml) in directory envs/<spec>.<config>
-./create.py container --config docker-ubuntu-gcc-openmpi --spec esmf
+spack stack create container docker-ubuntu-gcc-openmpi --app ufs-weather-model
```
climbfuji, line 87:
```suggestion
spack stack create container docker-ubuntu-gcc-openmpi.yaml --app ufs-weather-model
```

climbfuji: I just checked and it now wants the extension as part of the container name. This was probably not intentional.

climbfuji:
@kgerheiser The container creation spits out a nice message at the end:
```
==> Created container /Users/heinzell/scratch/spack-stack-feature-spack-stack-extension-final/envs/docker-ubuntu-gcc-openmpi.yaml.jedi-ewok
```
Would be nice if we could do the same for environments.

README.md
kgerheiser: Do you have an updated branch? This was a problem that I fixed. I can create a container with no `.yaml` suffix.

kgerheiser:
Added a message to the env create.

README.md
climbfuji: I saw you just pushed that (https://github.com/NOAA-EMC/spack/pull/63/commits/c8e99bc1a533db38b4eac7da56f1c2fbb385f816)

climbfuji (changes requested):
I am now looking at the container builds in a little more detail, and I noticed a few differences (positive or negative):
1. Negative/doesn't matter? Existing container build uses concretization `together`, now it uses `separately` and spack spits out a warning that `together` is recommended.
2. Negative. The `packages:all` section now has the problem with the double colons. I worked around this by replacing double colons with single colons when replacing @PACKAGE_CONFIG@ with the actual contents of the common `packages.yaml`. Now it is:
```
 # Basic package config from configs/common/packages.yaml
  # Additional package config for container
  packages:
    all:
      providers:
        'blas:': [openblas]
        'lapack:': [openblas]
        'fftw-api:': [fftw]
        'yacc:': [bison]
        'gl:': [opengl]
        'glu:': [openglu]
        'jpeg:': [libjpeg-turbo]
```
3. Positive: all the comments in the common `packages.yaml` are stripped out in the container `spack.yaml`, and long lists of `variants` are formatted for better readability.
4. Negative/doesn't matter? The `app` string is still empty:
```
    # Labels for the image
    labels:
      app: ''
      mpi: openmpi
```

kgerheiser:
> Negative/doesn't matter? Existing container build uses concretization together, now it uses separately and spack spits out a warning that together is recommended.

Yes, I'll fix that.

> Negative. The packages:all section now has the problem with the double colons. I worked around this by replacing double colons with single colons when replacing @PACKAGE_CONFIG@ with the actual contents of the common packages.yaml.

Yes, there is a problem when handling double colons. I'll do what you did and strip the double colons. Would be nice if Spack could handle this properly. It works everywhere else except spack.yaml (because it's not a real config file).

> Negative/doesn't matter? The app string is still empty:

I have code that handles this and it's working for me. I get:

```
 labels:
      app: ufs-weather-model
      mpi: openmpi
```

climbfuji:
> > Negative/doesn't matter? Existing container build uses concretization together, now it uses separately and spack spits out a warning that together is recommended.
> 
> Yes, I'll fix that.
> 
> > Negative. The packages:all section now has the problem with the double colons. I worked around this by replacing double colons with single colons when replacing @PACKAGE_CONFIG@ with the actual contents of the common packages.yaml.
> 
> Yes, there is a problem when handling double colons. I'll do what you did and strip the double colons. Would be nice if Spack could handle this properly. It works everywhere else except spack.yaml (because it's not a real config file).
> 
> > Negative/doesn't matter? The app string is still empty:
> 
> I have code that handles this and it's working for me. I get:
> 
> ```
>  labels:
>       app: ufs-weather-model
>       mpi: openmpi
> ```

Thanks, I'll try the labels again once everything is updated. Maybe the submodule pointer for `spack` wasn't updated to your latest code when I checked out the code 2hrs ago.

kgerheiser:
Double colons, labels, and concretization (merge precedence) has been fixed, and some unit tests added that loop over each app, and site checking that it works.

`spack unit-test -s --extension=stack`

.spackstack
```diff
@@ -0,0 +1 @@
+Let's spack stack command ffind top-level spack-stack directory for config location
```
climbfuji:
```suggestion
Lets spack stack command find top-level spack-stack directory for config location
```

configs/apps/empty/spack.yaml
```diff
@@ -2,6 +2,6 @@ spack:
   concretization: separately
   view: false
   
-  @CONFIG_INCLUDES@
+
```
climbfuji:
```suggestion
  include: []
```

climbfuji (commented):
Did the CI tests pass after the spack PR was merged?

kgerheiser:
Let's see - Ubuntu-gcc timed out at 6h on ewok (all other builds worked), and Intel failed because of some apt error.

I'll start them again because there were a decent number of changes since, and I didn't run macOS at all.

climbfuji (approved):
I'll approve this now so that you can merge when the test are passing and you are happy with it.

kgerheiser:
The GCC build works Along with the unit tests, the builds are all chugging along and no actual build settings were changed. Once the environment is created it's all Spack. 

Next up I'm going to move the meta module Python code into a custom command, which should be comparatively simple.

climbfuji:
> The GCC build works Along with the unit tests, the builds are all chugging along and no actual build settings were changed. Once the environment is created it's all Spack.
> 
> Next up I'm going to move the meta module Python code into a custom command, which should be comparatively simple.

Thanks! Yes, the meta module update and my doc update should be the last changes needed before the first release imo.

climbfuji:
I added the ability to detect qt5 to spack yesterday, we should be able to use that to speed up the ewok builds. I think qt5 gets installed already via brew / apt, but isn’t used yet (spack builds it if I am not mistaken).

PR #151, opened 2022-05-22
New EMC bundles

kgerheiser:
Add some new app environments. I just went through the modulefiles of global workflow submodules and some other repos like UFS_UTILS and WW3.

This adds:

* global-workflow-env
* ww3-env
* gldas-env
* ufs-utils-env
* gsi-env
* emc-verif-global-env
* emc-gfs-wafs-env
* ufs-srw-env

Also added some special environments:

* gfsv16.2 - production environment for gfsv16.2
* ufs-srw-public-v2 - UFS SRW release/public-v2


configs/common/packages.yaml
```diff
@@ -226,3 +226,9 @@
       variants: +python +grib2
     metplus:
       version: [4.1.0]
+    mapl:
```
climbfuji, line 229: These packages have been defined earlier in the same file

configs/apps/ufs-srw-public-v2/spack.yaml
```diff
@@ -0,0 +1,64 @@
+spack:
+  packages:
+    zlib:
+      version: [1.2.11]
```
climbfuji: Didn't we want to put those version informations into the `env` virtual packages? Or are these environments "special" enough to be treated differently? This would then apply to any release, or? WIth upcoming quarterly releases by EPIC and JCSDA, this will mean *a lot of* environments/apps. That's ok, but we should be really clear how we do the versioning so that it is consistent from spack-stack version to spack-stack version and we don't have to change it and the documentation.

.github/actions/setup-spack-stack/action.yaml
```diff
@@ -198,32 +204,12 @@ runs:
...
-
       # Spack external find is by default only looking for build-tools. Either
       # search for additional packages explicitly, or fall back to the old
       # behavior to find all external packages.
-      #spack external find 
-      #spack external find git-lfs openmpi mpich
-      spack external find --all
-
+      spack external find 
+      spack external find git-lfs openmpi mpich perl openssl python qt
```
climbfuji: we should also try to find curl (installed via homebrew on macOS, because building it fails with gcc)

configs/sites/macos.default/packages.yaml
```diff
@@ -0,0 +1,7 @@
+packages:
+  wgrib2:
+    variants: ~openmp
+  fms:
+    variants: +64bit +enable_quad_precision +gfs_phys ~openmp +pic
```
climbfuji: Since we don't use double colons, just `~openmp` should be enough.

.github/actions/setup-spack-stack/action.yaml
climbfuji: I think that qt5 and curl are not in the default search path on macOS with brew, may have to prepend PATH=...path-to-curl-bin:...path-to-qt5-bin:$PATH to this command (or export PATH correctly for this step in action.yaml)

climbfuji (changes requested):

configs/common/packages.yaml
kgerheiser: Are they? I thought they were too, but a find through this file reveals nothing. gftl-shared is there, but not MAPL or Yafyaml.

 https://github.com/NOAA-EMC/spack-stack/blob/develop/configs/common/packages.yaml

climbfuji: Same confusion on my end! Good to have each of them now. We should probably list all of them in alphabetical order (in a separate PR that does nothing else).

configs/apps/ufs-srw-public-v2/spack.yaml
kgerheiser: Hmmm, I think this is a case of a special standalone environment, and it should be handled in the `spack.yaml`. I think the apps could be cut down significantly to just special environments (or a separate directory). Most of them just add a single spec to an otherwise empty `spack.yaml`. The `-env` specifies what to build and the environment applies the necessary special versions.

In case of future releases, once everything is using Spack it would hopefully be as easy as providing an existing tag rather than having these special versions. Or it could be in a `release/public-v2` branch.

Seems nicer to add it to an environment then adding each time.

```
    depends_on('netcdf-fortran@4.5.3', when='public-v2')
    depends_on('parallelio@2.5.2', when='public-v2')
    ...
```

climbfuji: I was thinking of something like that:
```
    with when('@main'):
        if sys.platform == 'darwin':
            depends_on('libbacktrace', type='run')
        depends_on('cmake', type='run')
        depends_on('curl', type='run')
        depends_on('git', type='run')
        depends_on('hdf5', type='run')
        depends_on('nccmp', type='run')
        depends_on('netcdf-c', type='run')
        depends_on('netcdf-fortran', type='run')
        depends_on('parallel-netcdf', type='run')
        depends_on('parallelio', type='run')
        depends_on('python@3.7:')
        depends_on('szip', type='run')
        depends_on('wget', type='run')
        depends_on('zlib', type='run')

    with when('@1.0.0'):
        if sys.platform == 'darwin':
            depends_on('libbacktrace', type='run')
        depends_on('cmake@3.21:', type='run')
        depends_on('curl', type='run')
        depends_on('git', type='run')
        depends_on('hdf5@1.12.1', type='run')
        depends_on('nccmp@1.9.0.1', type='run')
        depends_on('netcdf-c@4.8.1', type='run')
        depends_on('netcdf-fortran@4.5.4', type='run')
        depends_on('parallel-netcdf@1.12.2', type='run')
        depends_on('parallelio@2.5.4', type='run')
        depends_on('python@3.7:')
        depends_on('szip', type='run')
        depends_on('wget', type='run')
        depends_on('zlib@1.2.11', type='run')
```

climbfuji: Of course this makes it long as well, but I think we need a way to go back and reproduce a certain environment/version of the stack, and if this can be done in a somewhat consistent way, that would be great.

.github/actions/setup-spack-stack/action.yaml
kgerheiser: Just going through the special handling we have in the Github and trying to reduce it if possible. Just wanted to see the errors myself. I ran into the wget/openssl thing (I think macOS just doesn't have the libopenssl-dev libraries). Also ran into the GCC/Curl thing. Do we require an external Curl library? Could also disable `dap` in NetCDF in macOS.default.

climbfuji: JEDI needs dap. We require curl on macOS to be installed via homebrew (in the reference config). It would build with clang, but not with gcc.

climbfuji: BTW I went through the same exercise with some of our users and on my laptops last week/this weekend, we really don't need many commands to run after `spack create env[ironment]` and before `spack concretize`. That can easily be included in the documentation (still working on it! ;-) )

configs/apps/ufs-srw-public-v2/spack.yaml
kgerheiser: Ok, maybe this is the way to go. The `with when` makes it slightly nicer, and it would nice to be able to `install ufs-srw-app@public-v2`. I'll add it like that and try it out.

climbfuji: I'll do the same for the jedi environments. The snippet I sent you is my suggestion for `base-env`, other than that we shouldn't have any overlap.

kgerheiser: A downside I see though is that you have to get real specific with low-level dependencies like `zlib` and `jasper` instead of just `depend_on(g2)` and let g2 handle the jasper dependency.

climbfuji: I don't think we have to. I added the `zlib` version dependency, but I don't think it is required. We would put the versions that we really care of (`g2`, `netcdf-c`, ...) in these env files and let them handle their dependencies. In case there are special restrictions on versions, such as the need to use `pythran@0.11:` because earlier versions don't build, we can keep those in the common `packages.yaml`? This file would only contain versions for those special cases and the flavors for the packages. Or we get rid of that file entirely and have all information in the `env` files?

kgerheiser: I think it'll be fine for future releases because we'll have consistent versions that are already being used, but for the `release/public-v2` the `srw_common` contains special versions of zlib, jasper, libpng.

climbfuji:
@kgerheiser - actually, I am wondering if we need those apps. Couldn't we have one single app with a list of specs to populate, and the user can provide a comma-separated list of the specs to add to the environment?

kgerheiser:
We could have one single app with a list of specs to populate, and that's what we should do. Environments should provide something unique instead of just different generic env specs. 

But there are scenarios where having a list of specs is useful. Like, we'll need a develop `spack.yaml` that'll provide multiple versions of some things. For example `ufs-weather-model-debug`, adding new specs over time, etc. Then, we use that as our master environment to sync things.

climbfuji:
Yeah, I was thinking that users could create an env and then say "I want to use `ufs-weather-model-env@1.0.0` and also `jedi-mpas-bundle-env@1.0.0`, for example.

.github/actions/setup-spack-stack/action.yaml
climbfuji: Figured we also need wget, because this doesn't build on macOS with gcc either. Currently creating a PR for `spack` to be able to detect it with `spack external find`.

kgerheiser: That's a good idea. I was just wondering thinking about that and why it wasn't a thing. Are there multiple verisons of wget like, gsed vs. BSD sed, find, etc?

climbfuji: > That's a good idea. I was just wondering thinking about that and why it wasn't a thing. Are there multiple verisons of wget like, gsed vs. BSD sed, find, etc?

No, but there is no "native" version in macOS - therefore the need to install via homebrew. PR will be ready in a minute.

kgerheiser:
> Yeah, I was thinking that users could create an env and then say "I want to use ufs-weather-model-env@1.0.0 and also jedi-mpas-bundle-env@1.0.0, for example.

What do these versions refer to? We start at 1.0.0 and go from there for each app env?

climbfuji:
> > Yeah, I was thinking that users could create an env and then say "I want to use ufs-weather-model-env@1.0.0 and also jedi-mpas-bundle-env@1.0.0, for example.
> 
> What do these versions refer to? We start at 1.0.0 and go from there for each app env?

Yes, I think so. It would be nice to keep them in sync, but if we don't it doesn't matter - every release/tagged version of spack-stack has clearly identified versions of the app/bundle-envs.

kgerheiser:
So, most of the versions come from the top-down with `packages.yaml`. There are a few exceptions (like esmf@8.3.0b09). This is more of a `spack-stack` version of base packages than it is an individual package env version. 

Listing each individual package version works for exceptions or in the case of ufs-srw-app where it's some completely different set of packages based on hpc-stack. Hypothetically, if we had all envs working with the same set of version we would use that single `packages.yaml` while leaving the `package.py` with just `depends_on(esmf)` with no version.

I think v1.0.0 should be a `spack.yaml` concretized with `v1.0.0` of the base packages. Then, new changes are added into a develop `packages.yaml` which we maintain concurrently, and at some point (when concretization changes?) we cut that and call that `v1.1.0` and then we create a develop `packages.yaml` starting at v1.1.0 and repeat.

climbfuji:
> So, most of the versions come from the top-down with `packages.yaml`. There are a few exceptions (like esmf@8.3.0b09). This is more of a `spack-stack` version of base packages than it is an individual package env version.
> 
> Listing each individual package version works for exceptions or in the case of ufs-srw-app where it's some completely different set of packages based on hpc-stack. Hypothetically, if we had all envs working with the same set of version we would use that single `packages.yaml` while leaving the `package.py` with just `depends_on(esmf)` with no version.
> 
> I think v1.0.0 should be a `spack.yaml` concretized with `v1.0.0` of the base packages. Then, new changes are added into a develop `packages.yaml` which we maintain concurrently, and at some point (when concretization changes?) we cut that and call that `v1.1.0` and then we create a develop `packages.yaml` starting at v1.1.0 and repeat.

Yes, that is basically what I had in mind, too.

kgerheiser:
I added a `--specs` option to the env create and removed most of the very similar `spack.yaml` files.

kgerheiser:
Have a new set of tests that runs on each PR. Instead of building everything it checks concretization across multiple systems/compilers. Found a bug in OpenMPI/MPICH where the wrong compiler was being used. The concretization changed from GCC to Clang because they are both in the same directory, and `%clang` was being added to the MPI spec.

kgerheiser:
This PR has a lot going on. I'm going to split it up into a bunch of smaller changes.

PR #155, opened 2022-05-25
Move MPI installation into its own Action.

kgerheiser:
Cuts out some of the noise and could be useful in other situations.


kgerheiser:
I've been doing runs with this in the other branch, and I started an Ubuntu build.

climbfuji (commented):
That looks great! Do we ave a consistency check somewhere that the Intel compiler goes with Intel MPI, and that for gcc/clang mpich or openmpi are ok?

kgerheiser:
I do actually, let me add that in first and this can go in after that.

PR #157, opened 2022-05-25
Add concretize_only test that concretizes environments without building

kgerheiser:
Concretizes environments across different compiler, OS, MPI for sanity check.

Adds a concretize_only option to setup-spack-stack

Concretizes a new jedi-ufs-all-dev environment.

Upload concretize log, build logs, and spack.yaml.


climbfuji (commented):
That app is great for testing the concretization step, but it won't build on most platforms. Anything with Intel compiler fails to build the `jedi-tools-env`, and macos+gcc can't build `jedi-ewok-env`.

I wonder if it is misleading to have that app if it only works on a subset of the platforms.

kgerheiser:
True, then there should be a conflict so the concretization fails.

Since we're going for the spec approach. I was thinking we remove most of the apps, and change that directory to `templates` and in there would be special environments. The default would be `empty` as it is now. Then, instead of specifying `app`, you specify specs which are appended to the template `spack.yaml`.

p.s. I'm pretty sure I built jedi-tools-env with Intel. I removed the conflict to test it out and it worked.

kgerheiser:
And just like the Intel test failed here.

climbfuji:
> On May 25, 2022, at 4:41 PM, Kyle Gerheiser ***@***.***> wrote:
> 
> 
> True, then there should be a conflict so the concretization fails.
> 
> Since we're going for the spec approach. I was thinking we remove most of the apps, and change that directory to templates and in there would be special environments. The default would be empty as it is now. Then, instead of specifying app, you specify specs which are appended to the template spack.yaml.
> 
> p.s. I'm pretty sure I built jedi-tools-env with Intel. I removed the conflict to test it out and it worked.
> 
That's good news. On your mac, I suppose. It didn’t build on the HPCs I tried (but it’s not needed there anyway).

climbfuji:
> On May 25, 2022, at 4:41 PM, Kyle Gerheiser ***@***.***> wrote: True, then there should be a conflict so the concretization fails. Since we're going for the spec approach. I was thinking we remove most of the apps, and change that directory to templates and in there would be special environments. The default would be empty as it is now. Then, instead of specifying app, you specify specs which are appended to the template spack.yaml. p.s. I'm pretty sure I built jedi-tools-env with Intel. I removed the conflict to test it out and it worked.
> That's good news. On your mac, I suppose. It didn’t build on the HPCs I tried (but it’s not needed there anyway).
> […](#)

How do you want to proceed?

kgerheiser:
Check this out @climbfuji https://github.com/NOAA-EMC/spack/pull/78

I use a list of specs and provide a `--specs` option instead of `--app`, and change that to `--templates.

I think this is a much cleaner way of handling it.

And I was wrong about jedi-tools, btw.

.gitmodules
```diff
@@ -1,6 +1,6 @@
 [submodule "spack"]
   path = spack
-  #url = https://github.com/spack/spack
-  #branch = develop
-  url = https://github.com/NOAA-EMC/spack
-  branch = jcsda_emc_spack_stack
+  url = https://github.com/kgerheiser/spack.git
```
climbfuji: The `.git` is coming back! Any reason we need that?

kgerheiser: No, I didn't know about that.

climbfuji (approved):

PR #173, opened 2022-06-02
New templates

### LINKED ISSUE #171: Add new envs to testing and builds
global-workflow-env adds some new dependencies with MET/METplu,s and ww3-env with METIS/ParMETIS.

These bundles were added, but are not yet built anywhere.

kgerheiser:
* gfs-v16.2 and ufs-srw-public-v2
* templates for different compiler/os to build everything

I think it's better to handle these as Spack environments than special versions within the bundle. It's more Spack-like, and gives the option for more customization within the spack.yaml. These are discrete environments that would be installed individually.

Fix #171


kgerheiser:
I did some testing, and the problem with listing the versions through packages.yaml is that those are just suggestions and Spack will often pick something different for various reasons, compared to giving a concrete version like `specs: [esmf@8.2.0]`.

configs/templates/ufs-srw-public-v2/spack.yaml
```diff
@@ -0,0 +1,47 @@
...
+    install_tree:
+      root: $env/install
+  modules:
+    default:
+      roots:
+        lmod: $env/install/modulefiles
+        tcl: $env/install/modulefiles
+      lmod:
+        projections:
+          parallelio: 'parallelio/2.5.2'
```
kgerheiser, line 12: PIO has a messed up tag with an extra underscore, and the module is installed as 2_5_2.

climbfuji:
> I did some testing, and the problem with listing the versions through packages.yaml is that those are just suggestions and Spack will often pick something different for various reasons, compared to giving a concrete version like `specs: [esmf@8.2.0]`.

Yes, therefore I think specifying the versions in the virtual packages and then having a spec of that virtual package with the corresponding virtual package version number will work better. But that of course only works for explicit dependencies in those virtual packages. Which is what we generally care for, but there are a few packages like pythran or shapely for which we need to use a specific version. This seems to work with `packages.yaml`, though.

climbfuji:
> > I did some testing, and the problem with listing the versions through packages.yaml is that those are just suggestions and Spack will often pick something different for various reasons, compared to giving a concrete version like `specs: [esmf@8.2.0]`.
> 
> Yes, therefore I think specifying the versions in the virtual packages and then having a spec of that virtual package with the corresponding virtual package version number will work better. But that of course only works for explicit dependencies in those virtual packages. Which is what we generally care for, but there are a few packages like pythran or shapely for which we need to use a specific version. This seems to work with `packages.yaml`, though.

I'll have something for you to look at later today for the ufs-weather-model-env.

climbfuji:
Also, FYI: When using something like this:
```
# spack-stack hash: 9f210ea
# spack hash: d3e41aa00a
spack:
  concretizer:
    unify: when_possible
  view: false
  include:
  - ${SPACK_STACK_DIR}/configs/common/modules.yaml
  - ${SPACK_STACK_DIR}/configs/common/config.yaml
  - site
  - packages.yaml
  specs:
  - jedi-fv3-env@1.0.0
  - ufs-weather-model-env@1.0.0+debug
  - ufs-weather-model-env@1.0.0~debug
```
the concretization step fails with
```
[jedi-all-apple-clang-openmpi-try2] heinzell@JCSDA-L-18146:~/work/spack-stack/spack-stack-package-cleanup-20220602 [brew-x86_64]> spack concretize 2>&1 | tee log.concretize
==> Warning: Missing a source id for ufs-weather-model-env@1.0.0
==> Warning: Missing a source id for base-env@1.0.0
==> Error: coerce_versions cannot be called on <class 'str'>
```
It doesn't happen when specifying no version numbers (`ufs-weather-model-env`, `ufs-weather-model-env+debug`). I'll look into it.

climbfuji (commented):
Do we really need all these templates? We've got one skylab-1.0.0 template and that should be good enough for JEDI, I think.

kgerheiser:
Maybe not, but I think it would be good to have a "build everything" environment in some form or another. Then, that led to splitting it up for each OS/compiler rather than having it tucked away in the `.github/workflows` list.

And then gfs-v16.2 and ufs-srw-env-v2 I want to have available.

I'm open to suggestions on how to split this up in a better way.

climbfuji:
I am worried about all the duplication in each of those templates. If you look at the skylab-1.0.0 template - specific for the skylab release end of this month - it lists everything that is needed (except jedi-tools-env) and is independent of compiler, OS, MPI, ... 

I think it's a good idea to have a build-them-all template, but ideally that would be independent of compiler and OS. But then, of course not everything builds with all compilers on all OSs. Hmm.

climbfuji:
We are going back and forth between templates (used to be apps) and specs. I think we need to decide on one of them and then stick with it, remove the other from the spack stack extension to avoid confusion and errors. I am ok going with templates, but we should try to remove the specs feature.

Should we also consider removing any redundant definition of packages in packages.yaml? Or would you want those to remain for the latest and greatest develop version of the envs? We'll also need the variants from that file. No good answer, but we need to agree on one or the other way so that I can hand out a beta release and instructions on Monday morning next week ...

kgerheiser:
> We are going back and forth between templates (used to be apps) and specs. I think we need to decide on one of them and then stick with it, remove the other from the spack stack extension to avoid confusion and errors. I am ok going with templates, but we should try to remove the specs feature.

The idea is to mostly use templates (i.e. spack.yaml). But by default you get an empty environment with no specs which you can add specs too. So, specs aren't really necessary. The same functionality is achieved through `spack add`. 

> Should we also consider removing any redundant definition of packages in packages.yaml? Or would you want those to remain for the latest and greatest develop version of the envs? We'll also need the variants from that file. No good answer, but we need to agree on one or the other way so that I can hand out a beta release and instructions on Monday morning next week ...

What do you mean by redundant definitions?

climbfuji:
I think I understand now ... fixed configurations/releases have their versions and variants encoded in the spack.yaml, e.g. in `configs/templates/gfs-v16.2/spack.yaml`. Others have `include` statements that include `configs/common/packages.yaml`. 

Yes, this would work, but it's a bit confusing and we need to definitely explain that in the section on "Environments" if we want to do that.

configs/templates/jedi-ufs-all-linux-gcc/spack.yaml
```diff
@@ -0,0 +1,26 @@
+spack:
```
climbfuji: I would create one jedi-ufs-all template and add to the instructions/CI tests commands to remove certain spack specs that don't build on one or the other system.

Then, remove the individual jedi-ufs-all-OS-COMPILER files.

kgerheiser:
Yes, the rule is that whatever is in the template always takes precedence.

The base `packages.yaml` is always included to provide a sane set of default versions for things that you don't specify and that is always kept up to date. But you can pin specific versions by overriding it with a `packages:` section in the template.

configs/templates/jedi-ufs-all-linux-gcc/spack.yaml
kgerheiser: That makes sense. One default list with a little manual intervention is better than carrying around 4 or5 files.

climbfuji:
> Yes, the rule is that whatever is in the template always takes precedence.
> 
> The base `packages.yaml` is always included to provide a sane set of default versions for things that you don't specify and that is always kept up to date. But you can pin specific versions by overriding it with a `packages:` section in the template.

Yay. We have a compromise :-)

.github/workflows/concretize.yaml
```diff
@@ -35,14 +26,34 @@ jobs:
       - name: install-env
         uses: ./.github/actions/setup-spack-stack
         with:
-          specs: ${{ env.mac_os_specs }}
+          template: jedi-ufs-all-macos-clang
```
climbfuji: The various template entries will need to be changed, but presumably you are  still working on this PR.

.github/workflows/concretize.yaml
```diff
@@ -35,14 +26,34 @@ jobs:
...
+      - name: checkout
+        uses: actions/checkout@v2
+        with:
+          submodules: true
+
+      - name: install-env
+        uses: ./.github/actions/setup-spack-stack
+        with:
+          template: jedi-ufs-all
+          compiler: ${{ matrix.compiler }}
```
climbfuji, line 51: I like that you are doing this for two compiler versions (same for clang and apple-clang above)!

climbfuji (approved):

climbfuji:
@kgerheiser Seems that the CI test is waiting for some concretize tests that are no longer run.

PR #186, opened 2022-06-07
Cleanup CI scripts and speed up CI tests

### LINKED ISSUE #100: Shorten run time of CI tests
We can shorten the run time of the CI tests by installing the MPI library using `apt` on Linux or `brew` on macOS.

It would be good to keep one workflow where the MPI library is built by spack to make sure this setup works, unless we decide to not support any configuration where spack builds the MPI library.

### LINKED ISSUE #134: CI tests on ubuntu w/ Intel are hitting build time limit
After adding `hdf` and `py-hdf` to the packages, the build time for the `jedi-ewok` app increased from 5h 45min to 6h+. The maximum build time for a single build on GitHub Actions is 6h, thus the build is canceled close to the end.

An idea to speed up the build is to increase the number of OpenMP threads and spack `build_jobs`. Or to install certain packages like `qt@5`, which we require to be an external package on macOS and all HPCs, using `apt-get`.

### LINKED ISSUE #137: CI tests on Linux: use external qt@5
In #131 `qt@5` is installed via `apt-get`, but using it as external module gave errors in the concretization step. We are therefore still building `qt@5` as part of spack (takes 2h+ with Intel).

We should consider using https://github.com/jurplel/install-qt-action for all runners in the future, and get rid of the macOS homebrew installation/package config.

climbfuji:
Cleanup CI scripts and speed up CI builds by:
- building in parallel
- using external qt@5
- using the correct site config (`linux.default` or `macos.default`)

CI tests all pass, see https://github.com/climbfuji/spack-stack/actions entry `ubuntu-intel-build #46` and above.

Fixes https://github.com/NOAA-EMC/spack-stack/issues/137
Fixes https://github.com/NOAA-EMC/spack-stack/issues/134
Fixes https://github.com/NOAA-EMC/spack-stack/issues/100

### DIFF
--- a/.github/actions/setup-spack-stack/action.yaml
+++ b/.github/actions/setup-spack-stack/action.yaml
@@ -59,6 +59,14 @@ runs:
         # Install krb5 to avoid compilation errors of "hdf" on Ubuntu
         sudo apt-get install krb5-user libkrb5-dev
 
+        # DH*
+        # For now, install qt5 using apt on Linux. Later, should consider
+        # using https://github.com/jurplel/install-qt-action for all runners
+        # and also get rid of the macOS homebrew installation/package config
+        sudo apt-get install qt5-default qttools5-dev-tools
+        #sudo apt-get install qt5-default
+        # *DH
+
         if [[ "${{ inputs.compiler }}" == "intel"* ]]; then
             cd /tmp
             wget https://apt.repos.intel.com/intel-gpg-keys/GPG-PUB-KEY-INTEL-SW-PRODUCTS.PUB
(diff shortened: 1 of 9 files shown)


kgerheiser:
I like this. The external things I find are limited to: `spack external find python perl cmake git-lfs git curl wget ninja m4` some of which are optional.

kgerheiser:
I found the build times for various packages and the majority comes from these (on a random macOS run from yesterday):

name | time (minutes)
-- | --
openssl | 3
eccodes | 3
numpy | 4
eckit | 5
hdf5 | 6
met | 6
pandas | 10
cdo | 10
fftw | 12
boost | 12
ecmwf-atlas | 13
openblas | 14
scipy | 18
esmf | 21
ecflow | 36

climbfuji:
> I like this. The external things I find are limited to: `spack external find python perl cmake git-lfs git curl wget ninja m4` some of which are optional.

Not exactly, qt and texlive are now also detectable (I added that functionality to the qt and texlive packages in a previous PR). Building qt itself takes long, and it has a lot of dependencies that are not needed otherwise, same for texlive.

climbfuji:
I am still fighting with the CI tests, this time the error occurred at the very last step of creating modules, so almost there ...

kgerheiser (approved):
Much cleaner

Issue #50, opened 2022-03-08
Reduce config duplication between tcl and lmod


climbfuji:
PR https://github.com/NOAA-EMC/spack-stack/pull/38 introduces tcl modules in addition to lmod/lua modules, with a substantial amount of duplication of the configuration for the modules. It may be possible to remove that duplication.

climbfuji:
@kgerheiser I tried moving the shared sections between `tcl` and `lmod` into the `default` section, unfortunately spack doesn't like that. Looks like we have to live with the duplication, unless we want to support only `tcl` modules (I don't want to do that, because `lmod` modules are more powerful), or only support `lmod` modules (I don't want to do that either, because some systems don't support `lmod`).

climbfuji:
Related to #703

climbfuji:
Closed via #796

PR #193, opened 2022-06-10
Add module vars

kgerheiser:
* NCEPLIBS module variables that are used by some applications.
* Update MAPL prefix for ESMF 8.3.0


climbfuji (commented):
Do we need to do the same for the `tcl` module section?

Note, I thought that in a later PR (post release 1), we could use an include statement for both `lmod` and `tcl` for the common pieces?

kgerheiser:
Ugh, do I have to repeat the same text for that?

climbfuji:
> Ugh, do I have to repeat the same text for that?

Unfortunately, yes - but maybe not 100%, but almost. Gotta be careful, because some of the module labeling rules are different for `tcl`. That's why I'd like to have the common sections in a modules_common.yaml that get included in both the `lmod` and `tcl` section. If you like to do this in this PR, might as well do it and fix the open issue for that (#50).

kgerheiser:
I duplicated the vars. Let me split this up if I'm going to factor out the common vars. It'll take a little Spack/Python yaml work for a `modules_common.yaml`.

climbfuji (commented):
Thanks! Should we merge this in the meanwhile and add your logic in a follow-up PR, or should we wait update this PR?

kgerheiser:
I would rather merge this, and add it in a follow-up PR

climbfuji (approved):

PR #199, opened 2022-06-13
Create Miniconda setup script

kgerheiser:
Setting up Miniconda is kind of tedious, so I created this script to automate the download/install/modulefile as described in the [instructions](https://github.com/NOAA-EMC/spack-stack/blob/develop/doc/source/Prerequisites.rst).

Now you just have to run `./bootstrap.sh -p <dir>` and the script will create a standard directory structure, then download Miniconda, install it along with Poetry, etc, and then create a modulefile.


bootstrap/bootstrap.sh
```diff
@@ -0,0 +1,85 @@
...
+############################################################
+# Help                                                     #
+############################################################
+Help()
+{
+   # Display Help
+   echo "Bootstrap spack-stack with Miniconda installation."
+   echo
+   echo "options:"
+   echo "-p     spack-stack installation prefix"
```
climbfuji, line 14: Should we follow the conventions and use `--prefix` instead/in addition?

bootstrap/bootstrap.sh
```diff
@@ -0,0 +1,85 @@
...
+}
+
+prefix=""
+miniconda_ver="py39_4.12.0"
```
climbfuji, line 21: Good to pin the version! Is that `python-3.9.7`?

climbfuji (commented):
I like that!

What is the intended directory structure for this setup?

If I read the script correctly, it not only sets up the directory tree for miniconda, but also imposes where envs, mirrors, ... go. Where would you check out spack-stack? In `src`?

bootstrap/bootstrap.sh
kgerheiser: Sure

bootstrap/bootstrap.sh
kgerheiser: It's actually 3.9.12. They don't encode the minor version.

kgerheiser:
> What is the intended directory structure for this setup?

I'm getting a little ahead of myself with these extra directories like mirrors.

I envision a structure like:

```
<prefix>
  /mirrors
  /envs
  /apps
  /modulefiles
  /src
```

And yeah I was thinking `spack-stack` would go in src or maybe `src/spack-stack/<env-name>`  because you may have multiple sources.

bootstrap/bootstrap.sh
kgerheiser: getopts does not support long args. I would just leave it as `-p`. There's only a single argument.

climbfuji: Ok

bootstrap/bootstrap.sh
climbfuji: Thanks!

bootstrap/templates/miniconda
```diff
@@ -0,0 +1,22 @@
...
+# Set this value
+set MINICONDA_PATH "@MINICONDA_PREFIX@"
+
+prepend-path PATH "${MINICONDA_PATH}/bin"
+prepend-path MANPATH "${MINICONDA_PATH}/share/man"
+prepend-path LD_LIBRARY_PATH "${MINICONDA_PATH}/lib"
+prepend-path LIBRARY_PATH "${MINICONDA_PATH}/lib"
+prepend-path CPATH "${MINICONDA_PATH}/include"
+prepend-path CMAKE_PREFIX_PATH "${MINICONDA_PATH}"
+prepend-path PYTHONPATH "${MINICONDA_PATH}/lib/python3.9/site-packages"
```
climbfuji, line 22: We'll need to take care to update this when we update the Python version. Maybe put a comment above the line where you specify the miniconda version you want to use?

climbfuji (commented):
I am happy with this script, makes life a lot easier. But I would prefer to agree on (and write down somewhere) the target directory structure before we merge this changes, if that's ok.

kgerheiser:
Updated the documentation with a directory structure section

climbfuji:
> Updated the documentation with a directory structure section

I was going to review that, but I can't see the doc update, unfortunately.

kgerheiser:
Fixed, changes weren't committed.

bootstrap/templates/git-lfs
```diff
@@ -0,0 +1,15 @@
+#%Module1.0
```
climbfuji: If you are moving the git-lfs template, then you also need to update the documentation. I recommend moving only the miniconda template for now. In the original location doc/modulefile_templates, there are now 5 templates altogether (ecflow, git-lfs, miniconda, openmpi, qt). I would only move the miniconda template and check if there is any documentation that needs to be updated.

doc/source/MaintainersSection.rst
```diff
@@ -296,3 +296,18 @@ Python packages can be added in various ways:
...
+==============================
+Recommended Directory Layout
+==============================
+
+To support multiple environments installs it is recommended to use a standard directory layout.
+
+After running `bootstrap.sh` the prefix will have the following directories:
+
+* apps - Miniconda and other prerequisites installations.
+* modulefiles - Prerquisite modules not built with Spack.
```
climbfuji:
```suggestion
* modulefiles - Prerequisite modules not built with Spack.
```
```suggestion
* modulefiles - Prerquisite modules not built with Spack.
```

bootstrap/templates/git-lfs
kgerheiser: I had added the ability to download git-lfs as well, but dropped it and then these new modules were added in the meantime. I removed git-lfs from this directory.

climbfuji (commented):
I am a little worried about the directory structure. If we want to be able to install major spack-stack versions in separate directories, then it will be difficult to identify which environment in the `envs` directory corresponds to which `spack-stack` version in the `src` directory.

If, on the other hand, we decided to make the top-level directory (where `apps` etc resides in) dependent on the spack-stack version, then we end up with a lot of duplicates for the apps/modulefiles that are generated outside of spack-stack, then we end up with a lot of redundancy.

kgerheiser:
I have had the same thoughts about grouping environments and having multiple spack-stack sources. What if we add another directory in `envs` that corresponds to the spack-stack version?

```
envs/
   v1.0.0/
      skylab-1.0.0
      another-env
   v2.0.0
       skylab-2.0.0
...
src/
   spack-stack-v1.0.0
   spack-stack-v2.0.0
```

Then the src and envs dir correspond.

kgerheiser:
Or 

```
envs/
   v1.0.0/
      src/spack-stack
      skylab-1.0.0
      another-env
   v2.0.0
       src/spack-stack
       skylab-2.0.0
```

Then you can have multiple environments with a single spack-stack source

climbfuji:
> I have had the same thoughts about grouping environments and having multiple spack-stack sources. What if we add another directory in `envs` that corresponds to the spack-stack version?
> 
> ```
> envs/
>    v1.0.0/
>       skylab-1.0.0
>       another-env
>    v2.0.0
>        skylab-2.0.0
> ...
> src/
>    spack-stack-v1.0.0
>    spack-stack-v2.0.0
> ```
> 
> Then the src and envs dir correspond.

Yes, I had similar thoughts. How do we handle minor versions that get installed into the existing environments (or do we not want to do that anymore)? Should those directories in `envs` just be called `spack-stack-v1`/`spack-stack-1` and everything that is `spack-stack-1.x.y` gets installed underneath there?

Or be super pragmatic and say that every new version, whether it is major/minor/patch, requires a full new build of every environment?

kgerheiser:
> Should those directories in envs just be called spack-stack-v1/spack-stack-1 and everything that is spack-stack-1.x.y gets installed underneath there?

I like that

> Or be super pragmatic and say that every new version, whether it is major/minor/patch, requires a full new build of every environment?

We could make use of chained installations for minor updates. A v1.1.0 uses v1.0.0 as an upstream. Then, we don't have to wait hours for a minor update and we don't have to do a full re-install until we get to v2.0.0 (which let's say has a new NetCDF)

climbfuji:
Problem with the chained installs it that it doesn't create the modules for the upstream repos, means that users have to use two `module use` commands, and it doesn't integrate with our meta modules either.

kgerheiser:
I like your idea of rolling releases every x months or so. I say we can append versions to an existing install during that stack's lifetime.

Then, in the next stack we try to include any one-off installs as the new base version, if possible. Like someone says we need ESMF 8.4.0, we install that alongside 8.3.0 in spack-stack 1.1.0, and then when the next release comes around we install 8.4.0 as the default and drop 8.3.0.

And this information should be encapsulated in a `spack.yaml`.

climbfuji:
@kgerheiser So is this what we want?
```
envs/
   v1/
      skylab-1.0.0
      another-env
   v2/
      skylab-2.0.0
...
src/
   spack-stack-v1.0.0
   spack-stack-v1.0.1 # updates envs/v1/...
   spack-stack-v1.1.0 # updates envs/v1/...
   ...
   spack-stack-v2.0.0
   spack-stack-v2.1.0 # updates envs/v2/...
```

kgerheiser:
How about this @climbfuji?

kgerheiser:
@climbfuji could you review my new changes? I updated the docs based on our previous conversation and added an example.

climbfuji (approved):
I think that's great and we should start working with this from the next major spack-stack release on (v2?) to not break the current directory structure for 1.0.0/1.0.1 (?).

kgerheiser:
I can install it like this on Hera/Jet starting now, right?  The directory structure is an implementation detail for users.

climbfuji:
> I can install it like this on Hera/Jet starting now, right? The directory structure is an implementation detail for users.

I was hoping to do that from version 2 on, so that the version 1 installs are identical everywhere.

climbfuji:
> > I can install it like this on Hera/Jet starting now, right? The directory structure is an implementation detail for users.
> 
> I was hoping to do that from version 2 on, so that the version 1 installs are identical everywhere.

But it's ok if Discover and Orion are different for v1

kgerheiser:
I can install the skylab-1.0.0 release using the same layout for consistency

climbfuji:
> I can install the skylab-1.0.0 release using the same layout for consistency

Thanks, Kyle. I'll do the same on Cheyenne.

PR #202, opened 2022-06-13
Add some notes and examples to the MacOS build instructions.

srherbener:
This PR includes some additional information based on trying out the MacOS build instructions on my iMac platform.

### DIFF
--- a/doc/source/Platforms.rst
+++ b/doc/source/Platforms.rst
@@ -342,7 +342,6 @@ This instructions are meant to be a reference that users can follow to set up th
    brew install python
    brew install git
    brew install git-lfs
-   brew install git-flow-avh
    brew install lmod
    brew install wget
    brew install bash
--- a/doc/source/Platforms.rst
+++ b/doc/source/Platforms.rst
@@ -350,27 +350,25 @@ This instructions are meant to be a reference that users can follow to set up th
    brew install cmake
    brew install openssl
    # Note - need to pin to version 5
-   brew install qt@5.15.3
+   brew install qt@5
 
-.. warning::
-  The :code:`brew install qt@5.15.3` command may not successfully run, in which case you can try :code:`brew install qt@5` instead.
+4. Configure your terminal to use the homebrew installed bash
 
-.. note::
   After installing bash with homebrew, you need to change your terminal application's default command to use :code:`/usr/local/bin/bash`.
   For example with iterm2, you can click on the :code:`preferences` item in the :code:`iTerm2` menu.
   Then click on the :code:`Profiles` tab and enter :code:`/usr/local/bin/bash` in the :code:`Command` box.
   This is done to avoid issues with the MacOS System Integrity Protection (SIP) mechanism when running bash scripts.
   See https://support.apple.com/en-us/HT204899 for more details about SIP.
 
-4. Activate the ``lua`` module environment
+5. Activate the ``lua`` module environment
 
 .. code-block:: console
 
    source /usr/local/opt/lmod/init/profile
 
-5. Install xquartz using the provided binary at https://www.xquartz.org. This is required for forwarding of remote X displays, and for displaying the ``ecflow`` GUI, amongst others.
+6. Install xquartz using the provided binary at https://www.xquartz.org. This is required for forwarding of remote X displays, and for displaying the ``ecflow`` GUI, amongst others.
 
-6. Temporary workaround for pip installs in spack (see https://github.com/spack/spack/issues/29308). Make sure that ``python3`` points to the Homebrew version.
+7. Temporary workaround for pip installs in spack (see https://github.com/spack/spack/issues/29308). Make sure that ``python3`` points to the Homebrew version.
 
 .. code-block:: console
 
@@ -378,7 +376,7 @@ This instructions are meant to be a reference that users can follow to set up th
    # test - successful if no output
    python3 -c "import poetry"
 
-7. Optional: Install MacTeX if planning to build the ``jedi-tools`` environment with LaTeX/PDF support
+8. Optional: Install MacTeX if planning to build the ``jedi-tools`` environment with LaTeX/PDF support
 
    If the ``jedi-tools`` application is built with variant ``+latex`` to enable building LaTeX/PDF documentation, install MacTeX 
    `MacTeX  <https://www.tug.org/mactex>`_ and configure your shell to have it in the search path, for example:
@@ -399,10 +397,6 @@ Remember to activate the ``lua`` module environment and have MacTeX in your sear
 .. code-block:: console
 
    spack stack create env --site macos.default [--template jedi-ufs-all] --name jedi-ufs.mymacos
-
-.. note::
-  It is advisable to use the :code:`--template jedi-ufs-all` option on the :code:`spack stack create env` command above.
-  This will add a number of specs to your :code:`envs/jedi-ufs.mymacos/spack.yaml` file, of which you can delete the ones you don't need.
 
 2. Temporarily set environment variable ``SPACK_SYSTEM_CONFIG_PATH`` to modify site config files in ``envs/jedi-ufs.mymacos/site``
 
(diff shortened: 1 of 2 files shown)


doc/source/Platforms.rst
```diff
@@ -351,6 +352,16 @@ This instructions are meant to be a reference that users can follow to set up th
    # Note - need to pin to version 5
    brew install qt@5.15.3
 
+.. warning::
```
climbfuji, line 355: Should we just say `brew install qt@5` instead?

climbfuji: Should we just say `brew install qt@5` instead in line 353?

doc/source/Platforms.rst
```diff
@@ -351,6 +352,16 @@ This instructions are meant to be a reference that users can follow to set up th
...
+  The :code:`brew install qt@5.15.3` command may not successfully run, in which case you can try :code:`brew install qt@5` instead.
+
+.. note::
```
climbfuji: Let's make this item 4. and update the numbers on all the following items?

srherbener:
@kgerheiser I am a software engineer in JCSDA working on @climbfuji's JEDI infrastructure team. I am helping test the spack-stack release candidate and want to add a few things to the MacOS instructions that I found during my testing.

doc/source/Platforms.rst
```diff
@@ -389,6 +400,10 @@ Remember to activate the ``lua`` module environment and have MacTeX in your sear
 
    spack stack create env --site macos.default [--template jedi-ufs-all] --name jedi-ufs.mymacos
 
+.. note::
+  It is advisable to use the :code:`--template jedi-ufs-all` option on the :code:`spack stack create env` command above.
```
climbfuji: Hmmm ... spack-stack is not specific to JCSDA or JEDI. Some people may use it to build UFS_UTILS or GSI or just ESMF or whatever else, and never touch any JEDI code. I suggest to add this type of information to our jedi-docs documentation, similar to what I am trying to collect in the working google doc now.

Once the page on `Environments` with package descriptions, templates etc. are updated, it will hopefully become easier to figure out what users can decide to build. It's easier to start with a small setup and add packages later on if needed, I think.

doc/source/Platforms.rst
```diff
@@ -342,6 +342,7 @@ This instructions are meant to be a reference that users can follow to set up th
    brew install python
    brew install git
    brew install git-lfs
+   brew install git-flow-avh
```
kgerheiser: What is that?

doc/source/Platforms.rst
```diff
@@ -351,6 +352,16 @@ This instructions are meant to be a reference that users can follow to set up th
    # Note - need to pin to version 5
    brew install qt@5.15.3
 
+.. warning::
+  The :code:`brew install qt@5.15.3` command may not successfully run, in which case you can try :code:`brew install qt@5` instead.
```
kgerheiser: Would it be better to just say `qt@5` and remove the reference to `qt@5.15.3`?

doc/source/Platforms.rst
```diff
@@ -443,11 +458,25 @@ Remember to activate the ``lua`` module environment and have MacTeX in your sear
...
+  When editing the :code:`envs/jedi-ufs.mymacos/spack.yaml` make sure that under the :code:`specs:` section you enter :code:`openmpi` and at least one of the jedi specs such as :code:`jedi-ufs-env`.
+  Here is an example:
+  
+  .. code-block:: yaml
+  
+    specs:
+        - jedi-ufs-env
+        - openmpi
+  
+  This will ensure that you build a complete stack for building the corresponding JEDI bundles.
```
climbfuji: See above, this repository is not about JEDI in my opinion. Also, the instructions to start with an empty template (or choose another template) and then add specs with `spack add` do the same, but in a `spack` way.

climbfuji (changes requested):

doc/source/Platforms.rst
climbfuji: :-) same thought

doc/source/Platforms.rst
climbfuji: Same question!

doc/source/Platforms.rst
climbfuji: This refers to line 359 ...

doc/source/Platforms.rst
climbfuji: Do we need this package? I don't know what it is doing and I didn't need it to build any of our codes.

If we need it, it needs to be added on all platforms and in the generic Linux instructions, too. If it is not mandatory, then maybe add a separate `code_block`, making it optional and providing a reason why this is good to have (and do the same for Linux)?

doc/source/Platforms.rst
```diff
@@ -456,6 +485,15 @@ Remember to activate the ``lua`` module environment and have MacTeX in your sear
    spack concretize
    spack install [--verbose] [--fail-fast]
 
+.. note::
```
climbfuji: I think this could be a general statement that could go in the quickstart section istead of the macOS specific instructions?

climbfuji: Definitely good to have and to recommend to users and developers!

doc/source/Platforms.rst
srherbener: This is the git flow package that we have been using for several years. This add macros to git that help manage the "git flow" methodology: https://nvie.com/posts/a-successful-git-branching-model/

doc/source/Platforms.rst
srherbener: Yes, I agree. Thanks for the suggestion (both of you).

srherbener:
Sorry, I just noticed that I have to go now. Thank you both for the reviews. I'll finish up my changes tomorrow and push them to github.

doc/source/Platforms.rst
srherbener: Made this change.

doc/source/Platforms.rst
srherbener: Made this change.

doc/source/Platforms.rst
srherbener: Thanks for the feedback. I was not thinking of the broader scope of spack-stack, and I think it makes sense to keep JEDI specific information out of the spack-stack documentation. I've removed this note.

doc/source/Platforms.rst
srherbener: I removed this note.

doc/source/Platforms.rst
srherbener: I moved this to the Quickstart section.

doc/source/Platforms.rst
srherbener: git-flow-avh is handy for managing feature branches and PRs for developers, and as you say is not necessary for building the code. We recommend it in the JEDI ReadTheDocs documentation and use the "git flow" commands in many tutorial examples including advocating it in our JEDI Academies.

I can see the point that adding the git-flow-avh package to the list of homebrew installs is introducing a JEDI specific need in the broader scope documentation. Plus you don't need it for doing the build.

However, the use of "git flow" perhaps goes beyond JEDI. The very next thing many people will want to do after running the build (if git-flow-avh is not mentioned here) will be to install git-flow-avh and start using it.

I'm leaning toward removing git-flow-avh from this list, and leaving it up to the different groups using spack-stack to advocate (or not advocate) for it.

What do you think?

climbfuji: I concur. We can always keep the recommendation in the JEDI docs. But I am also interested in learning more about it.

climbfuji (commented):
Let's remove the git-flow package from this PR and then it is good to go.

srherbener:
> Let's remove the git-flow package from this PR and then it is good to go.

Done!

kgerheiser (approved):

climbfuji (approved):

PR #208, opened 2022-06-14
release/v1 bug fixes: revert ESMF to 8.3.0b09, disable netCDF in HDF, TEST: run full CI tests for every PR

climbfuji:
Fixes https://github.com/NOAA-EMC/spack-stack/issues/205

Avoids issue https://github.com/NOAA-EMC/spack/issues/98 and https://github.com/NOAA-EMC/spack/issues/100 (to be confirmed) by reverting ESMF back to 8.3.0b09.

Note. The UFS is currently not moving to 8.3.0 due to a potential issue with the release, they are staying with 8.3.0b09 for now. See https://github.com/NOAA-EMC/fv3atm/issues/129

Also: This tests turns on the full CI build tests for the release/v1 branch. Currently a bit slow, but better to wait a little longer than to have to go back and fix broken builds.

### DIFF
--- a/configs/common/modules.yaml
+++ b/configs/common/modules.yaml
@@ -115,6 +115,8 @@ modules:
         suffixes:
           ^esmf@8.2.0~debug: 'esmf-8.2.0'
           ^esmf@8.2.0+debug: 'esmf-8.2.0-debug'
+          ^esmf@8.3.0b09~debug: 'esmf-8.3.0b09'
+          ^esmf@8.3.0b09+debug: 'esmf-8.3.0b09-debug'
           ^esmf@8.3.0~debug: 'esmf-8.3.0'
           ^esmf@8.3.0+debug: 'esmf-8.3.0-debug'
       openmpi:
@@ -334,6 +336,8 @@ modules:
         suffixes:
           ^esmf@8.2.0~debug: 'esmf-8.2.0'
           ^esmf@8.2.0+debug: 'esmf-8.2.0-debug'
+          ^esmf@8.3.0b09~debug: 'esmf-8.3.0b09'
+          ^esmf@8.3.0b09+debug: 'esmf-8.3.0b09-debug'
           ^esmf@8.3.0~debug: 'esmf-8.3.0'
           ^esmf@8.3.0+debug: 'esmf-8.3.0-debug'
       openmpi:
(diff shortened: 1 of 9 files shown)


climbfuji:
@kgerheiser I think we could speed up those CI tests by just building jedi-ufs-all and using spack add/remove for the few exceptions that we know of (no jedi-tools for intel, no jedi-ewok for macos/gnu).

climbfuji:
@kgerheiser @srherbener Ready from my side. All 47 (!) tests incl. the doc generation passed for commit https://github.com/NOAA-EMC/spack-stack/pull/208/commits/50ba52ecec4500394b1948246e27191c6ac4c5fd

climbfuji:
Commit https://github.com/NOAA-EMC/spack-stack/pull/208/commits/fc177bcc485194f3971eebc4cbc92d50b9d12086 shows that one can skip the entire CI tests (except the doc build) by including `[skip ci]` in the commit message.

kgerheiser:
I would say make it for PR and not `push`. Overkill to do it on every push in my opinion.

climbfuji:
> I would say make it for PR and not `push`. Overkill to do it on every push in my opinion.

Ok, let me make this change real quick.

kgerheiser:
> @kgerheiser I think we could speed up those CI tests by just building jedi-ufs-all and using spack add/remove for the few exceptions that we know of (no jedi-tools for intel, no jedi-ewok for macos/gnu).

I agree with this. It's something I've wanted to get around to. The script already removes the offending packages. Many of the packages it builds are duplicate so building `jedi-ufs-all` would be an improvement.,

climbfuji:
Ok, done. It's still building everything it seems. Let's make the switch to the jedi-ufs-all builds in the next PR if that's ok.

configs/common/modules.yaml
```diff
@@ -334,6 +336,8 @@ modules:
         suffixes:
           ^esmf@8.2.0~debug: 'esmf-8.2.0'
           ^esmf@8.2.0+debug: 'esmf-8.2.0-debug'
+          ^esmf@8.3.0b09~debug: 'esmf-8.3.0b09'
+          ^esmf@8.3.0b09+debug: 'esmf-8.3.0b09-debug'
```
kgerheiser, line 340: I have an improvement for this that will allow us to generically use versions instead of listing them manually.

kgerheiser (approved):

climbfuji:
Ok, will merge now since all tests passed earlier.

PR #210, opened 2022-06-14
Use new shumlib branch with support for Intel compilers

### LINKED ISSUE #106: shumlib always builds with gcc or clang
When I added the shumlib package a while back, I only configured the default `gcc` build (from UK MetOffice) and the `apple-clang`/`clang` build (added by me) in the spack package (`configs/repos/jedi/packages/shumlib/package.py`). It would be good to check if we can build shumlib with `Intel` as well, and add a guard for other compilers that don't have a config.

climbfuji:
Waiting on https://github.com/NOAA-EMC/spack/pull/103.

Fixes https://github.com/NOAA-EMC/spack-stack/issues/106.

CI tests were ok last night (which includes a ubuntu intel build). Will test again and also run some jedi-um code manually on one of the HPCs with Intel. The shumlib-GNU builds have been tested extensively in the past.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,9 +1,11 @@
 [submodule "spack"]
   path = spack
-  #url = https://github.com/spack/spack
-  #branch = develop
+  ##url = https://github.com/spack/spack
+  ##branch = develop
+  #url = https://github.com/NOAA-EMC/spack
+  #branch = release/jcsda_emc_release_v1
   url = https://github.com/NOAA-EMC/spack
-  branch = release/jcsda_emc_release_v1
+  branch = feature/shumlib_intel
 [submodule "doc/CMakeModules"]
   path = doc/CMakeModules
   url = https://github.com/noaa-emc/cmakemodules
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,11 +1,9 @@
 [submodule "spack"]
   path = spack
-  ##url = https://github.com/spack/spack
-  ##branch = develop
-  #url = https://github.com/NOAA-EMC/spack
-  #branch = release/jcsda_emc_release_v1
+  #url = https://github.com/spack/spack
+  #branch = develop
   url = https://github.com/NOAA-EMC/spack
-  branch = feature/shumlib_intel
+  branch = release/jcsda_emc_release_v1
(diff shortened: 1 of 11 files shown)


.gitmodules
```diff
@@ -1,9 +1,11 @@
...
   path = spack
-  #url = https://github.com/spack/spack
-  #branch = develop
+  ##url = https://github.com/spack/spack
+  ##branch = develop
+  #url = https://github.com/NOAA-EMC/spack
+  #branch = release/jcsda_emc_release_v1
   url = https://github.com/NOAA-EMC/spack
-  branch = release/jcsda_emc_release_v1
+  branch = feature/shumlib_intel
```
climbfuji: Needs to be reverted after the spack PR is merged, as usual

kgerheiser (approved):

climbfuji:
Full CI tests passed last night - https://github.com/NOAA-EMC/spack-stack/pull/210/commits/ef6ad7ba48ca129c7a3372fcd9f67d4769abb9bf. Since then there has been only an informal update to include the previous PR (which I had already pulled in), i.e. no code changes. And one bugfix in the skylab template that isn't used by CI.

Currently building/testing on Hera with Intel. If that all works, ready to merge.

.gitmodules
```diff
@@ -1,9 +1,11 @@
 [submodule "spack"]
   path = spack
-  #url = https://github.com/spack/spack
-  #branch = develop
+  ##url = https://github.com/spack/spack
+  ##branch = develop
+  #url = https://github.com/NOAA-EMC/spack
+  #branch = release/jcsda_emc_release_v1
```
srherbener: Please remove commented out code, unless you still need these for experimentation. Thanks!

climbfuji: This will be reverted once the testing is complete, the spack PR is merged and the submodule pointer is updated. (This is how git submodules work.)

climbfuji: But we should probably remove the standing comment that we ultimately want to point to the authoritative spack repo at some point.

srherbener: Thanks for the explanation and sounds good to me.

srherbener (approved):

PR #214, opened 2022-06-17
Feature/updates for rc2

### LINKED ISSUE #123: Duplicate packages being build with different hashes because of different Autoconf provider
See https://github.com/spack/spack/issues/29930

I found this on the NOAA Cloud platforms. Pkgconf is defined as an external, but it's not being used in all instances, and `pkg-config` is being used, resulting in different hashes for NetCDF. My solution was to set `pkg-config` to `buildable: False`.

### LINKED ISSUE #162: create mirror CI test failing
https://github.com/NOAA-EMC/spack-stack/runs/6663667075?check_suite_focus=true

It's still looking for `create.py`

### LINKED ISSUE #183: create-spack-mirror CI tests failing 06/06/2022
The create-spack-mirror CI tests are now (06/06/2022) failing with
```
==> Warning: included configuration files should be updated manually [files=${SPACK_STACK_DIR}/configs/common/modules.yaml, ${SPACK_STACK_DIR}/configs/common/config.yaml, site, packages.yaml]
==> Error: Invalid url format from url: ${SPACK_STACK_DIR}/cache/source_cache
Error: Process completed with exit code 3.
```

### LINKED ISSUE #206: py-pillow build error on Orion with gnu-10.2.0 (`gcc: error: unrecognized command-line option '-R'`) - update py-setuptools to 59.4.0 or remove version 57.4.0 from common packages.yaml
I think this error comes from using an external and very old `m4`. Trying to remove the external `m4` from the orion config and build again.

**If** that works, then also try without the external `m4` with Intel. (The intel build didn't have any problem with it.)

### LINKED ISSUE #211: ecflow not declared as a Python package
I'm testing the newly created stack on an AMI. I loaded all the modules needed for building fv3-bundle and running ewok. Running `import ecflow` gives me the error: 
```
ModuleNotFoundError: No module named 'ecflow'
```
It looks like ecflow is not declared as a Python package in spack and so is not set in `PYTHONPATH` correctly.

climbfuji:
A number of updates for spack-stack-1.0.0 release candidate 2:
- Update CI tests to use `jedi-ufs-all` template
- Fix `create-spack-mirror` CI test
- Add `yaksa` modulefile to blacklist
- Do not pin version for `py-setuptools`
- Update Discover site config
- Update Orion GNU site config
- By default, use environment modules (`tcl/tk`) instead of `lmod/lua` on AWS Ubuntu and Redhat
- Extend `jedi-ufs-all` template
- Update `skylab-1.0.0` template
- Add modulefile templates for `qt5` and `ecflow`
- Update documentation, especially instructions for building `ecflow`

Fixes #162 (see https://github.com/climbfuji/spack-stack/runs/6943697862?check_suite_focus=true)

Fixes https://github.com/NOAA-EMC/spack-stack/issues/211

Fixes https://github.com/NOAA-EMC/spack-stack/issues/213

Fixes https://github.com/NOAA-EMC/spack-stack/issues/123

Waiting on https://github.com/NOAA-EMC/spack/pull/106

CI tests all passed except for macos-apple-clang build (for some weird reason, the same test passed before - rerunning now without parallel spack installs). Rerunning macos-apple-clang now manually, check https://github.com/climbfuji/spack-stack/runs/6977275667?check_suite_focus=true for the progress.

### DIFF
--- a/.github/actions/setup-spack-stack/action.yaml
+++ b/.github/actions/setup-spack-stack/action.yaml
@@ -87,10 +87,7 @@ runs:
         brew install qt@5
         brew install readline
         brew install wget
-        brew list
       fi
-
-      exit 1
 
       # Install Python poetry to avoid install errors in spack
       python3 -m pip install poetry
@@ -217,11 +214,6 @@ runs:
         export "PATH=$HOME/mpi/bin:${PATH}"
       fi
 
-      # Make homebrew qt@5 detectable on macOS
-      if [[ "$RUNNER_OS" == "macOS" ]]; then
-        export "PATH=/usr/local/opt/qt@5/bin:${PATH}"
-      fi
-
       # Spack external find is by default only looking for build-tools, Either
       # need to search for additional packages explicitly.
       spack external find
@@ -229,8 +221,19 @@ runs:
       spack external find openmpi
       spack external find perl
       spack external find python@3
-      spack external find qt@5
       spack external find wget
+
+      # Find qt@5 on macOS
+      if [[ "$RUNNER_OS" == "macOS" ]]; then
+        PATH="/usr/local/opt/qt5/bin:$PATH" \
+             spack external find qt@5
+      fi
+
+      # Find curl on macOS
+      if [[ "$RUNNER_OS" == "macOS" ]]; then
+        PATH="/usr/local/opt/curl/bin:$PATH" \
+             spack external find curl
+      fi
 
   - name: configure-options
     shell: bash
--- a/.github/actions/setup-spack-stack/action.yaml
+++ b/.github/actions/setup-spack-stack/action.yaml
@@ -303,7 +303,7 @@ runs:
       #  tail -n 10 spack_install.log
       #  sleep 60
       #done
-      spack install --fail-fast
+      spack install --verbose --fail-fast
 
   - name: create-meta-modules
     if: ${{ inputs.concretize_only == 'false' }}
--- a/.github/actions/setup-spack-stack/action.yaml
+++ b/.github/actions/setup-spack-stack/action.yaml
@@ -257,6 +257,8 @@ runs:
         spack config add "packages:ecflow:variants:~ssl"
       fi
 
+      cat ${SPACK_ENV}/spack.yaml
+
       # Remove specs that don't build under certain configurations.
       # Only need that for the concretize_only steps, because the full
       # builds only build one spec at a time (and the cases that are
--- a/.github/actions/setup-spack-stack/action.yaml
+++ b/.github/actions/setup-spack-stack/action.yaml
@@ -79,10 +79,10 @@ runs:
             echo "source /opt/intel/oneapi/setvars.sh" >> ~/.bash_profile
         fi
       elif [[ "$RUNNER_OS" == "macOS" ]]; then
-        brew install curl
-        brew install git
-        brew install git-lfs
-        brew install openssl
+        # These are already installed
+        #brew install curl
+        #brew install git
+        #brew install git-lfs
         brew install lmod
         brew install qt@5
         brew install readline
@@ -214,6 +214,11 @@ runs:
         export "PATH=$HOME/mpi/bin:${PATH}"
       fi
 
+      # Make homebrew qt@5 detectable on macOS
+      if [[ "$RUNNER_OS" == "macOS" ]]; then
+        export "PATH=/usr/local/opt/qt@5/bin:${PATH}"
+      fi
+
       # Spack external find is by default only looking for build-tools, Either
       # need to search for additional packages explicitly.
       spack external find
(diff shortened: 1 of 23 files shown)


climbfuji:
@kgerheiser CI tests all pass, finally (https://github.com/NOAA-EMC/spack-stack/pull/214/commits/2934b243870563d0ff31509581487ed640a2eda2) - ready for merging from my side (first spack, then spack-stack).

Update. Looks like my last cleanup of the macOS CI build re-introduced the curl build error. I want to ignore this for the PR, as long as all the others pass, and fix it in a next PR, which has a few minor documentation updates that are not included here.

configs/sites/discover/packages.yaml
```diff
@@ -43,6 +37,44 @@ packages:
       prefix: /discover/swdev/jcsda/spack-stack/miniconda-3.9.7/
       modules:
       - miniconda/3.9.7
+###   glib:
```
climbfuji: Note. This junk will be removed later if not in a last minute doc push to this PR.

configs/sites/discover/packages.yaml
```diff
@@ -89,6 +121,14 @@ packages:
     externals:
     - spec: diffutils@3.3
       prefix: /usr
+  # NEEDS UPDATE!!!
```
climbfuji: This will be updated in a follow-up PR or in a last minute push to this PR.

configs/sites/discover/packages.yaml
```diff
@@ -177,10 +215,16 @@ packages:
     externals:
     - spec: pkg-config@0.28
       prefix: /usr
+  #qt:
```
climbfuji: This will be removed in a follow-up PR or in a last minute push to this PR.

doc/source/Prerequisites.rst
```diff
@@ -93,12 +106,61 @@ On HPC systems without a sufficient Qt5 installation, we install it outside of s
    gmake -j4 2>&1 | tee log.gmake
    gmake install 2>&1 | tee log.install
 
+Create modulefile ``/lustre/f2/pdata/esrl/gsd/spack-stack/modulefiles/qt/5.15.3`` from template ``doc/modulefile_templates/qt`` and update ``QT_PATH`` in this file.
```
climbfuji, line 109: This documentation will be updated and completed over the next week

climbfuji:
All other CI tests passed. This is ready to merge once the spack PR was merged, the submodule pointer updated, and the `.gitmodules` change reverted.

.github/actions/setup-spack-stack/action.yaml
```diff
@@ -280,27 +285,29 @@ runs:
...
-        nohup spack install --fail-fast >> spack_install.log 2>&1 &
-      done
-      #
-      while ps -ef | grep -ve 'grep' | grep 'spack install'; do
-        echo "Still running"
-        tail -n 10 spack_install.log
-        sleep 60
-      done
+      ## Note: This is unreliable ...
+      ## Speed up installation by building in parallel
```
kgerheiser, line 294: What makes it unreliable? I used this technique myself after seeing it here and it worked fine.

configs/common/packages.yaml
```diff
@@ -168,6 +168,9 @@
       version: [3.10.5]
     nlohmann-json-schema-validator:
       version: [2.1.0]
+    # Do not build pkg-config - https://github.com/NOAA-EMC/spack-stack/issues/123
+    pkg-config:
+      buildable: False
```
kgerheiser: Thanks!

.github/actions/setup-spack-stack/action.yaml
climbfuji: What happened a few times was that the install step was skipped, as if it ran the query if the jobs are still running faster than they appeared in the `ps` output. I thought this shouldn't be possible, but simply repeating the same run worked. Could probably be solved by adding a little sleep call in between?

configs/common/packages.yaml
kgerheiser: Actually, I would make this `pkgconf:buildable:false`. As far as I can tell, `pkg-config` is available on every platform as an external and we want to disable mixing the two.

climbfuji: Should we mark both as buildable: False?

configs/common/packages.yaml
```diff
@@ -208,8 +211,9 @@
       version: [1.55]
     py-openpyxl:
       version: [3.0.3]
-    py-setuptools:
-      version: [57.4.0]
+    # Try without
```
climbfuji, line 214: I think we can remove that comment as well.

configs/common/packages.yaml
climbfuji: Done.

kgerheiser (approved):

configs/common/packages.yaml
```diff
@@ -168,7 +168,9 @@
       version: [3.10.5]
     nlohmann-json-schema-validator:
       version: [2.1.0]
-    # Do not build pkg-config - https://github.com/NOAA-EMC/spack-stack/issues/123
+    # Do not build pkgconf/pkg-config - https://github.com/NOAA-EMC/spack-stack/issues/123
+    pkgconf:
+      buildable: False
     pkg-config:
       buildable: False
```
kgerheiser: One of them should be buildable. An external `pkg-config` is generally available on these HPC systems so that's why I say go for that and leave `pkgconf` as not buildable.

climbfuji:
If we think that pkg-config is generally available, then it won’t hurt to turn building off for both. I’ll try that for rc2, and if it doesn’t work across all platforms I’ll remove the `pkg-config` entry.

kgerheiser:
macOS does not have a `pkgconfig` provider built in

climbfuji:
> macOS does not have a `pkgconfig` provider built in

Ok, so let me change it right away then. Thanks!

PR #216, opened 2022-06-21
Enable two level namespace with openmpi on MacOS in spack

climbfuji:
Bug fix in spack to build openmpi correctly with two-level namespace (from @srherbener). This PR only handles the submodule pointer update.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,9 +1,11 @@
 [submodule "spack"]
   path = spack
-  #url = https://github.com/spack/spack
-  #branch = develop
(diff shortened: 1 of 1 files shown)


climbfuji:
I verified that the submodule pointer for spack is pointing to the head of jcsda_emc_spack_stack (https://github.com/NOAA-EMC/spack/commit/de7dfdba4a6cb204d670aee84684b4c81d7242ee). Ready to merge.

srherbener (approved):
It looks like the MAPL patch is missing, but that is consistent with the description and I believe that is the intent. (Just want to double check).

PR #227, opened 2022-06-24
Blacklist ecflow modules on Discover and Orion, documentation updates following release candidate 2

climbfuji:
- Blacklist `ecflow` modules on Discover and Orion, since this module is provided outside of spack (and the auto-generated module fails to set `PYTHONPATH` correctly).
- Build `mapl@2.12.3` on Discover after the build errors were fixed in a previous PR
- Put `solo`, `r2d2`, `ewok` back into `skylab-1.0.0`
- Documentation updates following release candidate 2

The change to the `ecflow` modules on Discover and Orion, as well as the `mapl@2.12.3` build on Discover, were tested with Intel and GNU. `solo`, `r2d2`, `ewok` were built on Discover, Orion, AWS (Red Hat, Ubuntu). All of these tests were performed in the rc2 installation on these systems.


configs/sites/discover/packages.yaml
```diff
@@ -37,11 +36,6 @@ packages:
       modules:
       - miniconda/3.9.7
 
-### Modifications of common packages
```
climbfuji, line 40: No longer needed, now that mapl@2.12.3 has been fixed.

configs/sites/orion/modules.yaml
```diff
@@ -2,3 +2,6 @@ modules:
   default:
     enable::
     - lmod
+    lmod:
+      blacklist:
+      - ecflow
```
climbfuji: Needs a newline

configs/sites/discover/modules.yaml
```diff
@@ -2,3 +2,6 @@ modules:
   default:
     enable::
     - lmod
+    lmod:
+      blacklist:
+      - ecflow
```
climbfuji: Needs a newline

configs/templates/skylab-1.0.0/spack.yaml
```diff
@@ -6,9 +6,9 @@ spack:
 
   packages:
     fiat:
-      version:: [1.0.0]
+      version: [1.0.0]
```
climbfuji, line 9: Double colons get messed up by the concretizer

doc/source/Quickstart.rst
```diff
@@ -71,6 +71,9 @@ The following instructions install a new spack environment on a pre-configured s
     spack concretize 2>&1 | tee log.concretize
     spack install [--verbose] [--fail-fast] 2>&1 | tee log.install
 
+.. note::
+  For platforms with multiple compilers in the site config, make sure that the correct compiler and corresponding MPI library are set ``envs/jedi-fv3.hera/site/packages.yaml`` before running ``spack concretize``. Also, check the output of ``spack concretize`` to make sure that the correct compiler is used (e.g. %intel-2022.0.1). If not, edit ``envs/jedi-fv3.hera/site/compilers.yaml`` and remove the offending compiler. Then, remove ``envs/jedi-fv3.hera/spack.lock`` and rerun ``spack concretize``.
```
srherbener: Looks like a typo in line 75. Should it be something like:
```suggestion
  For platforms with multiple compilers in the site config, make sure that the correct compiler and corresponding MPI library are set correctly in ``envs/jedi-fv3.hera/site/packages.yaml`` before running ``spack concretize``. Also, check the output of ``spack concretize`` to make sure that the correct compiler is used (e.g. %intel-2022.0.1). If not, edit ``envs/jedi-fv3.hera/site/compilers.yaml`` and remove the offending compiler. Then, remove ``envs/jedi-fv3.hera/spack.lock`` and rerun ``spack concretize``.
```

Also don't the new instructions for macOS and Linux, where you unset the SPACK_SYSTEM_CONFIG_PATH before running the `spack config add` commands, place the compiler spec in the `envs/my-env/spack.yaml` file instead of the `envs/my-env/site/package.yaml` file. Would it be worth adding a note on the Platform page about this difference? Sorry if that note is already there and I missed it.

srherbener (approved):

doc/source/Quickstart.rst
climbfuji: @srherbener that's actually a good point. I think it doesn't matter where those settings from `spack config add` go, but I agree it would be cleaner if they were in the site `package.yaml`. This way, one could probably use double colons and force spack to definitely use those settings. Let's try that in a follow-up PR!

PR #233, opened 2022-06-28
Add skylab-1.0.0-public template to avoid license issues, complete documentation on chaining spack environments

### LINKED ISSUE #230: Licensing restrictions require two versions of skylab template
We can't release GPL code to the public, and some of the software packages in the skylab-1.0.0 template are unfortunately GPL:
- fftw
- gsl
- libiconv - _ignore for now_ this may not be an issue
- parallelio - update: no longer needed, the authors agreed to change their license to Apache 2 (thanks!!!)

We'll need to create a second skylab template that gets used to install the software environment on AMIs and containers and which doesn't include any of these dependencies. This will require adding a few more variants to our convenience environment modules.

- jedi-base-env: need variant to not build fftw
    -  need to propagate up to not build ectrans, and create variant in atlas to turn off ectrans
- need to make sure that when we build openmpi, we turn off ibconv in libxml2 - _ignore for now_
- figure out what/who needs gsl
    - soca-env has it, it's used as standalone executable in soca-science
    - doesn't need to be in the public skylab template

climbfuji:
This PR:
- add `configs/templates/skylab-1.0.0-public/spack.yaml` - same as `configs/templates/skylab-1.0.0/spack.yaml` but with software that uses licenses that are incompatible with the JCSDA Apache 2 license turned off (basically only `fftw` --> `fiat`, `ectrans`, and build `ecmwf-atlas` without `ectrans` and without `fftw`)
- last documentation updates, including referencing `spack-stack-1.0.0` instead of a release candidate so that the code can be tagged after the upstream PR https://github.com/NOAA-EMC/spack/pull/113 is merged, the submodule pointer updated here, and `.gitmodules` reverted
- updates of the tag names for `solo`, `r2d2`, `ewok` - tested locally, because they are not built as part of the CI system.

Notes:
- waiting on https://github.com/NOAA-EMC/spack/pull/113
    - this spack PR adds the variant to turn off `fftw` for `jedi-base-env`
- fixes #230 
- full CI tests passed for https://github.com/NOAA-EMC/spack-stack/pull/233/commits/92e449a63bba52a01b1d8a962a6981687e2691fa
- manually tested on macOS - the installed software is sufficient to build everything that is required for this release (everything for `ufs-weather-model`, and everything for `jedi-ewok-env` and `jedi-fv3-env`)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2,10 +2,8 @@
   path = spack
   #url = https://github.com/spack/spack
   #branch = develop
-  #url = https://github.com/NOAA-EMC/spack
-  #branch = release/jcsda_emc_release_v1
-  url = https://github.com/climbfuji/spack
-  branch = feature/fftw_variant_for_jedi-base-env
+  url = https://github.com/NOAA-EMC/spack
+  branch = spack-stack-1.0.0
 [submodule "doc/CMakeModules"]
   path = doc/CMakeModules
   url = https://github.com/noaa-emc/cmakemodules
--- a/.gitmodules
+++ b/.gitmodules
@@ -2,8 +2,10 @@
   path = spack
   #url = https://github.com/spack/spack
   #branch = develop
-  url = https://github.com/NOAA-EMC/spack
-  branch = release/jcsda_emc_release_v1
+  #url = https://github.com/NOAA-EMC/spack
+  #branch = release/jcsda_emc_release_v1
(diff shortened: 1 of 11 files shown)


srherbener (approved):

climbfuji:
Everything is in place. I'll look at the documentation one more time after my lunchtime run before merging and tagging. Getting close!

PR #241, opened 2022-07-06
Update Spack submodule and new NCIO/MAPL versions

kgerheiser:
* Update MAPL to 2.22.0 which is used by ufs-weather-model
* Update NCIO to 1.1.2 which contains a bugfix
* Remove (LLVM) Clang tests for macOS. apple-clang is sufficent.
* Remove libjpeg-turbo from module blacklist and name it libjpeg.
* Update Spack submodule with latest jcsda_emc_spack_stack
* Update jedi-ufs-all with more specs


configs/common/packages.yaml
```diff
@@ -111,7 +111,7 @@
       version: [1.6.37]
     # Attention - when updating also check discover site config
     mapl:
-      version: [2.12.3]
+      version: [2.20.0]
```
climbfuji: Hopefully this builds on macOS ...

climbfuji (approved):

configs/common/packages.yaml
kgerheiser: Crap, this is the wrong version

climbfuji: Glad that I asked ;-) Let's see what the CI builds say.

kgerheiser:
I removed libjpeg-turbo from the module blacklist. I was building NCEPLIBS-g2 using Spack dependencies and libjpeg should be available to end user.

kgerheiser:
Something is wrong with the Clang build of MPI. C compiler cannot create executables.

climbfuji:
> Something is wrong with the Clang build of MPI. C compiler cannot create executables.

That's new. Maybe they updated the macOS image and it's a different version now? Should we use a fixed version instead of LATEST?

climbfuji:
> > Something is wrong with the Clang build of MPI. C compiler cannot create executables.
> 
> That's new. Maybe they updated the macOS image and it's a different version now? Should we use a fixed version instead of LATEST?

Also, we should probably be trying to do the correct two-level namespace change from Steve that we merged into spack rather than the half-baked solution that is in there now.

kgerheiser:
This is the problem: https://github.com/actions/virtual-environments/issues/5846

climbfuji:
> This is the problem: [actions/virtual-environments#5846](https://github.com/actions/virtual-environments/issues/5846)

Feel free to remove the (LLVM) clang tests (concretization, builds) entirely. It's enough in my opinion to have apple-clang and gcc.

kgerheiser:
I was thinking that too

kgerheiser:
Needed some fixes as noted here: https://github.com/NOAA-EMC/spack/pull/120

configs/common/modules.yaml
```diff
@@ -15,6 +15,7 @@ modules:
         ecmwf-atlas: '{compiler.name}/{compiler.version}/atlas/{version}'
         nlohmann-json: '{compiler.name}/{compiler.version}/json/{version}'
         nlohmann-json-schema-validator: '{compiler.name}/{compiler.version}/json-schema-validator/{version}'
+        libjpeg-turbo: '{compiler.name}/{compiler.version}/libjpeg/{version}'
```
climbfuji, line 18: This is only in the `tcl` section, I assume you want to do this for `lmod` as well?

climbfuji (changes requested):

climbfuji (approved):

Issue #243, opened 2022-07-08
Need to update Orion site config for GNU - use openmpi instead of mpich


climbfuji:
The acceptance testing for skylab-1.0 showed that gnu+mpich is several times slower than gnu+openmpi. Both of them were installed by the Orion sysadmins. The mpich installation also has the pmi and slurm dependency hardwired in, which means that any executable that gets linked against mpich mpi must run on a compute node using `srun`. The openmpi installation doesn't have the restriction, one can use both `srun` or `mpirun`.

Therefore we need to update Orion site config for GNU to use openmpi instead of mpich.

climbfuji:
Closed via #244

PR #244, opened 2022-07-08
Update Orion site config: use openmpi instead of GNU

climbfuji:
Update Orion site config: use `openmpi` instead of GNU - see #243 for a description of the problem.

Also: update submodule pointer for `spack` to include latest updates.


kgerheiser (approved):

climbfuji:
Probably good idea to run the full builds - I am also running the tests on my macOS using the self-hosted runner. See https://github.com/NOAA-EMC/spack-stack/runs/7252976910?check_suite_focus=true

PR #254, opened 2022-07-12
Ignore CI runs from docs, README, and issue template changes

kgerheiser:
The problem with this is that required checks cannot be skipped. They'll be left hanging because the run won't be triggered.

Github's proposed workaround is to create dummy jobs with the same name that pass

https://docs.github.com/en/enterprise-cloud@latest/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks

That is kind of hacky. There are also other tricks to detect which files have changed and skip steps based on that.

But for simplicity I think it would be best if we add these paths to the ignore list, and allow admin overrides of the status checks in the case where the workflow is not triggered (and only in that case), as @climbfuji mentioned here (https://github.com/NOAA-EMC/spack-stack/pull/252#issuecomment-1182259304).

### DIFF
--- a/.github/workflows/macos-apple-clang.yaml
+++ b/.github/workflows/macos-apple-clang.yaml
@@ -2,7 +2,11 @@ name: macos-apple-clang-build
 
 on:
   pull_request:
-  #push:
+    paths-ignore:
+      - 'doc/**'
+      - '**.md'
+      - '.github/ISSUE_TEMPLATE/*'
+      - '.gitignore'
   workflow_dispatch:
 
 # Use custom shell with -l so .bash_profile is sourced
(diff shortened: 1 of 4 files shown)


climbfuji (approved):
Fully agree!

kgerheiser:
I'm going to go ahead and disable the `Include administrators` option in the repository settings then.

edwardhartnett:
Are we disabling requirements for passing CI, just to prevent the CI from running on README.md and other documentation changes? I don't think that's a good idea...

climbfuji:
The discussion was around administrators being able to override CI using their good judgement. If something is changing README.md, then there is really no need to burn compute hours and wait for several hours for the builds to succeed. Unfortunately, Github doesn’t have a good way to skip required checks for such changes.

PR #262, opened 2022-07-19
Cancel duplicate workflows in same PR

### LINKED ISSUE #261: Prevent multiple workflows from running at once in same PR
When developing a PR it is common to make multiple pushes in a short time period as issues are resolved. But this leads to many long-running workflow jobs running at once. Github provides a setting so that when a push is made to a branch previous workflows are canceled.

Use the [`concurrency` setting described here](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency) to limit workflow runs.

kgerheiser:
When a push is made to a branch it will cancel in-progress workflows. I think this is usually the desired behavior, and prevents multiple long-running workflows from running at once which can stall workflows in other PRs. I often find myself manually canceling workflows as I add new commits to a branch.

With this when a push is made to a branch in-progress workflows will be canceled.

Fix #261

### DIFF
--- a/.github/workflows/macos-apple-clang.yaml
+++ b/.github/workflows/macos-apple-clang.yaml
@@ -9,6 +9,10 @@ on:
       - '.gitignore'
   workflow_dispatch:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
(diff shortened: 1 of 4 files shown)


climbfuji (approved):

kgerheiser:
Same issue exists in Jet config

PR #265, opened 2022-07-20
Fix Jet OpenMPI prefix

### LINKED ISSUE #263: External OpenMPI prefix on Hera and Jet not set
**Describe the bug**
No matter what I did Spack would rather choose `intel-oneapi-mpi` over OpenMPI, even after explicitly setting the MPI provider and compiler.

Then, I re-ran `spack external find openmpi` and updated my `packages.yaml` and now things seem to be working.

Edit: The `prefix` was missing from the `packages.yaml`.

Same issue on Jet.

kgerheiser:
Fix #263

### DIFF
--- a/configs/sites/jet/packages.yaml
+++ b/configs/sites/jet/packages.yaml
@@ -13,8 +13,10 @@ packages:
       prefix: /apps/oneapi
   openmpi:
     externals:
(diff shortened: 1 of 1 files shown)


climbfuji (approved):
Will merge without waiting for CI

Issue #266, opened 2022-07-20
Spack builds with Intel even when GCC is set as preferred


kgerheiser:
**Describe the bug**
@ulmononian reported not being able to build with GCC on Orion even if GCC/OpenMP is set as preferred in `packages.yaml`. I have also noticed this in the past. And then looking @climbfuji's Skylab 1.0.0 environment, I noticed he also commented out Intel MPI to force Spack to build with OpenMPI.

I created a bug report in the Spack repo here: https://github.com/spack/spack/issues/31669

climbfuji:
Yes, I've done this basically on every platform. When trying to build with GNU, I remove the Intel compiler so that spack doesn't take it. Not sure if it's needed everywhere, but that's the dumb fail safe approach.

Clearly a bug, setting `compilers::` in the site config should work.

climbfuji:
Issue https://github.com/spack/spack/issues/31669 is still unresolved

climbfuji:
This is basically fixed. We are moving forward with the unified environment, which uses compiler matrices, and we also have access to the new `requires` syntax that specifies hard requirements.

PR #272, opened 2022-07-22
Create hpc-dev environment

kgerheiser:
I think this should be our 'template' for deploying a single all-encompassing stack on HPC systems.

* It's self contained, with all non-implicit specs given in the spack.yaml

* You get multiple compilers at once without having to create separate environments. On all of our HPC systems we've been installing with Intel and GNU. A compiler can easily be added/removed.

* It excludes jedi-tools%intel

* It overrides the issue with #266 without having to manually edit anything

* The bundles get a version corresponding to the environment they are a part of.

I installed this on Hera successfully. If someone needs an extra package in this environment we simply add it to the list of packages until we retire this stack and move onto the next one.


climbfuji (commented):
That looks great. Not entirely clear to me ... do you get separate environments for each of the packages listed?

climbfuji:
> That looks great. Not entirely clear to me ... do you get separate environments for each of the packages listed?

Scrap that, you don't ... but I am surprised that the meta modules script worked out of the box?

kgerheiser:
Yes, the meta-modules work out of the box if you have them both in `packages:all:compilers/providers`.

climbfuji:
> Yes, the meta-modules work out of the box if you have them both in `packages:all:compilers/providers`.

Great! I wrote it that way but never tested it :-)

climbfuji (approved):

PR #274, opened 2022-07-25
Configure pull_request workflow runs using labels on Dom's self-hosted runner

climbfuji:
## Description
- Configure `pull_request` runs on Dom's macOS using self-hosted runners
- Use label `run-macos-dom` to trigger these runs
- Skip uploading artifacts, takes way too long (pushes the workflow past the 6h mark) and isn't needed
- Update submodule pointer for `CMakeModules` following the merge of https://github.com/NOAA-EMC/CMakeModules/pull/62 (does not affect spack-stack at all)

## Testing
- All CI tests (using Github runners and on Dom's Mac) passed for https://github.com/NOAA-EMC/spack-stack/pull/274/commits/e6a9ec92c2a8d5fac25c21724c0ba7b175de8a5d

### DIFF
--- a/.github/workflows/macos-dom.yaml
+++ b/.github/workflows/macos-dom.yaml
@@ -1,7 +1,16 @@
 name: macos-dom-build
 on:
-  #pull_request:
+  # uncomment if we decide to run tests when merging to develop branch
   #push:
+  #  # merges to develop branch
+  #  branches: [develop]
+  pull_request:
+    # pull request to develop with a label
+    types: [labeled]
+    branches: [develop]
+    defaults:
+      template: 'jedi-ufs-all'
+      specs: ''
   workflow_dispatch:
     inputs:
       template:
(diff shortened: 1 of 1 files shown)


.github/workflows/macos-dom.yaml
```diff
@@ -1,12 +1,18 @@
...
+  #  # merges to develop branch
+  #  branches: [develop]
+  pull_request:
+    # pull request to develop with a label
+    types: [labeled]
+    branches: [develop]
   workflow_dispatch:
     inputs:
       template:
         description: 'Base spack.yaml template. Default is an empty environment.'
```
climbfuji: I will update this description after the current round of tests are complete.

.github/workflows/macos-dom.yaml
```diff
@@ -20,6 +26,7 @@ defaults:
 
 jobs:
   macos-dom-build:
+    if: ${{ github.event.label.name == 'run-macos-dom' }}
```
climbfuji: Todo: check if this still works for manual `workflow_dispatch` runs

kgerheiser (approved):

PR #278, opened 2022-07-26
Add ecflow to Hera and Jet

### LINKED ISSUE #269: Add external ecflow to Hera/Jet
The system installed Ecflow on Jet/Hera should be sufficient (GUI + Python). Add this as an external.

kgerheiser:
[skip ci]

Fix #269


climbfuji (approved):
Nice, wish it was that easy on the other platforms!

PR #290, opened 2022-07-29
Add intel-based container build

mark-a-potts:
This adds ubuntu-intel as a site within the configs section and updates the versions of several libraries in the ufs-srw-public-v2 template (this could be an issue for official compatibility with the public release, but allows the stack to build).


kgerheiser:
Something I found helpful was to install a couple more packages with apt and add those as external so you don't have to build Perl, etc from scratch.

https://github.com/kgerheiser/spack-stack/blob/00bbef2cbbc019252c5fdf154b57798f40063f34/configs/containers/docker-ubuntu-gcc-openmpi.yaml#L26-L115

https://github.com/kgerheiser/spack-stack/blob/00bbef2cbbc019252c5fdf154b57798f40063f34/configs/containers/docker-ubuntu-gcc-openmpi.yaml#L148-L161

climbfuji:
Thanks, Mark. In the end, what we really want is the containers to be buildable by using `configs/containers/a-yaml-file`, for example:
```
configs/containers/docker-ubuntu-gcc-openmpi.yaml
```
And then follow the instructions in the spack-stack documentation https://spack-stack.readthedocs.io/en/latest/Quickstart.html#create-container:
```
See a list of preconfigured containers
spack stack create ctr -h

# Create container spack definition (spack.yaml) in directory envs/<container-config>
spack stack create ctr docker-ubuntu-gcc-openmpi --template=empty

# Descend into container environment directory
cd envs/docker-ubuntu-gcc-openmpi

# Edit config file and add the required specs in section "specs:"
emacs spack.yaml

# Docker: create Dockerfile and build container
# See section "container" in spack.yaml for additional information
spack containerize > Dockerfile
docker build -t myimage .
docker run -it myimage
```

Do you think it is possible to turn your site config into something like that? I can also see the benefit of having instructions for building generic linux/macos systems with Intel (outside containers), but for gcc, clang, ... what we have done so far is to use the linux.default and macos.default site configs and provide the instructions here: https://spack-stack.readthedocs.io/en/latest/Platforms.html#generating-new-site-configs

mark-a-potts:
I'll see what I can do in terms of converting it to a yaml-based approach next week.

mark-a-potts:
I have pushed a new docker-ubuntu-intel-impi implementation that works, but requires some manual manipulation of the Dockerfile. I have added a README.intel file that explains what must be changed after the "spack containerize" command is run. If anyone has an idea on how to get spack to install the intel-hpckit in a better manner, please let me know and I will try it out. Similarly, the spack.yaml file that gets created initially has both mpich and openmpi it it, which need to be removed. If there is a way to avoid that, please let me know that as well. Finally, I had to modify the Dockerfile template in spack to comment out the section that attempts to strip binaries because it fails and aborts the build.

mark-a-potts:
If it is possible, I would like to keep the ubuntu-intel "site" in some way for use in building containers for the UFS releases. I think that method of building containers is considerably more flexible and also allows spack to build and install the oneapi stuff directly, which eliminates many of the extra things that get installed via apt.

climbfuji:
I think it would be good if we could chat about this in a brief meeting this week, way more efficient than exchanging github comments?

kgerheiser:
Do you have a link to the container you have?

climbfuji:
@mark-a-potts I am going to send you some updates or ask you to try the new feature we added in for container builds, i.e. the extra instructions in the `pre_build` and `pre_final` phases. That should address the problem of having to move instructions around in the auto-generated Dockerfile.

climbfuji:
This PR is replaced by https://github.com/NOAA-EMC/spack-stack/pull/338 - #338 includes all the original commits from @mark-a-potts but uses recently developed features in our spack fork to have extra instructions in the `pre_build` and `pre_final` phases, plus making the Intel container more similar to the existing ones.

PR #357, opened 2022-09-25
release/1.1.0: update .gitmodules, add skylab-2.0.0 template

### LINKED ISSUE #355: CI test failure ubuntu+gcc when building `nccmp` due to changes in the GitHub actions CI images
**Describe the bug**
See https://github.com/NOAA-EMC/spack-stack/pull/354#issue-1383967319.

This needs to be fixed in the release branches for the spack-stack-1.1.0 release and then merged back to develop.

Note. `nccmp` built fine with Intel on Ubuntu for the same PR (https://github.com/NOAA-EMC/spack-stack/actions/runs/3113820342/jobs/5048922853). Maybe it was a one-off.

**To Reproduce**
Run CI tests

**Expected behavior**
CI tests pass

**System:**
What system(s) are you running the code on?

**Additional context**

climbfuji:
Three independent changes for the release branch only (not intended to be merged back to develop):

1. Template skylab-2.0.0 is required for the release branch only. It's a copy of the current skylab-dev template, but in addition contains the two older `crtm` versions (`2.3.0` and `v2.3-jedi.4`) so that users can choose between four versions altogether. The default is `2.4.0` for UFS and `v2.4-jedi` for JEDI.
2. Update `.gitmodules` to point to the release branch of the spack submodule.
3. Update CI tests to test the `skylab-2.0.0` template instead of `skylab-dev`.

Two additional changes that should find it's way back into develop:
1. Strip version numbers from virtual environment packages, and remove unnecessary ones (because they are implicitly required in the other env packages).
2. Bug fix for ubuntu CI tests - build curl instead of using OS curl (this also matches the instructions in readthedocs better). Fixes https://github.com/NOAA-EMC/spack-stack/issues/355

### DIFF
--- a/configs/templates/skylab-2.0.0/spack.yaml
+++ b/configs/templates/skylab-2.0.0/spack.yaml
@@ -1,0 +1,93 @@
+spack:
+  concretizer:
+    unify: when_possible
+
+  view: false
+
+  packages:
+    fiat:
+      version: [1.0.0]
+    ectrans:
+      version: [1.0.0]
+
+  view: false
+  include: []
+
+  specs:
+    # Virtual environment packages
+    - base-env@1.0.0
+    - jedi-base-env@1.0.0
+    - jedi-ewok-env@1.0.0
+    - jedi-fv3-env@1.0.0
+    - jedi-mpas-env@1.0.0
+    - jedi-ufs-env@1.0.0
+    - jedi-um-env@1.0.0
+    - soca-env@1.0.0
+
+    # Individual packages
+    - bacio@2.4.1
+    - bison@3.8.2
+    - bufr@11.7.1
+    - crtm@2.3.0
+    - crtm@2.4.0
+    - crtm@v2.3-jedi.4
+    - crtm@v2.4_jedi
+    - ecbuild@3.6.5
+    - eccodes@2.27.0
+    - ecflow@5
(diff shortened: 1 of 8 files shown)


configs/templates/skylab-2.0.0/spack.yaml
```diff
@@ -0,0 +1,93 @@
...
+      version: [1.0.0]
+    ectrans:
+      version: [1.0.0]
+
+  view: false
+  include: []
+
+  specs:
+    # Virtual environment packages
+    - base-env@1.0.0
```
mer-a-o: `1.0.0` or `2.0.0` for virtual environment packages?

climbfuji: Good question. It doesn't really matter since those environment packages have no real meaning, and some of the env packages haven't changed at all, some have (i.e. new dependencies). We could probably just strip the version and it be fine, too.

climbfuji: Trying this now. Also removing the unneeded env packages base-env and jedi-base-env, because these are implicitly required in the other env packages.

mer-a-o (approved):

srherbener (approved):
Looks good. (I double checked the spack submodule hash and that looks correct) Thanks!

PR #374, opened 2022-10-06
Update ufs-srw-dev spack.yaml

ulmononian:
This PR contains a few updates to the ufs-srw-dev template's spack.yaml. Changes include:

- Add MET/METPLUS verification packages to the specs list, as these will be required for the SRW v2.1 release.
- Add a wrf-io lmod/tcl projection so that module name formatting matches that of srw_common (i.e.: wrf_io)
- Ensures all spec versions match that of srw_common (https://github.com/ufs-community/ufs-srweather-app/blob/develop/modulefiles/srw_common), e.g. bufr.

Note: a successful installation of this template was performed on NOAA's GCLOUD platform and can be loaded using the following commands:

module use /contrib/EPIC/spack-stack/envs/srw-develop-intel/install/modulefiles/Core
module load stack-intel/2021.3.0
module load stack-intel-oneapi-mpi/2021.3.0
module load cmake/3.22.1

module load jasper/2.0.25
module load zlib/1.2.11
module load libpng/1.6.37

module load hdf5/1.10.6
module load netcdf-c/4.7.4
module load netcdf-fortran/4.5.4
module load pio/2.5.3
module load esmf/8.3.0b09 
module load fms/2022.01

module load bufr/11.7.0
module load bacio/2.4.1
module load crtm/2.3.0
module load g2/3.4.5
module load g2tmpl/1.10.0
module load ip/3.3.3
module load sp/2.3.3
module load w3nco/2.4.1

module load gftl-shared/1.5.0
module load yafyaml/0.5.1
module load mapl/2.22.0-esmf-8.3.0b09-esmf-8.3.0

module load nemsio/2.5.4
module load sfcio/1.4.1
module load sigio/2.3.2
module load wgrib2/2.0.8

module load w3emc/2.9.2
module load wrf_io/1.2.0
module load ncio/1.1.2

module load met/10.1.0
module load metplus/4.1.0

PR #380, opened 2022-10-13
Update cloud platform packages.yaml files

ulmononian:
This PR contains updates to the packages.yaml files for the NOAA cloud platforms (aws, azure, and gcp). The changes are relatively minimal, with the most significant adjustment being the switch for some external library prefixes to EPIC's new miniconda3 installation path. Other changes include pointers to updated git/git-lfs installation, as well as modified curl external configurations. These updates were catalyzed by various failures and errors (specifically with building cmake 3.22.1) caused when using the libraries provided by the current miniconda/python pointed to in these YAMLs, as well as repeated failures resulting from the current git and curl externals.

Installations using the updated package YAMLs in this PR were tested on aws, azure, and gcp w/ the ufs-srw-dev template (including the met/metplus additions from #374).


configs/sites/noaa-azure/packages.yaml
```diff
@@ -45,7 +45,6 @@ packages:
   berkeley-db:
     externals:
     - spec: berkeley-db@5.3.21
-      prefix: /usr
```
climbfuji: How is this working? I thought that spack either needs a prefix or a module to load for external packages?

climbfuji (commented):
I am having a bit of a problem with the directory structure - why do we need to replace the institution-independent `spack-stack` directory with `EPIC`? `JCSDA also wants to use the NOAA cloud environment in the next year, and we would definitely prefer an institution-independent structure. Unless we agree to have separate/duplicate site configs, which doesn't seem ideal. Let's discuss this before we make changes or merge as-is.

configs/sites/noaa-azure/packages.yaml
ulmononian: @climbfuji whoops...this was inadvertent. i did this commit in a roundabout way due to github issues on azure. fixed now.

ulmononian:
@climbfuji regarding the institutional independence: that's a very valid question and point. we ended up making these changes internally because the current package YAML files for the cloud sites were leading to several points of failure for srw and wm stack installs, one of which was the reference to the miniconda & python in the current develop versions of the files. we were unsure who installed/maintained the miniconda at /contrib/spack-stack/apps/miniconda, so we went ahead with our install. 

i'm happy to revert this to /contrib/spack-stack/apps/miniconda if the maintainer can either (i) let us know where we went wrong in trying to use packages.yaml as is to install ufs-srw-dev/ufs-srw-public-v2/ufs-wm-dev, or (ii)  update/fix the conda-provided packages to work with the spack envs. we need.

climbfuji:
Thanks, @ulmononian. I am not clear on how these containers are set up and used across NOAA and/or EPIC. It's probably best to have a quick chat early next week - I'll see if I can find an empty slot in our schedules.

ulmononian:
@climbfuji i installed a new miniconda/python in /contrib/spack-stack and made the changes in the packages.yaml files for the cloud platforms. i also removed the paths to external git/curl libraries that were installed in the EPIC space. while the git that is currently pointed to is pretty old, spack has lately been able to clone properly w/ it or it builds a newer git. pointing to the curl in /usr/bin never seems to work, so right now, i just have spack build it. something that can be done is to install new git & curl in /contrib/spack-stack/apps/utils, but i think that can come after this PR if it is wanted.

climbfuji (approved):

ulmononian:
@climbfuji thanks for the review on this, dom. perhaps we can touch base sometime in the coming week or so about the git/curl issues.

ulmononian:
@climbfuji is there more to be done on this one before it can be merged? i was not sure about the ci tests...

climbfuji:
Sorry, I was waiting for approval from EMC, but let's merge this. I will update from develop quickly, if everything looks ok (no need to wait for CI for these files), then I will merge. Sorry for the delay.

climbfuji (approved):

ulmononian:
> Sorry, I was waiting for approval from EMC, but let's merge this. I will update from develop quickly, if everything looks ok (no need to wait for CI for these files), then I will merge. Sorry for the delay.

no worries at all, @climbfuji!! i just wanted to make sure i was not forgetting to address something or missing a step in the standard spack-stack PR review process. thanks so much for reviewing and getting this merged.

PR #390, opened 2022-10-24
Update spack and CMakeModules submodules from authoritative develop branches as of 2022/10/24

climbfuji:
## Description

- Update spack and CMakeModules submodules from authoritative develop branches as of 2022/10/24
- Fix CI tests - revert back to a previous version where we use external curl and openssl on Ubuntu
- Bump `py-f90nml` version in `configs/containers/README.md` and `configs/templates/skylab-dev/spack.yaml`
- Bump `zlib` version to `1.12.13` or add `deprecated: true` in several templates, since the previous default `1.2.12` is now marked as deprecated.

### Dependencies

Waiting on https://github.com/NOAA-EMC/spack/pull/188

### Testing

- All CI tests pass
- Installed `skylab-dev` on Dom's macOS with both the old and new version of `py-f90nml`, and with all available versions of `parallelio` (2.5.2 to 2.5.9) to make sure that the changes in the `parallelio` package are correct

### DIFF
--- a/configs/templates/ufs-srw-public-v2/spack.yaml
+++ b/configs/templates/ufs-srw-public-v2/spack.yaml
@@ -2,6 +2,7 @@ spack:
   config:
     install_tree:
       root: $env/install
+    deprecated: true
   modules:
     default:
       roots:
@@ -19,7 +20,7 @@ spack:
 
   specs:
   - jasper@2.0.25
-  - zlib@1.2.13
+  - zlib@1.2.11
   - libpng@1.6.37
   - hdf5@1.10.6
   - netcdf-c@4.7.4
--- a/configs/templates/ufs-srw-public-v2/spack.yaml
+++ b/configs/templates/ufs-srw-public-v2/spack.yaml
@@ -19,7 +19,7 @@ spack:
 
   specs:
   - jasper@2.0.25
-  - zlib@1.2.12
+  - zlib@1.2.13
   - libpng@1.6.37
   - hdf5@1.10.6
   - netcdf-c@4.7.4
--- a/.github/actions/setup-spack-stack/action.yaml
+++ b/.github/actions/setup-spack-stack/action.yaml
@@ -54,9 +54,9 @@ runs:
         # Needed for the following apt-get install calls to work
         sudo apt-get update
 
-        # Don't do this, breaks several packages
-        ## Install Curl headers. Executable exists by default in spack external find.
-        #sudo apt-get install libcurl4-openssl-dev
+        # Install Curl/ssl headers. Executables exist by default in spack external find.
+        sudo apt-get install libcurl4-openssl-dev
+        sudo apt-get install libssl-dev
 
         # Install git-lfs to avoid compilation errors of "go" with Intel
         sudo apt-get install git-lfs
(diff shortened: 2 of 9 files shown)


climbfuji:
All, can I please have a review of this PR and its associates spack PR? Thanks ...

mer-a-o (approved):
Changes look good, I didn't test this branch.

configs/templates/ufs-srw-public-v2/spack.yaml
```diff
@@ -19,7 +19,7 @@ spack:
 
   specs:
   - jasper@2.0.25
-  - zlib@1.2.12
+  - zlib@1.2.13
```
ulmononian: i believe that the package version numbers for ufs application release branches (e.g.: this spack.yaml for ufs-srw release public v2) should be static, as the module versions required by those applications are fixed (e.g.: https://github.com/ufs-community/ufs-srweather-app/blob/release/public-v2/modulefiles/srw_common.spack). 

that is just to say, we need zlib to be 1.2.11 (it should not have been updated to 1.2.12 previously, either) for ufs-srw-public-v2. ufs-srw-dev's spack.yaml should be continually updated, though.

ulmononian:
@climbfuji regarding your description item 5 / note 1 from PR #380: you tested installs of pio versions 2.5.2-2.5.9 in this PR? just asking so i know if additional installation tests for pio are needed. if pio is good, my only other issue is w/ changing zlib version in ufs-srw-public-v2.

ulmononian (changes requested):

climbfuji:
> @climbfuji regarding your description item 5 / note 1 from PR #380: you tested installs of pio versions 2.5.2-2.5.9 in this PR? just asking so i know if additional installation tests for pio are needed. if pio is good, my only other issue is w/ changing zlib version in ufs-srw-public-v2.

Yes, I tested installs for all available versions between and including 2.5.2 and 2.5.9.

configs/templates/ufs-srw-public-v2/spack.yaml
climbfuji: I can revert this, no problem. I may have to add `deprecated: true` to the spack config section above. Will do after the CI tests finished, because there is no need to abort those (`ufs-srw-v2` is not CI tests, `ufs-srw-dev` is).

ulmononian: @climbfuji sounds good and thank you! i didn't even catch the zlib issue w/ this spack.yaml earlier. we have a deprecated: true statement in the srw-dev spack.yaml for the same reason.

ulmononian:
> 



> > @climbfuji regarding your description item 5 / note 1 from PR #380: you tested installs of pio versions 2.5.2-2.5.9 in this PR? just asking so i know if additional installation tests for pio are needed. if pio is good, my only other issue is w/ changing zlib version in ufs-srw-public-v2.
> 
> Yes, I tested installs for all available versions between and including 2.5.2 and 2.5.9.

nice! then i will approve too.

ulmononian (approved):
w/ zlib package verison fix forthcoming (after ci tests finish), and pio question confirmation, PR seems good to go.

configs/templates/ufs-srw-public-v2/spack.yaml
climbfuji: I made this change in my latest commit (https://github.com/NOAA-EMC/spack-stack/pull/390/commits/dbf89629a15a661152761abee273d5f6935aa3e6), merging now.

PR #398, opened 2022-11-07
Remove deprecated variant 'ensurepip' from external Python declarations

climbfuji:
## Description

The latest update of the spack submodule from the authoritative repo removed the `ensurepip` variant for Python. We need to remove this variant from all external Python package declarations (`configs/sites/*/packages.yaml`).

See https://github.com/NOAA-EMC/spack-stack/issues/397 for an example of the error (https://github.com/NOAA-EMC/spack-stack/issues/397#issuecomment-1306269234)

### DIFF
--- a/configs/sites/aws-pcluster/packages.yaml
+++ b/configs/sites/aws-pcluster/packages.yaml
@@ -16,7 +16,7 @@ packages:
   python:
     buildable: false
     externals:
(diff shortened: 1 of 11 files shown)


AlexanderRichert-NOAA (approved):
Looks good

climbfuji:
No need to wait for CI tests, none of these files is used by them. Thanks for the review.

PR #399, opened 2022-11-07
Update zlib version and macos instructions for handling the new Mac M1 native arm64 architecture

srherbener:
This PR contains updates to enable the successful build of the stack on the new Mac M1 arm64 architecture. The zlib version needs to be 1.12.13 for arm64 so this has been updated in the `configs/common/packages.yaml` file. The other config files were left alone since I was thinking these are deliberately set to specific versions according to different platforms. The documentation has been updated to reflect new steps necessary for building spack-stack on the M1 arm64 architecture.

### DIFF
--- a/doc/source/Platforms.rst
+++ b/doc/source/Platforms.rst
@@ -477,7 +477,7 @@ Note that applications are expected to run faster when the native Arm architectu
 
 A lot of binaries (bash for example) come in a "universal form" meaning they can run as Arm or Intel.
 MacOS provides a utility called ``arch`` which is handy for monitoring which architecture you are running on.
-For example, entering ``arch`` without any arguments will return which archtecture is running in your terminal window.
+For example, entering ``arch`` without any arguments will return which architecture is running in your terminal window.
 
 Homebrew notes
 --------------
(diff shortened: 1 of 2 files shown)


doc/source/Platforms.rst
```diff
@@ -465,6 +465,38 @@ The instructions below also assume a clean Homebrew installation with a clean Py
...
+The first architecture is Arm which is denoted by ``arm64`` and ``aarch64``, and the second is Intel which is denoted by ``x86_64`` and ``i386``.
+The Arm architecture is the native architecture on the M1 chip and the Intel architecture is what has existed for a number of years before the M1 chip showed up.
+
+With the new M1 chip, you can toggle between these two architectures, which is accomplished with a new app on M1 Macs called Rosetta2 (which is an Intel architecture emulator).
+When you get a new M1 mac, you may need to download Rosetta2.
+Note that applications are expected to run faster when the native Arm architecture is utilized.
+
+A lot of binaries (bash for example) come in a "universal form" meaning they can run as Arm or Intel.
+MacOS provides a utility called ``arch`` which is handy for monitoring which architecture you are running on.
+For example, entering ``arch`` without any arguments will return which archtecture is running in your terminal window.
```
climbfuji:
```suggestion
For example, entering ``arch`` without any arguments will return which architecture is running in your terminal window.
```

climbfuji (changes requested):
This is a great write-up! Thanks very much for contributing this. I will note, however, that on the latest macOS Ventura, it is no longer possible to duplicate the Terminal application. One option is to install both iTerm or Terminal and use one for Rosetta2 x86_64, one for native arch64, another option is to configure two profiles in your preferred terminal app to toggle between them (suggested by @ytremolet).

climbfuji (approved):

mer-a-o (approved):
@srherbener Thank you for updating the document. It is a lot clearer now.

climbfuji:
I am rerunning the failed CI tests, they got canceled for no obvious reason. Shouldn't take too long.

PR #406, opened 2022-11-18
Add skylab-no-python-dev template

### LINKED ISSUE #295: netcdf-c and netcdf-fortran can be updated, with special measures to enable zstandard compression
netcdf-c-4.9.0 and netcdf-fortran-4.6.0 are out! ;-)

In order for zstandard compression to work, the zstandard library must be installed before netcdf-c is built.

Also the configure option --with-plugin-dir=/some/dir.

Then the environment variable HDF5_PLUGIN_PATH must be set to /some/dir, for netcdf-fortran and every application that wants to use zstandard.

That is, if some rando netCDF application calls nc_def_var_zstandard() (or its fortran wrappers), the must have the HDF5_PLUGIN_PATH environment variable set correctly. or when they run that code they will get a "filter not found" error.

So this may present a challenge for spack. @kgerheiser you are now much more familar with spack than I am - what do you think is the best way to do this?

### LINKED ISSUE #403: spack v0.19.0 was released, update our spack branch
spack v0.19.0 was released, with many new features, bug fixes and improvements:

https://github.com/spack/spack/releases/tag/v0.19.0

We should update our branch with those changes. This will require several modifications to our configuration files, and open the door for new developments/easier ways to achieve what we are currently doing.

We should do this prior to the spack-stack-1.2.0 release mid December.

climbfuji:
## Description

Add `skylab-no-python-dev` template and update submodule pointer for spack for the necessary changes in bundle envs.

The submodule pointer update for spack also includes updates of maintainers for all "bundle" environments (we need to find a better name for those) and some minor cleanup of those environments (no impact on builds etc). I also had to add an update from the authoritative spack repository into the associated PR https://github.com/NOAA-EMC/spack/pull/198 because of updates to the Github action runners that broke the existing spack code. The Github action runners update also required pinning Ubuntu to 20.04 for now, and I used this opportunity to run the `create-spack-mirror` test (without uploading the artifacts) to the CI tests.

**Note.** The template `skylab-no-python-dev` is provided as a convenience for people who prefer to use their own Python environments on top off non-Python libraries from spack-stack, and as a workaround for people who have problems compiling the Python packages in spack-stack on their system. We do not recommend using this template, and we do not officially support it. Use it at your own risk, and be aware that certain features may not work if dependencies used by Python packages (e.g. `netcdf-c` used by `py-netcdf`) are compiled with other options than what we need/what is in spack-stack. These inconsistencies in redundantly provided libraries are the main reason why spack builds those Python packages itself.

## Dependencies

- waiting on https://github.com/NOAA-EMC/spack/pull/198

## Testing

- CI tests all passed (in the last run, the macos-gcc `skylab-dev` test timed out after six hours while building one of the last packagees `mapl`, but this passed in an earlier run).
- Template `skylab-no-python-dev` was tested (build only, not use with JEDI) on macOS Monterey with `apple-clang@13.1.6` and `openmpi@4.1.4`

## Issues

- fixes https://github.com/NOAA-EMC/spack-stack/issues/403
- fixes https://github.com/NOAA-EMC/spack-stack/issues/295
- progress on https://github.com/NOAA-EMC/spack-stack/issues/209

### DIFF
--- a/.github/actions/setup-spack-stack/action.yaml
+++ b/.github/actions/setup-spack-stack/action.yaml
@@ -91,9 +91,9 @@ runs:
         brew install readline
         brew install wget
 
+        ls -l /usr/local/Cellar
+
         # Remove macOS native Python 3.11 from /usr/local/bin
-        ls -l /usr/local/Cellar/
-        exit 1
         cd /usr/local/bin
         rm 2to3-3.11
         rm idle3.11
--- a/.github/actions/setup-spack-stack/action.yaml
+++ b/.github/actions/setup-spack-stack/action.yaml
@@ -92,6 +92,8 @@ runs:
         brew install wget
 
         # Remove macOS native Python 3.11 from /usr/local/bin
+        ls -l /usr/local/Cellar/
+        exit 1
         cd /usr/local/bin
         rm 2to3-3.11
         rm idle3.11
--- a/.github/actions/setup-spack-stack/action.yaml
+++ b/.github/actions/setup-spack-stack/action.yaml
@@ -90,7 +90,6 @@ runs:
         brew install qt@5
         brew install readline
         brew install wget
-
         ls -l /usr/local/Cellar
 
         # Remove macOS native Python 3.11 from /usr/local/bin
--- a/.github/workflows/concretize.yaml
+++ b/.github/workflows/concretize.yaml
@@ -2,7 +2,7 @@
 name: concretize-test
 
 on:
-  pull_request:
+  #pull_request:
   push:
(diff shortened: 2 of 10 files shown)


AlexanderRichert-NOAA:
Why do these changes require ubuntu 20.04? Is the expectation that we'll be able to revert to "latest" at some point?

climbfuji:
> Why do these changes require ubuntu 20.04? Is the expectation that we'll be able to revert to "latest" at some point?

It's got to do with Python 3.11 I believe, which breaks a lot of stuff.

PR #410, opened 2022-11-29
Speedup CI tests

climbfuji:
## Description
This PR rearranges the CI tests to make more use of caching (MPI builds, other external builds), to use spack mirrors to avoid repeated downloads of packages such as CRTM, and to reduce duplication. It makes heavy use of job dependencies and caching (which are per-PR, i.e. not between different PRs).

When we merge this, we need to update the branch protection rules to look for the new CI tests as requirements to be met.

## Testing
Only CI testing is needed. All tests pass for https://github.com/NOAA-EMC/spack-stack/pull/410/commits/1a92f1d4a021186af155f193a4d9925b02e9567f.


AlexanderRichert-NOAA:
Cool stuff, I'll need to look at this more thoroughly tomorrow. What's the overall time benefit here?

climbfuji:
> Cool stuff, I'll need to look at this more thoroughly tomorrow. What's the overall time benefit here?

If you have just one run of the CI tests for the PR, then the total amount of time is about the same. That's because the externals and mirror need to be created first, before the builds can kick off. The time for the build jobs themselves is shorter by 30-45 mins, which means it is less likely to run into the 6hr time limit for the skylab-dev builds (we really, really need to move from Github runners to self-hosted runners; and make use of build caches). You also need to download source codes only twice (macOS, Ubuntu) instead of four times.

The real benefit comes in when running the tests multiple times per PR (because something fails or updates are made etc.). In this case there are no additional downloads, no rebuilds of externals, just the actual spack-builds.

This helps a lot with CRTM, where the EMC tags download a huge amount of git-lfs data, which regularly exceeds the rate limit and leads to download errors.

.github/actions/setup-externals/action.yaml
```diff
@@ -0,0 +1,134 @@
...
+      path: /opt/intel
+      key: opt-intel-${{ inputs.mpi }}-${{ inputs.compiler }}.${{ runner.os }}3
+
+  - name: install-opt-intel
+    shell: bash
+    env:
+      OMP_NUM_THREADS: 4
+    if: steps.cache-opt-intel.outputs.cache-hit != 'true' && runner.os == 'Linux'
+    run: |
+      if [[ "$RUNNER_OS" == "Linux" ]]; then
```
climbfuji, line 35: This line (and 47-49) aren't really necessary. I added the runner condition in line 33 afterwards and didn't take 35 and 47-49 out. Can be done without having to rerun the CI tests.

.github/actions/setup-spack-mirror/action.yaml
```diff
@@ -0,0 +1,38 @@
...
+    with:
+      path: ~/spack-mirror
+      key: spack-mirror.${{ runner.os }}3
+
+  - name: create-spack-mirror
+    shell: bash
+    if: steps.cache-spack-mirror.outputs.cache-hit != 'true'
+    run: |
+      source ./setup.sh
+      # DH* TODO - REPLACE jedi-ufs-all with a better environment
```
climbfuji, line 24: Not this time - a follow-up when we clean up our various env packages and templates.

srherbener (commented):
I'm not sure I've interpreted correctly, but it looks like in the stack build tests, gnu is always tested with mpich and never with openmpi. Should there be a case where a build using gnu with openmpi is tested? I don't think we would have to do that on all platforms, I was just thinking of getting the gnu/openmpi combination in there at least once.

climbfuji:
> I'm not sure I've interpreted correctly, but it looks like in the stack build tests, gnu is always tested with mpich and never with openmpi. Should there be a case where a build using gnu with openmpi is tested? I don't think we would have to do that on all platforms, I was just thinking of getting the gnu/openmpi combination in there at least once.

We could do that, yes, good idea. Maybe using a different linux distribution than Ubuntu (CentOS)? Should we add this in one of the next PRs?

srherbener:
> > I'm not sure I've interpreted correctly, but it looks like in the stack build tests, gnu is always tested with mpich and never with openmpi. Should there be a case where a build using gnu with openmpi is tested? I don't think we would have to do that on all platforms, I was just thinking of getting the gnu/openmpi combination in there at least once.
> 
> We could do that, yes, good idea. Maybe using a different linux distribution than Ubuntu (CentOS)? Should we add this in one of the next PRs?

I'm good doing this in a subsequent PR.

Perhaps do RedHat (isn't that the OS EMC is moving to?).

climbfuji:
> > > I'm not sure I've interpreted correctly, but it looks like in the stack build tests, gnu is always tested with mpich and never with openmpi. Should there be a case where a build using gnu with openmpi is tested? I don't think we would have to do that on all platforms, I was just thinking of getting the gnu/openmpi combination in there at least once.
> > 
> > 
> > We could do that, yes, good idea. Maybe using a different linux distribution than Ubuntu (CentOS)? Should we add this in one of the next PRs?
> 
> I'm good doing this in a subsequent PR.
> 
> Perhaps do RedHat (isn't that the OS EMC is moving to?).

Unfortunately, neither CentOS nor RedHat (which are not the same but really very similar) are available as Github runners: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources 

We'll have to set those up as self-hosted runners, for example on AWS. I would prefer to create an issue to remind us of that idea, and address it when we look at improving the CI system in general (for example as part of a move of these repositories).

srherbener (approved):
I think this looks good, but I don't have a lot of experience with the CI, github workflow side of things. It would be good to get approval from other reviewers with more knowledge of CI.

.github/actions/setup-spack-mirror/action.yaml
ulmononian: wasn't sure about this "jedi-ufs-all" but based on your comment here, it seems like a change will come to this in a future PR post-release.

ulmononian (approved):
this looks nice to me. i would also concur with @srherbener about a gnu/openmpi combination, but otherwise five stars from me.

ulmononian:
> real benefit comes in when running the tests multiple times per PR (because something fails or updates are made etc.). In this case there are no additional downloads, no rebuilds of extern

i was also wondering about integrating this mirror functionality into the broader spack-stack environment installation process? i know that when we were getting spun up on the epic side and doing testing on various platforms, i at least was having to download packages from source and build them again and again, which can be expensive and monotonous/time-consuming. perhaps im just ignorant to existing tricks to avoid doing this, but it would be nice to be able to reference a standard set of source code and compiled libraries (assuming they were compiled w/ same dependency tree; hashes can handle those differences i suppose) from cache. is this something obvious i am missing?

climbfuji:
> > real benefit comes in when running the tests multiple times per PR (because something fails or updates are made etc.). In this case there are no additional downloads, no rebuilds of extern
> 
> i was also wondering about integrating this mirror functionality into the broader spack-stack environment installation process? i know that when we were getting spun up on the epic side and doing testing on various platforms, i at least was having to download packages from source and build them again and again, which can be expensive and monotonous/time-consuming. perhaps im just ignorant to existing tricks to avoid doing this, but it would be nice to be able to reference a standard set of source code and compiled libraries (assuming they were compiled w/ same dependency tree; hashes can handle those differences i suppose) from cache. is this something obvious i am missing?

The functionality is basically there, we just haven't added it to the instructions, and we haven't thought about how to make that easier.

Look at the hera site config. The mirror is already configured in there, but the process of creating or updating the mirror is manual and - in my opinion - too labor-intensive. We can probably add to our existing spack extension tools that facilitate creating/updating mirrors for source code and binaries (binary caches). In fact, I started working on this just this week on my mac locally. I should have something ready next week or the week after next to iterate on.

PR #415, opened 2022-12-01
Remove ~ensurepip from python spec in Acorn config

### LINKED ISSUE #414: Acorn config has deprecated variant in Python spec
**Describe the bug**
Acorn config (packages.yaml) still contains "~ensurepip" in the Python spec. This variant no longer exists, so causes concretization to fail.

**To Reproduce**
Fails with any environment using Acorn site config and Python.

**System:** Acorn

AlexanderRichert-NOAA:
~ensurepip variant is still in acorn, which causes any python-having builds with site=acorn to fail.

Fixes #414.

### DIFF
--- a/configs/sites/acorn/packages.yaml
+++ b/configs/sites/acorn/packages.yaml
@@ -41,7 +41,7 @@
         prefix: /usr
     python:
       externals:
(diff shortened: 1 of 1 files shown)


Hang-Lei-NOAA (approved):
Good catch!

climbfuji (approved):
Going to merge this right away, followed by the CI speedup PR (w/o rerunning the tests).

PR #426, opened 2022-12-28
Bug fix for S4 GNU compiler config

climbfuji:
Remove additional, empty `modules:` option from S4 GNU compiler config. Stumbled over this while testing on S4. No CI tests required for this PR.

### DIFF
--- a/configs/sites/s4/compilers.yaml
+++ b/configs/sites/s4/compilers.yaml
@@ -30,6 +30,5 @@ compilers:
     target: x86_64
     modules:
     - gnu/9.3.0
-    modules: []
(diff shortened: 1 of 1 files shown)


shlyaeva (approved):

AlexanderRichert-NOAA:
Seems reasonable. What happens when the "modules:" is left in there?

climbfuji:
> Seems reasonable. What happens when the "modules:" is left in there?

`spack stack setup-meta-modules` doesn't load the `gnu` compiler module, and subsequently it doesn't have path, library path etc set correctly.

PR #427, opened 2022-12-28
Always build Python with spack in CI tests / update submodule pointer for spack (moving packages to builtin)

### LINKED ISSUE #429: CI scripts not working as intended: wrong compiler gets used
**Describe the bug**
The CI script currently don't work as intended, because spack may use different compilers than those specified. For example. the macOS gcc build uses apple-clang, because cpython doesn't build with gcc. Similar problems may exist for Ubuntu.

**To Reproduce**
Look at the CI runs for previous PRs.

**Expected behavior**
CI runs use exactly the compilers that are requested.

**System:**
macOS/Ubuntu in CI

**Additional context**
This issue will be resolved by https://github.com/NOAA-EMC/spack-stack/pull/427

climbfuji:
## Description

1. Fix problem with CI runs not using the correct compilers by using the new-ish `require` syntax (hard requirements). Because the correct compilers are now enforced,
    (a) We can't use `gcc` on macOS anymore (`cpython` doesn't build with `gcc` on macOS) - move to Linux
    (b) We specify explicitly the compiler dependency for the MPI library (externally supplied)
2. Always build Python with spack in the CI tests (in preparation for doing the same on all user platforms in the future)
3. Add the `mvapich2` MPI library to the CI tests.

The CI changes also clean up how version information is supplied (it is now all in the workflow yaml files, as opposed to buried inside the action files).

The PR also updates the submodule pointer for spack for the changes described in https://github.com/NOAA-EMC/spack/pull/212 (Move packages from jcsda-emc "repository" to builtin)

## Issues

Fixes https://github.com/NOAA-EMC/spack-stack/issues/429

## Testing

- Tested locally on @climbfuji's macOS
- CI tests currently running (need to change the "expected" tests in the repo/branch settings due to the changes described above)

### DIFF
--- a/.github/actions/setup-externals/action.yaml
+++ b/.github/actions/setup-externals/action.yaml
@@ -133,18 +133,15 @@ runs:
         tar -xzf mvapich2-${MPI_VERSION}-1.tar.gz
         cd mvapich2-${MPI_VERSION}-1
         if [[ "$RUNNER_OS" == "macOS" ]]; then
-          #./configure --prefix=$HOME/mpi \
-          #    --enable-fortran=yes --enable-cxx \
-          #    --enable-two-level-namespace \
-          #    --with-device=ch4:ofi \
-          #    LIBS="-Wl,-commons,use_dylibs"
-          echo "mvapich2 not configured/tested for macOS!"
-          exit 1
+          ./configure --prefix=$HOME/mpi \
+              --enable-fortran=yes --enable-cxx \
+              --enable-two-level-namespace \
+              --with-device=ch3:sock \
+              LIBS="-Wl,-commons,use_dylibs"
         else
           ./configure --prefix=$HOME/mpi \
-              --enable-fortran=yes --enable-cxx
-              #--enable-fortran=yes --enable-cxx \
-              #--with-device=ch4:ofi
+              --enable-fortran=yes --enable-cxx \
+              --with-device=ch3:sock
         fi
         make -j4
         make install
--- a/.github/actions/setup-externals/action.yaml
+++ b/.github/actions/setup-externals/action.yaml
@@ -93,11 +93,6 @@ runs:
 
       MPI_NAME=`echo "${{ inputs.mpi }}" | cut -d "@" -f 1`
       MPI_VERSION=`echo "${{ inputs.mpi }}" | cut -d "@" -f 2`
-      echo "MPI NAME IS ${MPI_NAME}, VERSION IS ${MPI_VERSION}"
-      exit 1
-
-      export MPICH_VERSION="4.0.2"
-      export OPENMPI_VERSION="4.1.4"
 
       if [[ "${MPI_NAME}" == "openmpi" ]]; then
         wget https://download.open-mpi.org/release/open-mpi/v4.1/openmpi-${MPI_VERSION}.tar.gz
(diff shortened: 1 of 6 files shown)


.github/actions/setup-spack-stack/action.yaml
```diff
@@ -166,8 +168,8 @@ runs:
       spack config add "packages:mpi:buildable:False"
       spack config add "packages:all:providers:mpi:[${{ inputs.mpi }}]"
       spack config add "packages:all:compiler:[${{ inputs.compiler }}]"
-      # This is not working as expected
-      #spack config add "packages:all:require:'%${{ inputs.compiler }}'"
+      ## Set compiler as hard requirement for all packages
```
climbfuji: Note to self, replace ## with #

.github/actions/setup-os/action.yaml
```diff
@@ -29,43 +29,40 @@ runs:
...
-        rm python3.11-intel64
-        ln -sf ../Cellar/python@3.10/3.10.8/bin/2to3-3.10 2to3 
-        ln -sf ../Cellar/python@3.10/3.10.8/bin/idle3.10 idle3
-        ln -sf ../Cellar/python@3.10/3.10.8/bin/pip3.10 pip3
-        ln -sf ../Cellar/python@3.10/3.10.8/bin/pydoc3.10 pydoc3
-        ln -sf ../Cellar/python@3.10/3.10.8/bin/python3.10 python3
-        ln -sf ../Cellar/python@3.10/3.10.8/bin/python3.10-config python3-config
-        #ln -sf ... python3-intel64 # doesn't exist
-        ls -l /usr/local/bin
+        ### # Remove macOS native Python 3.11 from /usr/local/bin
```
climbfuji: Note to self, remove

.github/actions/setup-os/action.yaml
```diff
@@ -29,43 +29,40 @@ runs:
...
+        ### ln -sf ../Cellar/python@3.10/3.10.8/bin/python3.10-config python3-config
+        ### #ln -sf ... python3-intel64 # doesn't exist
+        ### ls -l /usr/local/bin
 
         # Print version of xcode
         pkgutil --pkg-info=com.apple.pkg.CLTools_Executables
 
-        # Install Python poetry to avoid install errors in spack
-        python3 -m pip install poetry
+        ### # Install Python poetry to avoid install errors in spack
```
climbfuji: Note to self, remove

.github/actions/setup-spack-stack/action.yaml
```diff
@@ -109,46 +114,43 @@ runs:
...
-      if [[ "${{ inputs.mpi }}" == "intel-oneapi-mpi" ]]; then
-        impi_ver=$(find /opt/intel/oneapi/mpi -maxdepth 1 -mindepth 1 -type d | xargs basename)
-        echo "" >> ${SPACK_ENV}/spack.yaml
-        echo "  packages:" >> ${SPACK_ENV}/spack.yaml
-        echo "    intel-oneapi-mpi:" >> ${SPACK_ENV}/spack.yaml
-        echo "      buildable: False" >> ${SPACK_ENV}/spack.yaml
-        echo "      externals:" >> ${SPACK_ENV}/spack.yaml
-        echo "      - spec: intel-oneapi-mpi@${impi_ver}" >> ${SPACK_ENV}/spack.yaml
-        echo "        prefix: /opt/intel/oneapi" >> ${SPACK_ENV}/spack.yaml
+      # Need to find external packages with the correct compiler,
```
climbfuji: Note to self. Rewrite as:
```
# Need to find external packages and annotate with the correct compiler,
```

.github/actions/setup-spack-stack/action.yaml
```diff
@@ -247,21 +249,17 @@ runs:
...
-          # This needs more work: need to replace "${{ inputs.compiler }}" = "gcc@9"
-          # with "gcc/9", otherwise the module load command will fail.
-          # module load stack-${{ inputs.compiler }}
-          # module load stack-python
-          # # jedi-tools doesn't have any mpi dependencies
-          # if [[ ! "${{ inputs.specs }}" == "jedi-tools" ]]; then
-          #   module load stack-${{ inputs.mpi }}
-          # fi
-        fi
+        #if [[ "$RUNNER_OS" == "Linux" ]]; then
```
climbfuji, line 252: This is just commenting out the entire block 250-262, doesn't make a difference.

.github/actions/setup-externals/action.yaml
```diff
@@ -112,10 +110,10 @@ runs:
...
         make -j4
         make install
-      elif [[ "${{ inputs.mpi }}" == "mpich"* ]]; then
-        wget http://www.mpich.org/static/downloads/${MPICH_VERSION}/mpich-${MPICH_VERSION}.tar.gz
-        tar -xzf mpich-${MPICH_VERSION}.tar.gz
-        cd mpich-${MPICH_VERSION}
+      elif [[ "${MPI_NAME}" == "mpich" ]]; then
+        wget http://www.mpich.org/static/downloads/${MPI_VERSION}/mpich-${MPI_VERSION}.tar.gz
+        tar -xzf mpich-${MPI_VERSION}.tar.gz
+        cd mpich-${MPI_VERSION}
```
AlexanderRichert-NOAA, line 116: Not that it's a problem, but is there a reason not to install openmpi/mvapich2 through spack?

srherbener (commented):
I think the code changes look good. My preference is to remove the commented code in this PR, unless that commented code needs to remain for some purpose (such as debugging).

Thanks!

.github/actions/setup-externals/action.yaml
climbfuji: The driver for that was that building MPI takes about 45mins and we are already hitting the 6hr walltime. Building it outside of spack-stack allows us to run it as a separate step (setup externals) and cache it. Reduces the times for iterating over build failures for spack packages by a lot.

climbfuji:
> I think the code changes look good. My preference is to remove the commented code in this PR, unless that commented code needs to remain for some purpose (such as debugging).
> 
> Thanks!

Thanks! We should keep the commented lines 250-262 in setup-spack-stack, because we need to get this to work as soon as possible. Otherwise I agree that comments should be removed.

srherbener (approved):
Thanks for the explanation @climbfuji. I think I understand - getting this PR in now will enable subsequent PRs to have faster running CI testing so it's important to merge now and clean up the commented code later.

AlexanderRichert-NOAA (approved):

PR #433, opened 2023-01-04
Add python to default modules for linux and macos

### LINKED ISSUE #432: Confusion with python version for mac build
**Describe the bug**

I tried building using the latest (as of yesterday) spack-stack, spack repos on my MacBook. The stack got built successfully and a subsequent build of jedi-bundle worked, but the ioda python API was broken because of finding a python in `/usr/local/bin` instead of in the stack.

I discovered that python 3.10.8 was built as part of the stack,  but there was not a lua module built for python/3.10.8 due to python being listed in the tcl/lua module blacklists in the common/module.yaml file.

**To Reproduce**

Build spack-stack using the environment created by: `spack stack create env --site macos.default --template skylab-dev --name skylab.mymacos`

Then clone and build jedi-bundle, and run `ctest`.

**Expected behavior**

All ioda tests pass.

**System:**

MacOS, Ventura 13.1, command line tools version 14.2.

**Additional context**

Is the intention now for python to always be part of the stack. If so, then should we remove python from the module blacklists?

srherbener:
This PR places the python package on the module whitelist for generic linux and macos sites. It leaves python on the module blacklist for the HPC sites since the HPC platforms are currently providing an external python module. This change is being done to support the methodology of always building python in the stack for non HPC linux and macos platforms.

Fixes #432

### DIFF
--- a/configs/sites/linux.default/modules.yaml
+++ b/configs/sites/linux.default/modules.yaml
@@ -7,3 +7,4 @@ modules:
       # List of packages for which we need modules that are blacklisted by default
       - openmpi
       - mpich
+      - python
--- a/configs/sites/macos.default/modules.yaml
+++ b/configs/sites/macos.default/modules.yaml
@@ -7,3 +7,4 @@ modules:
       # List of packages for which we need modules that are blacklisted by default
       - openmpi
       - mpich
+      - python


AlexanderRichert-NOAA (approved):
Looks good.

climbfuji (approved):
Thanks @srherbener !

PR #437, opened 2023-01-11
Acorn config update

AlexanderRichert-NOAA:
This PR
- updates acorn config to use site-specific ESMF settings (esmf_of & esmf_comm);
- fixes the external python spec (no more ucs4 variant); and
- updates spack submodule pointer.

The first item raises a possibly interesting question. The "variants:" setting in packages.yaml is not cumulative when using multiple packages.yaml's. I tested for example keeping the "~xerces ~pnetcdf +pio" in common/packages.yaml and just adding the new esmf settings to acorn/packages.yaml, but then those settings under common/ get lost. If there were a way to specify variants cumulatively across files and avoid redundancy, that would be great.

### DIFF
--- a/.github/actions/setup-spack-stack/action.yaml
+++ b/.github/actions/setup-spack-stack/action.yaml
@@ -205,6 +205,7 @@ runs:
       cd ${{ inputs.path }}
       source setup.sh
       spack env activate ${{ inputs.path }}/envs/${{ inputs.name }}
+      sed -i '/curl:/,+3d' ${SPACK_ENV}/spack.yaml
       cat ${SPACK_ENV}/spack.yaml
       spack concretize | tee ${GITHUB_WORKSPACE}/logs/concretize.txt
 
--- a/.github/actions/setup-spack-stack/action.yaml
+++ b/.github/actions/setup-spack-stack/action.yaml
@@ -54,7 +54,7 @@ runs:
 
   - name: check-cache-opt-intel
     shell: bash
-    if: steps.cache-opt-intel.outputs.cache-hit != 'true' && runner.os == 'Linux'
+    if: steps.cache-opt-intel.outputs.cache-hit != 'true' && runner.os == 'Linux' && inputs.compiler == 'Intel*'
     run: |
       echo "Cached /opt/intel not found!"
       exit 1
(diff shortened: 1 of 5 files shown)


Hang-Lei-NOAA (approved):
I have implemented these changes in the spack-stack installation on Acorn . These are proved to be necessary changes for esmf/830.

climbfuji:
> If there were a way to specify variants cumulatively across files and avoid redundancy, that would be great.

I am not aware of any, but that would be great to have, yes

climbfuji (approved):

AlexanderRichert-NOAA:
@climbfuji Any idea what's going on with https://github.com/NOAA-EMC/spack-stack/actions/runs/3895250997/jobs/6650856481? In the nccmp build, libnetcdf.so has undefined refs to a handful of curl functions, even though curl is pretty clearly installed.

climbfuji:
I just reran the build. I’ve seen weird errors in the past when exceeding the resources of the free github runners. But I doubt that’s the problem. More likely they (github actions) updated their images in the last few days. That happens frequently and often breaks our builds, because spack is having problems with curl libraries being preinstalled or not on Ubuntu (we’ve been going back and forth between installing them with apt and using them, or building them with spack instead).

climbfuji:
Two of 20 CI tests timed out, I am rerunning them.

.github/actions/setup-spack-stack/action.yaml
```diff
@@ -54,7 +54,7 @@ runs:
 
   - name: check-cache-opt-intel
     shell: bash
-    if: steps.cache-opt-intel.outputs.cache-hit != 'true' && runner.os == 'Linux'
+    if: steps.cache-opt-intel.outputs.cache-hit != 'true' && runner.os == 'Linux' && inputs.compiler == 'Intel*'
```
climbfuji: This should not be necessary, it worked w/o this until now

.github/actions/setup-spack-stack/action.yaml
```diff
@@ -148,7 +148,7 @@ runs:
       spack external find --path /usr/local/opt/qt@5 qt
 
       if [[ "$RUNNER_OS" == "Linux" ]]; then
-        spack external find curl
+        echo spack external find curl
```
climbfuji: If we really can't use the OS curl, then we should remove the logic entirely.

climbfuji:
@AlexanderRichert-NOAA I will try what we do in our instructions for Ubuntu (https://spack-stack.readthedocs.io/en/latest/Platforms.html#linux) for my current PR https://github.com/NOAA-EMC/spack-stack/pull/442.

climbfuji:
One of the CI tests timed out after 6hrs close to the finishing line, all others passed. Will merge now.

PR #438, opened 2023-01-11
Update python load command for AWS Red Hat 8

climbfuji:
See https://github.com/JCSDA-internal/jedi-docs/pull/542#pullrequestreview-1244246609 (if you can).
> One command is that iin the list of modules to load for AWS Red Hat 8, it is now module load stack-python/3.9.13 instead of module load stack-python/3.9.7

No testing needed, can be merged anytime.


AlexanderRichert-NOAA (approved):
I dunno this seems pretty complicated...

PR #439, opened 2023-01-11
Add gaea-c5 config

AlexanderRichert-NOAA:
This PR adds "gaea-c5" to site configs. I've only added intel compiler so far. For packages.yaml, I added cmake, python, git, and git-lfs based on modules, then used "spack external find --all" for everything else. I've set "buildable: false" for several packages to make builds faster but these aren't strictly needed. I've added the same esmf scheme as for acorn (esmf_os/esmf_comm).

### DIFF
--- a/configs/sites/gaea-c5/compilers.yaml
+++ b/configs/sites/gaea-c5/compilers.yaml
@@ -1,0 +1,22 @@
+compilers:
+- compiler:
+    spec: intel@2022.0.2
+    paths:
+      cc: cc
+      cxx: CC
+      f77: ftn
+      fc: ftn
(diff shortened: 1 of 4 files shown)


climbfuji:
Thanks @AlexanderRichert-NOAA! I'll check out your branch and add the missing bits and pieces for skylab in the next days.

climbfuji:
This PR is (with its original commit history) included in https://github.com/NOAA-EMC/spack-stack/pull/445 and should be flagged as "merged" automatically when #445 is merged.

climbfuji:
Replaced by #445.

PR #449, opened 2023-01-20
Run full CI tests only on PRs that are ready for review

climbfuji:
## Description
1. Run full CI tests only on PRs that are ready for review, not on draft PRs
    - but for now, still create the caches (mirror and external dependencies) for draft PRs, since they get reused
3. Update self-hosted CI runs on Dom's macOS
    - reflect changes from earlier PRs that build Python in spack

## Testing
- Tested successfully on Dom's macOS
- CI tests mostly successful, one test timed out close to the finish line and doesn't want to restart because of the missing cache issue that we don't understand

### DIFF
--- a/.github/workflows/macos-dom.yaml
+++ b/.github/workflows/macos-dom.yaml
@@ -44,7 +44,6 @@ jobs:
 
           spack external find
           spack external find perl
-          spack external find python
           spack external find wget
           PATH="/usr/local/opt/qt@5/bin:${PATH}" spack external find qt
           PATH="/usr/local/opt/curl/bin:${PATH}" spack external find curl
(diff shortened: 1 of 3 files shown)


.github/workflows/macos-dom.yaml
```diff
@@ -44,7 +44,6 @@ jobs:
 
           spack external find
           spack external find perl
-          spack external find python
```
ulmononian, line 47: assuming you intentionally removed this, but just wanted to ask in case.

ulmononian (approved):
good call to hold off on the full suite until prs converted from draft. looks good.

.github/workflows/macos-dom.yaml
climbfuji: Yes, we build Python with spack these days, therefore don't try to find an external package.

srherbener (approved):
This looks great. Thanks for adding this feature.

Issue #451, opened 2023-01-21
srw-app templates are missing gsi-ncdiag


climbfuji:
**Describe the bug**
I built the srw-app env on Cheyenne and noticed that in order to build the entire srw-app, the `gsi-ncdiag` package is missing in both templates:
https://github.com/NOAA-EMC/spack-stack/blob/develop/configs/templates/ufs-srw-public-v2/spack.yaml
https://github.com/NOAA-EMC/spack-stack/blob/develop/configs/templates/ufs-srw-dev/spack.yaml

**To Reproduce**
See above

**Expected behavior**
The templates contain everything needed to build the entire ufs-srw-app code

**System:**
Cheyenne/Intel

**Additional context**
n/a

ulmononian:
thanks for pointing this out. i'll update both of these templates w/ ncdiag@1.0.0. for consistency when i test the updated env. install: i assume you ran something like `./devbuild.sh --platform=cheyenne gsi` to do the srw install with gsi turned on and it failed somewhere along the line in the gsi build?

ulmononian:
actually wondering if we should update the ufs-srw-public-v2 template to v2.1.0 given the latest release branch of the srw is 2.1.0 (https://github.com/ufs-community/ufs-srweather-app/releases/tag/ufs-srw-v2.1.0)? not sure the best way to handle env. template naming/updating for minor releases. it seems like it would prove too cumbersome/clunky to have a template for every minor release...

ulmononian:
@climbfuji i added gsi-ncdiag@1.0.0 (w/ a projection to "ncdiag") to our srw-dev intel env. on cheyenne here:  /glade/work/epicufsrt/spack-stack/envs/srw-develop-intel. i'll run a test of the srw build w/ gsi in the morning.

climbfuji:
> thanks for pointing this out. i'll update both of these templates w/ ncdiag@1.0.0. for consistency when i test the updated env. install: i assume you ran something like `./devbuild.sh --platform=cheyenne gsi` to do the srw install with gsi turned on and it failed somewhere along the line in the gsi build?

I ran `./devbuild.sh --platform=cheyenne` without any arguments, i.e. I assume it was building everything that comes with the srw app.

ulmononian:
[...]

PR #452, opened 2023-01-22
Add gsi-ncdiag to srw template(s)

ulmononian:
@climbfuji pointed out in #451 that both of the srw environment templates are missing ncdiag, which is needed if gsi is to be built as part of the srw.

this is just a simple update to (at least) the ufs-srw-dev template.


ulmononian:
added gsi-ncdiag@1.0.0 (w/ a projection to "ncdiag") to epic maintained srw-dev intel env. on cheyenne here: /glade/work/epicufsrt/spack-stack/envs/srw-develop-intel. 

@climbfuji i utilized some changes from a branch i made a while back to address some cheyenne packages.yaml & compilers.yaml issues i had: https://github.com/ulmononian/spack-stack/tree/feature/cheyenne_update_compiler. i don't know if you want to merge any of minor those changes (key commits were to some external package paths [note: both cheyenne and orion packages.yaml still used institution-specific python paths] and commenting out the gnu compiler option in the compilers.yaml. it seems to want to default to that on cheyenne. other option is just to force the compiler at install time)

climbfuji (approved):
At some point we should stop remaining the modules for legacy reasons and simply use the package names from spack ...

ulmononian:
> At some point we should stop remaining the modules for legacy reasons and simply use the package names from spack ...

agreed. however, requiring the change of parallelio -> pio and wrf-io -> wrf_io everytime is an excellent way to test general negligence and/or fastidiousness of stack developers, though ;)

also: moved this from draft to regular PR now.

AlexanderRichert-NOAA (approved):

PR #456, opened 2023-01-27
Add 'beta' variant for ESMF to config files

AlexanderRichert-NOAA:
This PR goes with https://github.com/NOAA-EMC/spack/pull/218. That PR adds a dummy variant for ESMF that allows it to be designated as a beta version. This can then be used to distinguish beta vs. non-beta versions in modules.yaml. See https://github.com/NOAA-EMC/spack/pull/218 description for caveats and questions.

The main changes here are to config/common/{modules,packages}.yaml, plus a number of other config files where it seemed appropriate (but please look for any files I missed, or point out ones that don't need to be changed).


mathomp4:
@AlexanderRichert-NOAA Are the `snapshot=none` needed for the non-snapshot uses:
```
          ^esmf@8.2.0~debug snapshot=none: 'esmf-8.2.0'
```
From my reading, `snapshot=none` should be the default, right?

AlexanderRichert-NOAA:
> @AlexanderRichert-NOAA Are the `snapshot=none` needed for the non-snapshot uses:
> 
> ```
>           ^esmf@8.2.0~debug snapshot=none: 'esmf-8.2.0'
> ```
> 
> From my reading, `snapshot=none` should be the default, right?

Yes, this only needs to be specified if we ever use a beta snapshot version of 8.2.0, because if the spec to match were just `^esmf@8.2.0~debug`, then it would still get matched and applied in the event we're installing, say, `^esmf@8.2.0~debug snapshot=b01`.

srherbener (approved):
I didn't test, but I assume that this has been tested and works for your needs. Thanks for adding this.

configs/common/packages.yaml
```diff
@@ -91,6 +91,8 @@
       # Note: Uncommenting this entry will break
       # the container builds.
       #version: [2.11.0]
+    glib:
```
climbfuji: What do we need this for in the ESMF variant PR? If I had to guess I'd say this will break the builds on more than just one platform.

climbfuji (changes requested):

configs/common/packages.yaml
AlexanderRichert-NOAA: I don't remember adding this, I'll have to see where this came from...

PR #462, opened 2023-02-01
Update spack submodule to include latest ufs-weather-model (package) updates and fiat tags

climbfuji:
## Description

This PR updates the submodule pointer for spack for the changes in https://github.com/NOAA-EMC/spack/pull/217 (add `@develop` to ufs-weather-model) and https://github.com/NOAA-EMC/spack/pull/219 (`fiat` tags).

## Issues

Working towards https://github.com/NOAA-EMC/spack/issues/28

## Testing

This PR needs proper testing - not only to build, but also to run jedi-bundle (because the fiat tags that ECMWF created are slightly different hashes than what we used as "fake" 1.0.0 and 1.1.0 versions until now).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,9 +1,11 @@
 [submodule "spack"]
   path = spack
-  #url = https://github.com/spack/spack
-  #branch = develop
-  url = https://github.com/NOAA-EMC/spack
-  branch = jcsda_emc_spack_stack
+  ##url = https://github.com/spack/spack
+  ##branch = develop
+  #url = https://github.com/NOAA-EMC/spack
+  #branch = jcsda_emc_spack_stack
+  url = https://github.com/climbfuji/spack
+  branch = feature/fiat_tags
 [submodule "doc/CMakeModules"]
   path = doc/CMakeModules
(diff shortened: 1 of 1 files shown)


srherbener (commented):
Thanks. I agree that this needs a proper testing to be completed before merging.

srherbener:
I tested on my M1 Mac using the arm64 architecture. Spack-stack built successfully, then jedi-bundle built successfully, and the subsequent ctest returned with 21 test failures. This is 9 less than what we have been seeing so something changed that allowed 9 more tests to pass!

climbfuji:
> I tested on my M1 Mac using the arm64 architecture. Spack-stack built successfully, then jedi-bundle built successfully, and the subsequent ctest returned with 21 test failures. This is 9 less than what we have been seeing so something changed that allowed 9 more tests to pass!

This is great, thanks very much! Let me try this on one a Linux system (AWS parallel cluster) before we merge it.

srherbener (approved):

climbfuji:
@srherbener I tested this successfully on AWS parallelcluster with Intel, so I am going to merge the spack PR now and then update the submodule pointer in this PR.

PR #466, opened 2023-02-09
Add emc-rhel config

### LINKED ISSUE #464: Build spack-stack on standard NCEP/EMC Linux workstations
spack-stack builds on RHEL8.
The NCEP/EMC linux workstations are RHEL8.

Support to build and deploy on NCEP/EMC RHEL8 linux workstations to allow local development 
A recipe for these workstations similar to other supported machines.

**Additional context**
Attempts to build on NCEP/EMC RHEL8 linux workstations were unsuccessful.
It is likely a user error.
However, since these machines are standard across EMC, spack-stack should be able to build and deploy the library stack on these class of machines.

AlexanderRichert-NOAA:
This PR adds a site config for EMC's RedHat workstations, called emc-rhel. Tested on my workstation and @aerorahul's. I removed a couple packages (e.g., curl) from the results of `spack external find --all`. Uses `/opt/rh/gcc-toolset-11`. MPI needs to be installed through Spack (based on the two workstations I've tested). Rahul ran into system resource issues installing with the full number of CPUs on his machine, so for now I'm setting the limit to 8 processes.

Fixes https://github.com/NOAA-EMC/spack-stack/issues/464


configs/sites/emc-rhel/modules.yaml
```diff
@@ -0,0 +1,15 @@
+modules:
+  default:
+    enable::
+    - tcl
```
climbfuji: Which one are you using, tcl or lmod? I guess only one of them. I'd throw out the other one together with the corresponding section below.

AlexanderRichert-NOAA: The default is tcl, but at least some users prefer lmod... I'll see if the tcl modules all work with lmod anyway, in which case I'll throw out lmod.

aerorahul: Noooo. Throw out tcl. Please.

AlexanderRichert-NOAA: Based on my testing it looks like the tcl modules are all compatible with lmod (all the logic is pretty basic), but the reverse is not true. @aerorahul unless you have a specific need for the lua format (like adding certain functionality to the modules), then I'd be inclined to support both tcl and lmod users by using tcl format. I'm not necessarily averse to having it only build the lmod modules, but it makes it a little more likely that I'll get confused questions from users who use the system default...

climbfuji: If your rhel has lmod support, then use lmod please.

climbfuji (approved):
I am fine with this PR, but let's merge the ESMF PR first (with successful CI tests) and then this one.

Actually, it doesn't matter. It doesn't change anything except adding a site config

srherbener (approved):

PR #467, opened 2023-02-09
EMSF beta variant/MAPL modulefile names

AlexanderRichert-NOAA:
This PR makes use of the `snapshot` variant now in NOAA-EMC spack's ESMF package, and also adds ESMF 8.4.0 to the modules config.

### DIFF
--- a/configs/common/modules.yaml
+++ b/configs/common/modules.yaml
@@ -115,12 +115,14 @@ modules:
             'PNG_ROOT': '{prefix}'
       mapl:
         suffixes:
-          ^esmf@8.2.0~debug: 'esmf-8.2.0'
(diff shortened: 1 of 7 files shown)


climbfuji (changes requested):
Your spack submodule pointer is off, otherwise it looks ok

AlexanderRichert-NOAA:
> Your spack submodule pointer is off, otherwise it looks ok

Can you confirm what it should be? As far as I can tell it's pointing to [83743d3](https://github.com/NOAA-EMC/spack/tree/83743d32adb1ea1166248a4f64b93f5d4694c502), which I believe is what I want.

climbfuji (approved):
I am sorry, I was confused by so many changed files, but in fact we haven't updated the submodule pointer in a while. Let's wait for the CI tests and then merge.

Issue #478, opened 2023-02-18
Install and test unified environment on supported HPCs


climbfuji:
**Is your feature request related to a problem? Please describe.**
We need to install and test the unified environment all supported HPCs. A good starting point is the list of preconfigured and configurable (generic) platforms in https://spack-stack.readthedocs.io/en/latest/Platforms.html.

**Describe the solution you'd like**

Left over from previous PR #454:
- [x] Remove `ncl` from global-workflow-env (also affects macos site config)

See epic https://github.com/NOAA-EMC/spack-stack/issues/503 for a list of final installations and successful tests. Consider this issue as completed if all the required boxes are ticket in the epic.

Preliminary testing done beforehand:
 - Orion
    - [x] Update site config to contain Intel, GNU, and legacy Intel 18 (for global workflow) configurations
    - [x] Install unified environment for Intel + GNU
    - [x] Install unified environment for legacy Intel
    - Test unified environment for Intel + GNU
        - [x] JEDI-Skylab
        - [x] UFS Weather Model
        - [x] UFS SRW App

**Additional context**
n/a

ulmononian:
@climbfuji i can install this in the role.epic space on orion, jet, and cheyenne to start. hera may have to wait for an epic-owned installation because our nems account is at capacity.

would you mind sharing the install recipe?

climbfuji:
> @climbfuji i can install this in the role.epic space on orion, jet, and cheyenne to start. hera may have to wait for an epic-owned installation because our nems account is at capacity.
> 
> would you mind sharing the install recipe?

Thanks for volunteering. I think we need to agree on the directory structure and naming conventions first, then create an install recipe that we can more or less copy and paste or automate with Jenkins. I wonder if this can wait until Thursday when we have our spack-stack meeting.

Also, we need to update all site configs to have the compilers configured correctly. That can be a separate PR that goes in first. For example, we have this for Orion (https://github.com/NOAA-EMC/spack-stack/blob/develop/configs/sites/orion/packages.yaml):
```
packages:
  all:
    compiler:: [intel@2022.0.2, intel@18.0.5, gcc@10.2.0]
    providers:
      mpi:: [intel-oneapi-mpi@2021.5.1, intel-mpi@2018.5.274, openmpi@4.0.4]
```
but what we want is
```
packages:
  all:
    compiler:: [intel@2022.0.2, gcc@10.2.0]
    #compiler:: [intel@18.0.5]
    providers:
[...]

PR #480, opened 2023-02-21
Update compiler/MPI entries in site configs for the UE

ulmononian:
As @climbfuji mentioned in #478, the compiler and MPI entry formatting in the sites/packages.yaml files needs to be adjusted for the UE (to properly support both Intel-latest and GNU installs). Further, for sites on which the global-workflow is supported, Intel-18 is needed, so a few sites' (Jet, Hera, Orion, S4) packages.yaml and compilers.yaml files need to be updated to include proper paths to/info. about their respective Intel-18 installations.

I will start with these updates on the following sites: Jet, Hera, Cheyenne, Orion, and Gaea. Perhaps @climbfuji can comment as to what the plan is for installation and testing plan is on the Parallelworks clusters (Azure, AWS, GCP); if UE will be installed there, I can update those too. 

I do not have access to the remaining sites from https://spack-stack.readthedocs.io/en/latest/Platforms.html, so I can't check the compiler/MPI info there.


climbfuji (commented):
This looks good to me. I'll update site configs that you do not have access to in a follow-up PR, possibly as part of the R2D2 MYSQL PR.

ulmononian:
@climbfuji thanks for looking this over. i did not test any of this, but most of it was just re-arranging what was already present plus adding the intel 18 info (which i based on the core intel modules used in the respective global-workflow modules).

climbfuji:
I think it would be good to test this as part of the R2D2 mysql PR, which requires updates to every site config. I am fine with merging your updates as they are, or pulling them into my PR.

ulmononian:
@climbfuji feel free to merge them as is from this pr. i will follow up with your r2d2 updates.

climbfuji (approved):

PR #483, opened 2023-02-24
ufs-weather-model-static: remove fms@2022.01; set hdf5~szip

AlexanderRichert-NOAA:
Avoiding szip-related complications for ufs-weather-model-static since it's not needed. Also removing old fms version.


AlexanderRichert-NOAA:
Also setting `crtm~fix`, @Hang-Lei-NOAA please tell me if that's wrong.

climbfuji (approved):
I am ok with this PR. Actually, I don't care so much about the static ufs-weather-model build, because that only affects acorn, cactus and dogwood, neither of which I have access to, so please review this PR @Hang-Lei-NOAA . Thanks!

PR #489, opened 2023-03-07
Build zstd with the +programs variant

srherbener:
This PR adds a spec to the `configs/common/packages.yaml` that tells spack to always build zstd with the +programs variant. This was done to avoid duplicate zstd library builds since some client packages of zstd want the zstd application binaries (zstd, unzstd, etc) and others don't. Note that it doesn't hurt the packages that don't want the programs variant to include the programs variant for zstd.

### DIFF
--- a/configs/common/packages.yaml
+++ b/configs/common/packages.yaml
@@ -259,3 +259,6 @@
       version: [0.5.1]
     zlib:
       version: [1.2.13]
+    zstd:
+      version: [1.5.2]
+      variants: +programs


configs/common/packages.yaml
```diff
@@ -259,3 +259,6 @@
       version: [0.5.1]
     zlib:
       version: [1.2.13]
+    zstd:
+      version: [1.5.2]
```
climbfuji, line 263: We may not need to pin the version, but given that we are doing it for all the other packages, and that we have to do some cleanup after this release anyway, let's keep it.

climbfuji (approved):
Let's ignore the macos-dom test failures (errors in the github actions yaml), they will be fixed with the next PR. The new "standard" CI tests pass on Ubuntu and macOS, and the changes look straightforward. Thanks for digging into this.

ulmononian (approved):
looks good, @srherbener!

PR #497, opened 2023-03-15
Set cdo to ~openmp by default

### LINKED ISSUE #496: CDO / HDF5+threadsafe / UFS conflict
The latest version of UFS WM cannot run with hdf5 built with +threadsafe, but CDO requires it. I think the solution will be to disable OpenMP and add `+openmp` to the `when` statement for CDO's hdf5+threadsafe requirement, but I need to make sure global workflow is only running CDO serially.

AlexanderRichert-NOAA:
This PR sets ~openmp for cdo so that threadsafe hdf5 isn't required in the global workflow or unified stacks.
Goes with https://github.com/NOAA-EMC/spack/pull/243
Fixes https://github.com/NOAA-EMC/spack-stack/issues/496

### DIFF
--- a/configs/common/packages.yaml
+++ b/configs/common/packages.yaml
@@ -26,6 +26,7 @@
     # work with all compilers that are currently in use in spack-stack
     cdo:
       version: [2.0.5]
+      variants: ~openmp
     cmake:
       version: [3.23.1]
       variants: +ownlibs


climbfuji:
@AlexanderRichert-NOAA Can you pull in develop to resolve the conflict (submodule pointer), please?

PR #501, opened 2023-03-15
Disable threadsafe for hdf5 in common/packages.yaml

AlexanderRichert-NOAA:
When making the changes for cdo/openmp/hdf5+threadsafe, I forgot to switch hdf5 to ~threadsafe...

### DIFF
--- a/configs/common/packages.yaml
+++ b/configs/common/packages.yaml
@@ -94,7 +94,7 @@
       variants: ~fortran ~netcdf
     hdf5:
       version: [1.14.0]
-      variants: +hl +fortran +mpi +threadsafe +szip
+      variants: +hl +fortran +mpi ~threadsafe +szip
     ip:
       version: [3.3.3]
     ip2:


climbfuji:
Didn't we turn off openmp for cdo so that we can keep hdf +threadsafe? Sorry if that's a dumb question, I am getting confused with all these little changes :-(

AlexanderRichert-NOAA:
No, we changed cdo so that hdf5 _can_ be switched over to `~threadsafe` (previously it was demanding threadsafe hdf5 even when it wasn't compiled with openmp). Based on what @mathomp4 said about hdf5/parallelism/thread safe mode, I _believe_ that in the future if for some reason we wanted to have cdo+openmp, we would need to build it on a non-parallel version of hdf5. But I'm still pretty fuzzy on why this is an issue within hdf5.

climbfuji:
Got it ... let's wait for the CI tests to finish, they'll rebuild everything I assume, since you changed the hdf5 variant --> hash.

climbfuji (approved):

PR #506, opened 2023-03-20
Remove python spec details from external package listing on acorn

AlexanderRichert-NOAA:
Having the detailed spec for python in the external packages has caused issues ("Attempted to use external for 'python' which does not satisfy any configured external spec") on account of changing specs. For now I think the safe bet is to remove the variants.

### DIFF
--- a/configs/sites/acorn/packages.yaml
+++ b/configs/sites/acorn/packages.yaml
@@ -71,7 +71,7 @@
     python:
       buildable: false
       externals:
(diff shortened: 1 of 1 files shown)


AlexanderRichert-NOAA:
@climbfuji if we can make this part of the release that'd be great, one less thing to break... I think a couple of people have now run into this issue.

climbfuji:
I recall having to do the same for other platforms. I'll change the target branch for this PR to the release branch and merge.

climbfuji (approved):

Hang-Lei-NOAA (approved):
NCO has the same request.

PR #515, opened 2023-03-24
release/1.3.0: fix mapl module filenames

climbfuji:
## Description

Fix mapl module filenames. Tested for lmod on Dom's macOS, and for tcl on Gaea.

PR #520, opened 2023-03-24
release/1.3.0: Once again, update configs/common/modules.yaml for ESMF/MAPL

climbfuji:
## Description
Turns out that https://github.com/NOAA-EMC/spack-stack/pull/515 fixed the wrong module names for `mapl` with `esmf@8.4.1`, but in turn messed up the ones with `esmf@8.3.0b09`. I couldn't find a clean way to do this except for removing versions that are not being built (and will never be built again with spack-stack), so I think that's ok.

With this PR, I get:
```
mapl/2.22.0-debug-esmf-8.3.0b09-debug
mapl/2.22.0-esmf-8.3.0b09            
mapl/2.35.2-debug-esmf-8.4.1-debug   
mapl/2.35.2-esmf-8.4.1               
```

### DIFF
--- a/configs/common/modules.yaml
+++ b/configs/common/modules.yaml
@@ -116,12 +116,8 @@ modules:
             'PNG_ROOT': '{prefix}'
       mapl:
         suffixes:
-          ^esmf@8.2.0~debug: 'esmf-8.2.0'
-          ^esmf@8.2.0+debug: 'esmf-8.2.0-debug'
           ^esmf@8.3.0b09~debug: 'esmf-8.3.0b09'
           ^esmf@8.3.0b09+debug: 'esmf-8.3.0b09-debug'
-          ^esmf@8.3.0~debug: 'esmf-8.3.0'
-          ^esmf@8.3.0+debug: 'esmf-8.3.0-debug'
           ^esmf@8.4.0~debug: 'esmf-8.4.0'
           ^esmf@8.4.0+debug: 'esmf-8.4.0-debug'
           ^esmf@8.4.1~debug: 'esmf-8.4.1'
@@ -344,12 +340,8 @@ modules:
             'PNG_ROOT': '{prefix}'
       mapl:
         suffixes:
-          ^esmf@8.2.0~debug: 'esmf-8.2.0'
-          ^esmf@8.2.0+debug: 'esmf-8.2.0-debug'
           ^esmf@8.3.0b09~debug: 'esmf-8.3.0b09'
           ^esmf@8.3.0b09+debug: 'esmf-8.3.0b09-debug'
-          ^esmf@8.3.0~debug: 'esmf-8.3.0'
-          ^esmf@8.3.0+debug: 'esmf-8.3.0-debug'
           ^esmf@8.4.0~debug: 'esmf-8.4.0'
           ^esmf@8.4.0+debug: 'esmf-8.4.0-debug'
           ^esmf@8.4.1~debug: 'esmf-8.4.1'


configs/common/modules.yaml
```diff
@@ -116,12 +116,8 @@ modules:
...
         suffixes:
-          ^esmf@8.2.0~debug: 'esmf-8.2.0'
-          ^esmf@8.2.0+debug: 'esmf-8.2.0-debug'
           ^esmf@8.3.0b09~debug: 'esmf-8.3.0b09'
           ^esmf@8.3.0b09+debug: 'esmf-8.3.0b09-debug'
-          ^esmf@8.3.0~debug: 'esmf-8.3.0'
-          ^esmf@8.3.0+debug: 'esmf-8.3.0-debug'
           ^esmf@8.4.0~debug: 'esmf-8.4.0'
           ^esmf@8.4.0+debug: 'esmf-8.4.0-debug'
           ^esmf@8.4.1~debug: 'esmf-8.4.1'
```
AlexanderRichert-NOAA, line 123: What happened to the snapshot specifications? Seems like this is back to the problem of double-applying the esmf suffixes.

climbfuji: With `snapshot=none` it didn't recognize 8.4.1 at all and left it as `mapl/2.35.2` instead of `mapl/2.53.2-esmf-8.4.1`.

Only adding `snapshot=b09` for 8.3.0b09 still had spack match the 8.3.0 release for 8.3.0b09, and it returned `mapl/2.22.0-esmf-8.3.0b09-esmf-8.3.0`.

climbfuji: I think it's ok if we update this module file as we update ESMF versions ...

AlexanderRichert-NOAA: I only just saw #515, not clear on what happened there

AlexanderRichert-NOAA: That doesn't make sense, why would it ignore 8.4.1 with `snapshot=none` but not other versions? Are you sure you didn't leave snapshot=b09 set when installing 8.4.1?

climbfuji: That's a good point. This is the command I shared earlier on slack:
```
    spack install --add --verbose --source esmf@8.4.1~debug+parallelio%intel 2>&1 | tee log.install.unified-env.004.esmf-8.4.1.intel
    spack install --add --verbose --source esmf@8.4.1~debug+parallelio%gcc   2>&1 | tee log.install.unified-env.004.esmf-8.4.1.gcc
    spack install --add --verbose --source esmf@8.4.1+debug+parallelio%intel 2>&1 | tee log.install.unified-env.004.esmf-8.4.1-debug.intel
    spack install --add --verbose --source esmf@8.4.1+debug+parallelio%gcc   2>&1 | tee log.install.unified-env.004.esmf-8.4.1-debug.gcc
```
Probably that inherits the `snapshot=b09` from the default config. Thus, using 
```
spack install --add --verbose --source esmf@8.4.1+debug+parallelio snapshot=none%intel 2>&1 | tee log.install.unified-env.004.esmf-8.4.1-debug.intel
```
would work. Then we'd have to add the same stuff to the mapl install commands:
```
    spack install --add --verbose --source mapl@2.35.2%intel +debug ^esmf@8.4.1%intel +debug+parallelio snapshot=none 2>&1 | tee log.install.unified-env.005.mapl-2.35.2-debug.intel
```

This is getting so tedious and complicated, very easy to make mistakes everywhere.

spack-stack is supposed to be a less than five commands solution for installing entire environments ...

climbfuji:
I'll close this one and revert the previous module PR, because the error was that the esmf@8.4.1 build inherited snapshot=b09 from the common package config.

PR #538, opened 2023-04-10
Bug fix in instructions for Ubuntu 20.04 and 22.04 to install complete mysql

climbfuji:
## Description

Add missing `apt install` command on Ubuntu 20.04 and 22.04. Not needed on CentOS where `mysql-server` contains what is needed. Kriti from JCSDA tested this on her laptop and it works.


srherbener (approved):
Looks good!

PR #539, opened 2023-04-10
Update Acorn site config for 1.3.0 release

AlexanderRichert-NOAA:
No changes needed to common files.

Main changes:
- Switching to gcc@11.2.0 and corresponding Python to avoid some Python-related build issues.
- Eliminating some external packages which create more problems than they solve.
- Using external ecFlow and MySQL.
- Moving build stage directory off of Lustre space.


climbfuji (approved):
Looks reasonable to me, and was used to successfully install the 1.3.0 release on Acorn.

PR #570, opened 2023-05-08
Rename jedi-ewok-env to ewok-env

### LINKED ISSUE #527: Rename jedi-ewok-env to ewok-env
We have jedi-ufs-env, jedi-mpas-env, jedi-fv3-env, soca-env, and jedi-ewok-env. To me, it makes sense to drop the jedi- from the jedi-ewok-env, because ewok-env doesn't depend on jedi to run (it can be used with just soca-env as well).

climbfuji:
## Description

Rename `jedi-ewok-env` to `ewok-env` as described in https://github.com/NOAA-EMC/spack-stack/issues/527.

Also includes spack submodule pointer updates for recent merges.

## Dependencies

Waiting on https://github.com/NOAA-EMC/spack/pull/263

## Issues

Fixes https://github.com/NOAA-EMC/spack-stack/issues/527

## Testing

Only requires CI testing:
- [x] documentation
- [x] Ubuntu
- [x] macOS

### DIFF
--- a/doc/source/Environments.rst
+++ b/doc/source/Environments.rst
@@ -18,7 +18,7 @@ Environments can be constructed in two ways in spack-stack:
            spack add netcdf-c@4.7.4
            spack add ufs-weather-model-env
            spack add ufs-weather-model-env@1.0.0 +debug
-           spack add jedi-ewok-env +solo +r2d2 +ewok 
+           spack add ewok-env
            spack add mapl@2.12.3 +debug ^esmf@8.3.0 +debug
            spack add mapl@2.12.3 +debug ^esmf@8.3.0 +debug
            ...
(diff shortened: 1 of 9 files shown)


ulmononian (approved):

doc/source/Environments.rst
```diff
@@ -18,7 +18,7 @@ Environments can be constructed in two ways in spack-stack:
            spack add netcdf-c@4.7.4
            spack add ufs-weather-model-env
            spack add ufs-weather-model-env@1.0.0 +debug
-           spack add jedi-ewok-env +solo +r2d2 +ewok 
+           spack add ewok-env
```
srherbener, line 21: Why were the variants (solo, r2d2, ewok) removed? I didn't see a change in the companion spack PR that might have accounted for this change, but I realize that the variant specs here might have been redundant. Just want to double check.

climbfuji: I decided not to remove the variants from the ewok-env package yet, just in case we want to use them again in the future. But since we haven't updated these packages in spack in a while, and since they are still not publicly available, I thought it would be best to not enable them here in this example.

srherbener: Got it - that makes sense. Thanks for the explanation.

srherbener (approved):

PR #589, opened 2023-05-19
Add instructions for building openmpi@4.1.5 on Cheyenne, bug fix for pythran 0.12+ with  scipy and older Intel compilers

climbfuji:
## Description

This PR adds instructions for building `openmpi@4.1.5` on Cheyenne with `gcc@10.1.0`. @ulmononian and I were able to run several tests with it, and it seems to be working ok. But it does not fix the problem that we built it for (a GOCART-enabled version of the UFS Weather Model that fails with openmpi@4.1.1 and openmpi@4.1.4). I suggest to keep the instructions just in case we need them again before the machine goes out of service.

The PR also does some cleanup of the documentation for Cheyenne and Casper, and updates the submodule pointer to fix a bug when trying to build `py-scipy` with the Pythran extension enabled using the older Intel compiler (19.x.y) on Cheyenne or Casper. The short answer is: it doesn't work. `py-pythran@0.12.2` itself builds fine with for older and newer Intel compilers, and the Pythran extension in `py-scipy` builds fine with newer Intel compilers - but not with older versions.

## Definition of Done

I tested this on Cheyenne.

### Issue(s) addressed

n/a

## Dependencies

- waiting on https://github.com/JCSDA/spack/pull/271

## Impact

none

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,11 +1,9 @@
 [submodule "spack"]
   path = spack
-  ##url = https://github.com/spack/spack
-  ##branch = develop
-  #url = https://github.com/jcsda/spack
-  #branch = jcsda_emc_spack_stack
-  url = https://github.com/climbfuji/spack
-  branch = bugfix/pythran_012_intel19
+  #url = https://github.com/spack/spack
+  #branch = develop
+  url = https://github.com/jcsda/spack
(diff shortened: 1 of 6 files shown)


ulmononian (approved):
thanks for this work, @climbfuji. looks good.

Issue #593, opened 2023-05-23
Lmod version in Ubuntu 22.04 breaks spack-stack modules: "attempt to call global 'depends_on' (a nil value)"


eap:
## TLDR: 
This is a tracking bug largely to document what is now a known issue. In the immediate term we will document a workaround for this limitation for most users we will recommend the continued use of tcl based modules.

### Issue description:

spack modules are authored for recent versions of lmod (7.0+)  and include the `depends_on` directive to establish relationships between modules. Ubuntu 22.04's packaged version of lmod is 6.6 and does not include the `depends_on` directive. Using lmod 6.6 to load spack modules may lead to the error `Unable to load module ... attempt to call global 'depends_on' (a nil value)`. See full output below:

```
$ module load stack-python
Lmod has detected the following error:  Unable to load module: python/3.10.8
     /home/ubuntu/spack-stack-1.3.1/envs/skylab-4/install/modulefiles/gcc/11.3.0/python/3.10.8.lua : [string "-- -*- lua -*-..."]:16: attempt to call global 'depends_on' (a nil value)

While processing the following module(s):
    Module fullname      Module Filename
    ---------------      ---------------
    python/3.10.8        /home/ubuntu/spack-stack-1.3.1/envs/skylab-4/install/modulefiles/gcc/11.3.0/python/3.10.8.lua
    stack-python/3.10.8  /home/ubuntu/spack-stack-1.3.1/envs/skylab-4/install/modulefiles/gcc/11.3.0/stack-python/3.10.8.lua
```

Verifying Ubuntu's lmod version

```sh
$ apt-cache policy lmod
lmod:
  Installed: (none)
  Candidate: 6.6-0.4
  Version table:
     6.6-0.4 500
        500 http://us-east-2.ec2.archive.ubuntu.com/ubuntu jammy/universe amd64 Packages
```

### Use tcl modules

You can do this by following the instructions for [Ubuntu 20.04 LTS](https://spack-stack.readthedocs.io/en/latest/NewSiteConfigs.html#prerequisites-ubuntu-20-04-one-off) in the spack-stack docs. Shortly, use `apt install -y environment-modules` instead of lmod, and __do not__ run the `sed -i 's/tcl/lmod/g'` in step 7 of "Creating a new environment".

### Manually install lmod

For more info see this:
https://github.com/JCSDA/spack-stack/blob/develop/configs/sites/aws-pcluster/README.md#prerequisites

climbfuji:
[...]

PR #594, opened 2023-05-23
Get rid of ubuntu 22.04 lmod install instructions in `NewSiteConfigs.rst`

eap:
Update (and simplify) ubuntu setup instructions to use tcl modules. The current 22.04 ubuntu lmod is incompatible with spack-generated modules. This is being tracked in issue #593 

Since the only difference between 22.04 and 20.04 was the use of lmod, I've combined the ubuntu instructions.

### DIFF
--- a/doc/source/NewSiteConfigs.rst
+++ b/doc/source/NewSiteConfigs.rst
@@ -327,10 +327,10 @@ The following instructions were used to prepare a basic Red Hat 8 system as it i
 
 This environment enables working with spack and building new software environments, as well as loading modules that are created by spack for building JEDI and UFS software.
 
-Prerequisites: Ubuntu 20.04 (one-off)
+Prerequisites: Ubuntu (one-off)
 -------------------------------------
 
-The following instructions were used to prepare a basic Ubuntu 20.04 system as it is available on Amazon Web Services to build and install all of the environments available in spack-stack (see :numref:`Sections %s <Environments>`).
+The following instructions were used to prepare a basic Ubuntu 20.04 or 22.04 LTS system as it is available on Amazon Web Services to build and install all of the environments available in spack-stack (see :numref:`Sections %s <Environments>`).
 
 1. Install basic OS packages as `root`
 
@@ -344,9 +344,8 @@ The following instructions were used to prepare a basic Ubuntu 20.04 system as i
    apt install -y gcc g++ gfortran gdb
 
    # Environment module support
+   # Note: lmod is available in 22.04, but is out of date: https://github.com/JCSDA/spack-stack/issues/593
    apt install -y environment-modules
-
-   # Do *not* install MPI with yum, this will be done with spack-stack
 
    # Misc
    apt install -y build-essential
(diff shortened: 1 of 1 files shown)


doc/source/NewSiteConfigs.rst
```diff
@@ -483,28 +434,29 @@ It is recommended to increase the stacksize limit by using ``ulimit -S -s unlimi
 7. Set default compiler and MPI library (make sure to use the correct ``gcc`` version for your system and the desired ``openmpi`` version)
 
 .. code-block:: console
-
+   GCC_VERSION="$(gcc --version | grep -o -m1 -P "\d{1,2}\.\d{1,2}\.\d{1,2}$")"
```
climbfuji: Can we please not do this and keep it simple? Just have two lines, one starting with `#For ubuntu 20, do this`, the other `#For ubuntu 22, do this`? Same for redhat.

doc/source/NewSiteConfigs.rst
```diff
@@ -483,28 +434,29 @@ It is recommended to increase the stacksize limit by using ``ulimit -S -s unlimi
...
+   spack config add "packages:all:compiler:[gcc@${GCC_VERSION}]"
+   
    # Example for Red Hat 8 following the above instructions
    spack config add "packages:all:providers:mpi:[openmpi@4.1.4]"
```
climbfuji: Please change this to 4.1.5

doc/source/NewSiteConfigs.rst
```diff
@@ -483,28 +434,29 @@ It is recommended to increase the stacksize limit by using ``ulimit -S -s unlimi
...
-   spack config add "packages:all:compiler:[gcc@11.2.1]"
-
-   # Example for Ubuntu 20.04 following the above instructions
+   
+   # Example for Ubuntu 20.04 or 22.04 following the above instructions
    spack config add "packages:all:providers:mpi:[mpich@4.0.2]"
```
climbfuji: please change this to 4.1.1

doc/source/NewSiteConfigs.rst
```diff
@@ -361,54 +360,6 @@ The following instructions were used to prepare a basic Ubuntu 20.04 system as i
...
-   apt install -y m4
-   apt install -y git
-   apt install -y git-lfs
-   apt install -y unzip
-   apt install -y automake
-   apt install -y xterm
-   apt install -y texlive
-   apt install -y libcurl4-openssl-dev
-   apt install -y libssl-dev
    apt install -y meson
```
climbfuji, line 365: do we know if this ubuntu 22 line works on ubuntu 20? does it hurt?

climbfuji (changes requested):

doc/source/NewSiteConfigs.rst
eap: I don't know if it breaks anything later (I'm not sure how it would), but I do know that there is a meson package for 20.04 (verified with the `ubuntu:20.04` docker image), so this line works in all currently supported LTS releases.

doc/source/NewSiteConfigs.rst
eap: the old version was out of date, showing 11.2.0, currently ubuntu is on 11.3.0. Can we just recommend that people check their version first? Giving definitive guidance here implies that we know the version.

Also; note that the ubuntu `apt-cache policy gcc` is misleading; while it reports 11.2, it is just a redirect package that depends on the real package `gcc-11` which is more up to date.

doc/source/NewSiteConfigs.rst
eap: done

doc/source/NewSiteConfigs.rst
eap: done

doc/source/NewSiteConfigs.rst
climbfuji: yes, that's also fine. We should do the same for macOS then.

eap: done

climbfuji (approved):
Thanks!

ulmononian (approved):
haven't tested these instructions but it appears good.

PR #598, opened 2023-05-23
Add documentation about supported compilers

### LINKED ISSUE #578: gcc@13 not yet supported
People are now getting gcc@13 as defaults in macOS Homebrew. This compiler is problematic. Many packages don't build with them (see some of the conversations in spack-stack over the last few days), and even if the environment builds there are segmentation faults when running code (https://github.com/JCSDA-internal/fv3-bundle/issues/104 if you can access this page).

Of course, in the long run we need to fix the build and runtime problems. But for now, we should implement a guard that prevents users from building with gcc@13 / gfortran@13.

**Additional context**
n/a

climbfuji:
## Description

Add documentation about supported compilers - to address #578. I decided not to code up some complicated logic that we'll have to revert again later. Documentation should do.

Rendered documentation is here: https://spack-stack--598.org.readthedocs.build/en/598/ 

## Definition of Done

- [ ] Render documentation looks correct

### Issue(s) addressed

Closes #578 

## Dependencies

none

## Impact

none

### DIFF
--- a/doc/source/NewSiteConfigs.rst
+++ b/doc/source/NewSiteConfigs.rst
@@ -13,15 +13,15 @@ It is also instructive to peruse the GitHub actions scripts in ``.github/workflo
 +-------------------------------------------+----------------------------------------------------------------------+---------------------------+
 | Compiler                                  | Versions tested/in use in one or more site configs                   | Spack compiler identifier |
 +===========================================+======================================================================+===========================+
-| Intel classic (icc, icpc, ifort)          | 18.0.5.274 to the latest available version in oneAPI 2022.2.1        | `intel@`                  |
+| Intel classic (icc, icpc, ifort)          | 18.0.5.274 to the latest available version in oneAPI 2022.2.1        | ``intel@``                |
 +-------------------------------------------+----------------------------------------------------------------------+---------------------------+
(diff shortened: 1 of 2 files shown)


climbfuji:
readthedocs integration isn't working anymore - @climbfuji to fix

ulmononian (approved):
thanks for adding this in. it's useful to have notes of this.

doc/source/NewSiteConfigs.rst
svahl991, line 1: For those who may skip or skim reading the top of this file, and instead follow by rote the Mac setup instructions further down in the `Install Homebrew` section (line 131 in the modified version of this file): would it make sense to say `brew install gcc@12`, maybe with a pin comment as well, similar to what has been done with `brew install qt@5` at line 141? This can always be removed again when gcc@13 works, but for now it might prevent some errors.

svahl991 (commented):
One suggestion noted.

eap (approved):

doc/source/NewSiteConfigs.rst
climbfuji: That is a good suggestion for now. Just one more place to remember.

climbfuji: Done.

Issue #599, opened 2023-05-23
spack-stack-1.4.0 release work


climbfuji:
## Description

Summary of all tasks for releasing spack-stack-1.4.0.

Starting from the release branches (`git clone -b release/1.4.0 --recursive https://github.com/jcsda/spack-stack`), install on platforms - any bug fixes/documentation updates for specific platforms go into the release branch before the final tag is made.

### Platforms

- [ ] Acorn (Intel): (@AlexanderRichert-NOAA)
    - [ ] redone with `hdf5@1.14.0` in ...
- [x] Casper (Intel): `/glade/work/epicufsrt/contrib/spack-stack/spack-stack-1.4.0-casper/envs/unified-env/` (@climbfuji)
    - [x] redone with `hdf5@1.14.0` in `/glade/work/epicufsrt/contrib/spack-stack/spack-stack-1.4.0-casper/envs/unified-env-v2`
- [x] Cheyenne (Intel): `/glade/work/epicufsrt/contrib/spack-stack/spack-stack-1.4.0/envs/unified-env/` (@climbfuji)
    - [x] redone with `hdf5@1.14.0` in `/glade/work/epicufsrt/contrib/spack-stack/spack-stack-1.4.0/envs/unified-env-v2`
- [x] Discover (GNU & Intel): `/discover/swdev/jcsda/spack-stack/spack-stack-1.4.0/envs/unified-env/` (@climbfuji)
    - [x] redone with `hdf5@1.14.0` in `/discover/swdev/jcsda/spack-stack/spack-stack-1.4.0/envs/unified-env-v2`
- [x] Gaea C4 (Intel): `/lustre/f2/dev/wpo/role.epic/contrib/spack-stack/spack-stack-1.4.0-c4/envs/unified-env` (@climbfuji)
    - [x] redone with `hdf5@1.14.0` in `/lustre/f2/dev/wpo/role.epic/contrib/spack-stack/spack-stack-1.4.0-c4/envs/unified-env-v2`
- [x] Gaea C5 (Intel): `/lustre/f2/dev/wpo/role.epic/contrib/spack-stack/spack-stack-1.4.0-c5/envs/unified-env` (@climbfuji)
    - [x] redone with `hdf5@1.14.0` in `/lustre/f2/dev/wpo/role.epic/contrib/spack-stack/spack-stack-1.4.0-c5/envs/unified-env-v2`
[...]

PR #607, opened 2023-05-26
Update documentation for spack-stack-1.4.0 release part 3

climbfuji:
## Description

Update documentation for spack-stack-1.4.0 release part 3. Orion and updates to AWS AMIs, with additional commands needed to build successfully on Ubuntu 20 and Red Hat 8. We may be able to move those additional settings into `configs/common/packages.yaml` without harm, but I am not going to risk that in the release branch w/o testing on all platforms (plan to do that later in develop).

This fixes a problem with rendering the documentation, because there needs to be a blank line between the "code block console" command and the actual text (otherwise the text doesn't show).

Rendered text is here: https://spack-stack--607.org.readthedocs.build/en/607/

Note that CI tests will not run because changes are in `doc/` only - these suppress the CI runs.

## Definition of Done

- [x] Rendered text looks good
- [x] Installations successful on Orion and AWS EC2 instances

### Issue(s) addressed

Working towards #599 

## Dependencies

none

## Impact

none

### DIFF
--- a/configs/sites/jet/packages.yaml
+++ b/configs/sites/jet/packages.yaml
@@ -59,10 +59,11 @@ packages:
     externals:
     - spec: berkeley-db@5.3.21
       prefix: /usr
-  bison:
-    externals:
-    - spec: bison@3.0.4
-      prefix: /usr
+  # Do not use, can lead to duplicate versions of nco being installed
+  #bison:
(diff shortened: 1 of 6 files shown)


ulmononian (approved):
👍

PR #611, opened 2023-05-27
Change dummy versions in skylab-dev template

climbfuji:
## Description

Change dummy versions in skylab-dev template from unified-dev to skylab-dev to avoid confusion. This is for develop.

It was mentioned that it is confusing to build the skylab-dev template and then load `jedi-fv3-env/unified-dev` for example.

## Issue(s) addressed

n/a

## Dependencies

n/a

## Impact

n/a

## Checklist

n/a


fmahebert (approved):
Not an expert, but looks reasonable. Thanks @climbfuji !

ulmononian (approved):

PR #617, opened 2023-06-01
Update met and metplus default versions

AlexanderRichert-NOAA:
## Description

This PR updates the default versions of MET and METplus in common/packages.yaml

## Checklist

- [x] I have made corresponding changes to the documentation

## Systems
Builds tested on:
- [x] Orion (@AlexanderRichert-NOAA)
- [x] Acorn (@AlexanderRichert-NOAA)
- [x] home PC

### DIFF
--- a/configs/common/packages.yaml
+++ b/configs/common/packages.yaml
@@ -122,7 +122,7 @@
       version: [2.35.2]
       variants: ~shared
     met:
-      version: [11.0.1]
+      version: [11.0.2]
       variants: +python +grib2
     metplus:
       version: [5.0.1]
--- a/configs/common/packages.yaml
+++ b/configs/common/packages.yaml
@@ -122,10 +122,10 @@
       version: [2.35.2]
       variants: ~shared
     met:
-      version: [10.1.1]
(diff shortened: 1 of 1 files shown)


configs/common/packages.yaml
```diff
@@ -122,10 +122,10 @@
       version: [2.35.2]
       variants: ~shared
     met:
-      version: [10.1.1]
+      version: [11.0.1]
```
srherbener: Would it be okay to move to met@11.0.2 (latest release)? I was able to get a build using https://github.com/JCSDA/spack/pull/278 on my M1 Mac to complete successfully using met@11.0.2.

No problem if we need to stay on met@11.0.1. I'm pretty sure that https://github.com/JCSDA/spack/pull/278 will work with 11.0.1, and can test if that if necessary.

Thanks!

AlexanderRichert-NOAA: That sounds good, and I'm glad to hear that worked. I'll do one or two test installations and if it's all good then I'll make the change.

srherbener (approved):
Looks good! Thanks!

PR #618, opened 2023-06-01
Update PR template

### LINKED ISSUE #582: Customize pull request templates?
spack-stack has its own issue templates, but no pull request templates. We are getting the JCSDA default pull request template now that the repository moved.

We should check if the default template from JCSDA works for us or not, and if not create our own template.

AlexanderRichert-NOAA:
This PR adds a spack-stack repo-specific PR template, so that the JCSDA default will no longer be used.

Closes https://github.com/JCSDA/spack-stack/issues/582


srherbener (approved):
Thanks for updating the template! I think this will work well.

.github/pull_request_template.md
```diff
@@ -0,0 +1,28 @@
...
+
+Describe the changes made in this PR and why they are needed.
+
+### Testing
+
+Describe the testing done for this PR.
+
+### Applications affected
+
+List all known applications (UFS WM, JEDI, SRW, etc.) intentionally or unintentionally affected by this PR.
```
ulmononian, line 11: just one question about wording here: if an app or system is unintentionally affected by a pr, would the PR author be aware of that at the time of creating the PR?

climbfuji, line 11: How do I know that my PR affects something unintentionally?

.github/pull_request_template.md
```diff
@@ -0,0 +1,28 @@
...
+
+### Systems affected
+
+List all systems intentionally or unintentionally affected by this PR.
```
climbfuji, line 15: Same here?

.github/pull_request_template.md
```diff
@@ -0,0 +1,28 @@
...
+
+### Dependencies
+
+If there are PRs that need to be merged before or along with this one, please add "waiting for another PR" label and link to the relevant issues/PRs.
```
climbfuji: This label does not exist in the spack-stack repository. How about this:
```suggestion
If there are PRs that need to be merged before or along with this one, please add "- [ ] waiting on LINK_TO_PR" for each of them
```

.github/pull_request_template.md
AlexanderRichert-NOAA: I kept this language from the JCSDA template because I like that it sets the expectation that PR submitters are responsible for not breaking things that aren't relevant to their application of interest (and for that matter, it might prompt them to think harder about the impacts of their changes).

climbfuji: Interesting twist that you have to "defend" the PR template against folks from that very organization ;-) Let's keep that. But we don't have (and don't use) then "waiting for another PR" label for spack-stack, so maybe we can change the wording on just that line and be good with the PR?

.github/pull_request_template.md
climbfuji: @AlexanderRichert-NOAA If you can make this one change, then we can merge this PR and be done with it. Thanks!

.github/pull_request_template.md
climbfuji: See above

climbfuji (approved):

Issue #616, opened 2023-06-01
Concretization issues


AlexanderRichert-NOAA:
We continue to have issues with duplicate packages, especially when an environment is reconcretized part way through an installation. This may be solvable by getting to the point where we can set `unify: true`, or by reporting the issue to the Spack developers. I'm working on producing a test case for the latter, though it's proven more difficult that I expected-- it may have to do with the subtleties of our config file layout, our use of certain external packages, etc.

AlexanderRichert-NOAA:
I appear to have gotten around this problem for the moment on Acorn by using external 'flex'. I still want to understand why this is happening and get a fix from main Spack.

climbfuji:
Looks like we are doing ok right now, whether that is due to updates from mainline spack or adding specs explicitly in spack.yaml.

PR #619, opened 2023-06-02
Add util to show duplicate pkgs in a spack env

AlexanderRichert-NOAA:
## Description

This PR adds a script that checks the output of `spack concretize` for duplicate packages (as well as a 'util/' subdirectory in spack-stack, anticipating we'll want to put other miscellaneous helper scripts, etc. there).

## Issue(s) addressed

Helps with #616

## Checklist


srherbener (approved):
Tried this out on my local clones - works great! Thank you for this addition, it will be very useful.

climbfuji:
Should we incorporate this in the CI tests? And update the documentation in readthedocs to instruct people to run this? Maybe readthedocs should have an additional section on utilities that lists those and describes what they do.

AlexanderRichert-NOAA:
Yes to all of the above, those are on my to-do list once we get UFS using spack-stack :)

PR #621, opened 2023-06-02
Acorn update jun2023

AlexanderRichert-NOAA:
This PR updates Acorn configuration following the 1.4.0 installation.
- The Cray compilers are, strangely, using MPI even in instances where it doesn't seem to be needed, leading to compilation failures when libfabric can't be found. For now, adding libfabric to the list of modules in compilers.yaml resolves this.
- Adding the external 'flex' library resolved an issue of duplicate packages.
- The py-scipy version setting wasn't being respected but the more recent version managed to compile anyway, so I removed that setting.

### DIFF
--- a/configs/sites/acorn/compilers.yaml
+++ b/configs/sites/acorn/compilers.yaml
@@ -12,6 +12,7 @@ compilers:
     - PrgEnv-intel/8.3.3
     - craype/2.7.13
     - intel/19.1.3.304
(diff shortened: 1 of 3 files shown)


Hang-Lei-NOAA (approved):

srherbener (approved):
Thanks for updating the docs with this one!

PR #624, opened 2023-06-06
Add scripts for facilitating installations on acorn

AlexanderRichert-NOAA:
This PR adds a couple utility scripts for doing parallel installations on Acorn through PBS batch jobs.


Hang-Lei-NOAA (approved):
This will convenience the process. Great!

PR #625, opened 2023-06-06
Add +pic variants to common/packages.yaml to enable position-independent code (PIC) compiler flag where needed

eap:
This issue appears to have been included in the 1.4.0 release but it only affects platforms without pixman installed. I suspect this issue hasn't come up because many of our targets already have pixman and preventing spack from attempting to build it. Without this flag we will get the following error during spack install.

```
/usr/bin/ld: .libs/pixman-implementation.o: relocation R_X86_64_TPOFF32 against `fast_path_cache' can not be used when making a shared object; recompile with -fPIC
```

After making this change locally I have observed the following concretize output for pixman, fontconfig, and freetype.

```
^pixman@0.42.2%gcc@11.3.0+pic+shared build_system=autotools arch=linux-ubuntu22.04-
^freetype@2.11.1%gcc@11.3.0+shared build_system=autotools arch=linux-ubuntu22.04-icelake
^fontconfig@2.14.2%gcc@11.3.0+pic build_system=autotools arch=linux-ubuntu22.04-icelake
```


configs/common/packages.yaml
```diff
@@ -71,6 +71,12 @@
     fms:
       version: [2023.01]
       variants: precision=32,64 +quad_precision +gfs_phys +openmp +pic constants=GFS
+    fontconfig:
+      version: [2.14.2]
```
climbfuji, line 75: I wouldn't add a version here, just the variant.

configs/common/packages.yaml
```diff
@@ -71,6 +71,12 @@
...
+      variants: +pic
+    freetype:
+      version: [2.11.1]
```
climbfuji, line 78: I wouldn't add a version here, just the variant.

climbfuji (changes requested):

climbfuji:
@eap Can you address the comments in the PR review please? I think it's ok to merge the `+pic` flags, but we don't want those versions hardcoded. Thanks

configs/common/packages.yaml
```diff
@@ -172,6 +179,9 @@
       variants: +pnetcdf
     parallel-netcdf:
       version: [1.12.2]
+    pixman:
+      version: [0.42.2]
```
climbfuji, line 185:
```suggestion
```

configs/common/packages.yaml
climbfuji, line 78:
```suggestion
```

configs/common/packages.yaml
climbfuji, line 75:
```suggestion
```

climbfuji (changes requested):

eap:
A different way to fix this to edit the `site/compilers.yaml` and add the cc flags manually:

example (includes -O3)
```
flags:
      cflags: -O3 -fPIC
      cxxflags: -O3 -fPIC
      cppflags: -O3 -fPIC
 ```
 
 The last time I edited this I was still struggling to find all packages that needed the variant. I'm not sure if I should attempt to continue or not.

climbfuji:
> A different way to fix this to edit the `site/compilers.yaml` and add the cc flags manually:
> 
> example (includes -O3)
> 
> ```
> flags:
>       cflags: -O3 -fPIC
>       cxxflags: -O3 -fPIC
>       cppflags: -O3 -fPIC
> ```
> 
> The last time I edited this I was still struggling to find all packages that needed the variant. I'm not sure if I should attempt to continue or not.

It's fine to close it. We can fix it later if it comes up again. I don't want to throw blanket `-fPIC` flags over everything using the default compiler flags.

eap:
Sounds good - closing now. We can revisit later if needed. For the moment I've only had this issue on one platform (ubuntu 22.04). If it comes up elsewhere we can reopen.

PR #632, opened 2023-06-20
Update FMS & SCOTCH defaults

AlexanderRichert-NOAA:
This PR updates default specs for fms and scotch in order to support UFS WM.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2,8 +2,8 @@
   path = spack
   #url = https://github.com/spack/spack
   #branch = develop
-  url = https://github.com/AlexanderRichert-NOAA/spack
-  branch = ufs_jun2023
+  url = https://github.com/jcsda/spack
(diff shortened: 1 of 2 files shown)


ulmononian (approved):
approving this now, but may revisit if it turns out that the `esmumps` variant is required...

climbfuji (approved):
Let's see what the CI tests are doing ...

climbfuji:
Because I pushed my submodule pointer update a few seconds before you (https://github.com/JCSDA/spack-stack/pull/635), CI tests are running for that PR now.

AlexanderRichert-NOAA:
@climbfuji I've updated to the current pointer, so would it make sense to cancel those and let these run since they'll need to anyway?

climbfuji:
I've also updated the submodule pointers so same here ... we'll have to run them for both anyway.

PR #636, opened 2023-06-20
Update Acorn config and include bufr@12.0.0 for 1.4.1

AlexanderRichert-NOAA:
Might as well include updates made post-1.4.0 installation to facilitate Acorn installation.

Also, update bufr default to 12.0.0.

### DIFF
--- a/configs/common/packages.yaml
+++ b/configs/common/packages.yaml
@@ -20,7 +20,7 @@
       version: [1.78.0]
       variants: ~atomic +chrono +date_time +exception +filesystem ~graph ~iostreams ~locale ~log ~math ~mpi ~numpy +pic +program_options +python ~random +regex +serialization ~signals +system +test +thread +timer ~wave cxxstd=14 visibility=hidden
     bufr:
-      version: [11.7.1]
+      version: [12.0.0]
       variants: +python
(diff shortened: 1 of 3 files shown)


climbfuji (approved):

climbfuji:
After this, we only need the bufr version update in spack-stack and then we are done with 1.4.1 I think.

AlexanderRichert-NOAA:
Well, I accidentally pushed the bufr update to the same branch, so, I'm just gonna include it in this PR...

climbfuji:
Can you also update it in https://github.com/JCSDA/spack-stack/blob/release/1.4.1/configs/containers/README.md (two places in that file) *after* the CI tests passed, please? No need to rerun CI again for README changes. Thanks!

AlexanderRichert-NOAA:
Yep no problem. Should 1.4.0 go to 1.4.1 as well?

climbfuji:
> Yep no problem. Should 1.4.0 go to 1.4.1 as well?

Ah, yes please. I'll have to make more updates to the container yamls, but we can start installing from the release branch after this PR is merged. I usually create the tags post-install, one never knows what issues pop up during the install on various HPCs that need changes in their configs.

Issue #644, opened 2023-06-23
Configuration template for UFS_UTILS.


HenryRWinterbottom:
**Is your feature request related to a problem? Please describe.**
The UFS global workflow requires several software packages for research and development. One of them is [UFS_UTILS](https://github.com/ufs-community/UFS_UTILS). Currently a template does not exist to build the required software stack for using spack-stack.

**Describe the solution you'd like**
A configuration template beneath `config/templates` specific to the `UFS_UTILS` software packages.

climbfuji:
@HenryWinterbottom-NOAA Are the current templates and environment packages sufficient, or is there more that needs to be done?

HenryRWinterbottom:
@climbfuji I haven't worked with this since our last conversation https://github.com/JCSDA/spack-stack/pull/645.

However, my testing at the time resulted in successful compilation. If the templates are sufficient please use them. My recollection is that you had to load the modules manually. The templates I provided were able to load the necessary modules.

climbfuji:
Thanks for the quick reply. In #645 you mentioned:
> I would like to try and configure this myself using your explanations above. It will provide me an opportunity to learn something new. If it's OK, maybe I just try to learn through this PR? Once it is working as it should we could just move the ufs-utils-env into the release as I think you are suggesting above?
Please revive the PR or open a new one once/if you have updates. Thanks!

HenryRWinterbottom:
Thank you for the reminder.

Can I take a look at this next week? I can see if I can (1) successfully implement your suggestions and (2) prepare things accordingly for a JCSDA PR. 

How does that work for you?

climbfuji:
No rush, just checking what should go into spack-stack 1.5.0 around end of August.

HenryRWinterbottom:
@climbfuji OK, great. I will try to carve out some time next week to take a look. I will follow-up with questions.

Thanks again and have a nice weekend.

climbfuji:
Closed by #707

PR #645, opened 2023-06-23
Spack configuration for UFS_UTILS.

HenryRWinterbottom:
## Description

This adds a configuration to build the software stack for the [UFS_UTILS](https://github.com/ufs-community/UFS_UTILS).

## Issue(s) addressed

This PR resolves issue #644. 

## Checklist

- [ ] I have made corresponding changes to the documentation
- [x] I have run the unit tests before creating the PR


climbfuji (commented):
I am ok doing it this way, another way would be to update the https://github.com/JCSDA/spack/blob/jcsda_emc_spack_stack/var/spack/repos/jcsda-emc-bundles/packages/ufs-utils-env/package.py and add any missing packages (but no version numbers) there, provided that the default versions in `configs/common/packages.yaml` work. This has the advantage that we can add `ufs-utils-env` to `unified-dev` so that we know that any unified environment we build supports ufs-utils.

HenryRWinterbottom:
> I am ok doing it this way, another way would be to update the https://github.com/JCSDA/spack/blob/jcsda_emc_spack_stack/var/spack/repos/jcsda-emc-bundles/packages/ufs-utils-env/package.py and add any missing packages (but no version numbers) there, provided that the default versions in `configs/common/packages.yaml` work. This has the advantage that we can add `ufs-utils-env` to `unified-dev` so that we know that any unified environment we build supports ufs-utils.

@climbfuji Thank you for the information! I'd be happy to use your suggested approach. I just wasn't aware of that preference until now.

If it is OK with your timelines, I will take a stab at it next week.

climbfuji:
> > I am ok doing it this way, another way would be to update the https://github.com/JCSDA/spack/blob/jcsda_emc_spack_stack/var/spack/repos/jcsda-emc-bundles/packages/ufs-utils-env/package.py and add any missing packages (but no version numbers) there, provided that the default versions in `configs/common/packages.yaml` work. This has the advantage that we can add `ufs-utils-env` to `unified-dev` so that we know that any unified environment we build supports ufs-utils.
> 
> @climbfuji Thank you for the information! I'd be happy to use your suggested approach. I just wasn't aware of that preference until now.
> 
> If it is OK with your timelines, I will take a stab at it next week.

No rush. if you want it to be part of the next release, you have about 2 months ;-)

HenryRWinterbottom:
HA! Sounds good! But, we all know how quickly two months can come and go...

ulmononian:
i would second @climbfuji's suggestion, but hey it can also be nice to have both options!

climbfuji:
I would also like to note that using the last spack-stack release 1.4.0, I was able to compile ufs-utils on AWS Parallel Cluster from the unified-env (unified-dev template). It currently doesn't have the `ufs-utils-env` meta package enabled, but all the packages were there (I had to just load the modules manually). So maybe `ufs-utils-env` is already functional and can be used as is in your new template, and for us it's just a case of enabling `ufs-utils-env` in the unified-dev` template.

HenryRWinterbottom:
@climbfuji @ulmononian I am happy to do whatever you think it is best. I am not interested in messing up JCSDA plans. I just needed it for development at the moment. If it is useful for you to use what's in the PR or me pursuing how you suggested, I am happy to do so. If you have something that exists (or will) I am happy to use it as well. I won't be offended. ;-)

Thanks for the feedback guys. This taught me something new about how the spack-stack capabilities are being used. It will certainly be useful for me moving forward.

climbfuji:
It's still going to be useful to have a ufs-utils-env for folks that don't care about the other packages. But maybe it just needs one line in the specs (`- ufs-utils-env`) and, if certain versions are required, those can be requested above in a `packages` section. Let me know if this is not clear and I can help. Thanks!

HenryRWinterbottom:
> It's still going to be useful to have a ufs-utils-env for folks that don't care about the other packages. But maybe it just needs one line in the specs (`- ufs-utils-env`) and, if certain versions are required, those can be requested above in a `packages` section. Let me know if this is not clear and I can help. Thanks!

@climbfuji Thank you for the feedback. I would like to try and configure this myself using your explanations above. It will provide me an opportunity to learn something new. If it's OK, maybe I just try to learn through this PR? Once it is working as it should we could just move the `ufs-utils-env` into the release as I think you are suggesting above?

Thoughts?

climbfuji:
Yes, sounds good!

HenryRWinterbottom:
@climbfuji OK, thank you. I have a few things in the global-workflow that I need to address ASAP -- from last week's conversation, about 2-months remain before the next release, correct? If so, I should be able to get to this by then (hopefully :-)).

PR #652, opened 2023-06-27
Add util for checking installation permissions

AlexanderRichert-NOAA:
### Summary

This PR adds a utility (with doco) for checking the permissions of installations, specifically to ensure that non-owning users/groups can access them.

### Testing

Tested the functionality/logic of the utility in CentOS 8 with some dummy files.

### Applications affected

No impact.

### Systems affected

No impact (except for helping identify permissions problems).

### Dependencies

n/a

### Issue(s) addressed

n/a

### Checklist


AlexanderRichert-NOAA:
@climbfuji let me know what you think-- I don't know if the script is portable for Macs since I can't test, but I also don't know that there will ever be a situation where a Mac user needs to worry about permissions for other users.

climbfuji:
This is a neat utility, however I think what would be even better (or a good additional step) is a simple check in `spack stack create env` if the umask will lead to an inaccessible install (e.g. `0077 = rwx------` or `0027 = rwxr-x---`)?

AlexanderRichert-NOAA:
That sounds good-- I can think of some cases that would make it worth doing both (weird build systems; network FS quirks on Acorn). The one limitation I can think of is that creating and installing the environment could theoretically be done at two different times with different umasks... but short of creating our own install command, it couldn't hurt to have `spack stack create env` spit out a warning for a "bad" umask.

climbfuji:
> That sounds good-- I can think of some cases that would make it worth doing both (weird build systems; network FS quirks on Acorn). The one limitation I can think of is that creating and installing the environment could theoretically be done at two different times with different umasks... but short of creating our own install command, it couldn't hurt to have `spack stack create env` spit out a warning for a "bad" umask.

I can take care of that part, thanks for the feedback.

climbfuji (commented):
Do you want to add this step to the CI? We know that there should be no wrong permissions, but it would be good to run the script nonetheless?

AlexanderRichert-NOAA:
Generally I think using this command will lead us to run chmod on our installations, as opposed to leading to "deep" changes to configs/recipes/codes. So for now I would say we don't need it in the CI, but we should revisit it if we run into issues that need a solution beyond just running chmod.

climbfuji:
I would like to see the script being exercise = tested to not blow up, that's why I was asking. But it's ok as it is, too.

AlexanderRichert-NOAA:
Ah I see what you mean. I think in a separate PR I'll add CI testing for utilities (including testing various cases for the duplicate checker).

climbfuji (approved):

PR #666, opened 2023-07-07
Update default nemsio to 2.5.4

### LINKED ISSUE #664: Update nemsio to 2.5.4
Let's update nemsio to 2.5.4 for v1.5.0 so we can be rid of the w3nco dependency.

AlexanderRichert-NOAA:
### Summary

Update nemsio to 2.5.4, which eliminates w3nco dependency, and is already used by most applications that use nemsio.

### Testing

GLDAS compiles successfully.

### Applications affected

GLDAS (other applications that use nemsio already use 2.5.4).

### Systems affected

n/a

### Dependencies
none

### Issue(s) addressed
Fixes #664

### Checklist

### DIFF
--- a/configs/common/modules.yaml
+++ b/configs/common/modules.yaml
@@ -196,7 +196,7 @@ modules:
         environment:
           set:
             'NEMSIO_INC': '{prefix}/include'
-            'NEMSIO_LIB': '{prefix}/lib'
+            'NEMSIO_LIB': '{prefix}/lib64'
             'MKGFSNEMSIOCTL': '{prefix}/mkgfsnemsioctl'
             'NEMSIO_CHGDATE': '{prefix}/bin/nemsio_chgdate'
             'NEMSIO_GET': '{prefix}/bin/nemsio_get'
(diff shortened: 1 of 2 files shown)


configs/common/modules.yaml
```diff
@@ -196,7 +196,7 @@ modules:
         environment:
           set:
             'NEMSIO_INC': '{prefix}/include'
-            'NEMSIO_LIB': '{prefix}/lib'
+            'NEMSIO_LIB': '{prefix}/lib64/libnemsio.a'
```
climbfuji, line 199: What downstream changes does this require?

AlexanderRichert-NOAA: I'm not aware of any codes currently using this variable. Most codes use CMake and so don't use this variable. The only place I'm aware of this variable being used is GLDAS (which still uses hpc-stack), so this change is basically to make it usable by GLDAS.

climbfuji (approved):

configs/common/modules.yaml
```diff
@@ -441,7 +441,7 @@ modules:
...
```
climbfuji, line 444: I just merged this, but I will note that this won't work on systems that put libraries in `lib` instead of `lib64`.

Issue #609, opened 2023-05-26
tar command failed on Hercules when zlib/1.2.13 is loaded


BijuThomas-NOAA:
With the following spack-stack environment, the tar extraction failed(on Hercules) with the tar file: (/work2/noaa/hwrf/scrub/bthomas/tar_test_spack_stack/hafs.t12z.dropsonde.tar)

```
module use /work/noaa/epic-ps/role-epic-ps/spack-stack/spack-stack-1.4.0-hercules/envs/unified-env/install/modulefiles/Core
module load stack-intel/2021.7.1                                                                    
module load stack-intel-oneapi-mpi/2021.7.1                                                          
module load zlib/1.2.13
```

I can confirm that zlib/1.2.13 is from /work/noaa/epic-ps/role-epic-ps/spack-stack/spack-stack-1.4.0-hercules/envs/unified-env/install/modulefiles/intel/2021.7.1

 If I create the tar file(tar -cvf) within the above spack-stack environment, I have no problem in untar it (tar -xvf).

 I can untar successfully hafs.t12z.dropsonde.tar, if I do not load spack-stack environment(default system environment).

climbfuji:
We will try to fix this on develop before skylab-1.5.0

BijuThomas-NOAA:
@climbfuji Thank you

ulmononian:
just a note on this: @BijuThomas-NOAA and i realized that if one includes `module load tar/1.34` after loading the stack, either in the modulefile for the workflow environment or from the command line, the tar command succeeds.

climbfuji:
@ulmononian Please confirm (later) that https://github.com/JCSDA/spack-stack/pull/674 fixed this issue. Thanks!

ulmononian:
i copied `/work2/noaa/hwrf/scrub/bthomas/tar_test_spack_stack/hafs.t12z.dropsonde.tar` to my user space and did 

```
module purge
module use /work/noaa/epic/role-epic/spack-stack/hercules/modulefiles
module load ecflow/5.8.4
module load mysql/8.0.31

module use /work/noaa/epic/role-epic/spack-stack/hercules/spack-stack-dev-20230717/envs/unified-env/install/modulefiles/Core
module load stack-intel/2021.9.0
module load stack-intel-oneapi-mpi/2021.9.0
module load stack-python/3.9.14

tar -xvf hafs.t12z.dropsonde.tar
```

it successfully untarred the file for me. 

[...]

PR #674, opened 2023-07-17
Remove external zlib from Hercules 

ulmononian:
### Summary

Using an external zlib was causing tar command failures during stack installations. qt requires this external zlib, and ecflow depends on qt at buildtime. however, ecflow does not need qt at runtime, so we can remove the qt dependency from the ecflow modulefile, and then remove the external zlib from the hercules packages.yaml. doing so allows spack to build zlib and the tar failures are fixed.

### Testing

test stack was built on hercules w/ both external zlib & ecflow moduefile qt module dependency removed. tar commands work during stack install. 

### Applications affected

none known.

### Systems affected

hercules.

### Dependencies

none

### Issue(s) addressed

potentially #609 

### Checklist

### DIFF
--- a/configs/sites/hercules/packages.yaml
+++ b/configs/sites/hercules/packages.yaml
@@ -164,10 +164,10 @@ packages:
     externals:
     - spec: wget@1.21.1
       prefix: /usr
-  # Need to use external zlib, because of qt dependence on it (otherwise issues with tar command)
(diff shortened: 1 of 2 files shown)


climbfuji (approved):
Perfect, thank you! This does not need CI testing, obviously. Merging now.

PR #688, opened 2023-07-25
Adding brew install pkg-config for MacOS

ericlingerfelt:
### Summary

When executing `spack install` for spack-stack 1.4.1 on MacOS, the configuration step of building libevent 2.1.12 reports 

`configure: error: openssl is a must but can not be found. You should add the directory containing openssl.pc to the PKG_CONFIG_PATH environment variable, or set CFLAGS and LDFLAGS directly for openssl, or use --disable-openssl to disable support for openssl encryption`

After a Google Search, it was determined that the command `brew install pkg-config` must be executed in order for `openssl` to be resolved locally. 

See https://github.com/libevent/libevent/issues/1136. 

```
checking for openssl/ssl.h... no
configure: error: openssl is a must but can not be found. You should add the directory containing `openssl.pc' to the `PKG_CONFIG_PATH' environment variable, or set `CFLAGS' and `LDFLAGS' directly for openssl, or use `--disable-openssl' to disable support for openssl encryption
==> Error: ProcessError: Command exited with status 1:
    '/Users/eric2/jedi/spack-stack-1.4.0/cache/build_stage/spack-stage-libevent-2.1.12-twy4bcexa6xhzcnrobdeqbfpn3asyenz/spack-src/configure' '--prefix=/Users/eric2/jedi/spack-stack-1.4.0/envs/skylab-5.0.0/install/apple-clang/14.0.3/libevent-2.1.12-twy4bce' '--enable-openssl'

1 error found in build log:
     114    checking for pkg-config... no
     115    checking for library containing SSL_new... no
     116    checking for library containing SSL_new... no
     117    checking openssl/ssl.h usability... no
     118    checking openssl/ssl.h presence... no
     119    checking for openssl/ssl.h... no
  >> 120    configure: error: openssl is a must but can not be found. You should add the directory containing `openssl.pc' to the `PKG_CONFIG_PAT
            H' environment variable, or set `CFLAGS' and `LDFLAGS' directly for openssl, or use `--disable-openssl' to disable support for openss
            l encryption

See build log for details:
  /Users/eric2/jedi/spack-stack-1.4.0/cache/build_stage/spack-stage-libevent-2.1.12-twy4bcexa6xhzcnrobdeqbfpn3asyenz/spack-build-out.txt

...

==> Error: Terminating after first install failure: ProcessError: Command exited with status 1:
    '/Users/eric2/jedi/spack-stack-1.4.0/cache/build_stage/spack-stage-libevent-2.1.12-twy4bcexa6xhzcnrobdeqbfpn3asyenz/spack-src/configure' '--prefix=/Users/eric2/jedi/spack-stack-1.4.0/envs/skylab-5.0.0/install/apple-clang/14.0.3/libevent-2.1.12-twy4bce' '--enable-openssl'
```

### Testing

This has been tested on MacOS 13.4.1. 

### Applications affected

All end result JEDI executables. 

### Systems affected

MacOS

### Checklist


doc/source/NewSiteConfigs.rst
```diff
@@ -138,6 +138,7 @@ These instructions are meant to be a reference that users can follow to set up t
    brew install curl
    brew install cmake
    brew install openssl
+   brew install pkg-config
```
srherbener, line 142: On my mac I've built spack-stack-1.4.0 and the latest develop branches successfully and in both cases spack-stack built pkg-config, libevent and openssl. Does allowing spack-stack to build these packages (pkg-config, libevent, openssl) work for your build of spack-stack-1.4.1? If so, we can keep the homebrew requirements to a minimum.

ericlingerfelt: This did not work using spack-stack 1.4.1.

ericlingerfelt: Update to this test. I also tried to use 1.4.0 with a brand new Homebrew installation and received the same error. I have attached the output of `spack concretize` for 1.4.0. 

[spack_con_output.txt](https://github.com/JCSDA/spack-stack/files/12176841/spack_con_output.txt)

srherbener: The concretize log is indicating that openssl is a dependency for libevent, and neither of those are dependencies for pkg-config. The error you reported occurred during the build of libevent and was complaining about not being able to find openssl.

Comparing your concretize log to my 1.4.0 build indicates these two differences:
- Your mac is intel based, while mine is arm64 based
- You are using openssl@3.1.1, while I am using openssl@1.1.1t

It doesn't make sense to me that installing pkg-config in homebrew makes the libevent problem go away since pkg-config and libevent are unrelated. I wonder if the version of openssl got changed via a chain of dependencies initiated by the pkg-config install.

Do you have the concretize log from the run that worked? I'm assuming that the concretize log attached above is from the run that did not work.

Also, what does `brew list` return, as well as `brew info pkg-config`?

Thanks!

srherbener:
@ericlingerfelt I think I have an idea of what is happening. I was wrong in thinking that libevent and pkg-config are unrelated. I did some research into pkg-config and this is a tool that helps set up command line options for compile steps in build processes. pkg-config is not listed as a dependency of libevent in the spack formula, but I suspect that it is being used to do the compile steps. If this is true, then it does make sense that installing pkg-config with homebrew is fixing the issue with your build.

If this is correct, I think the proper way to fix this is to add pkg-config as a dependency for libevent in the spack formula and add the proper variants/version/etc for pkg-config. In other words I think installing pkg-config with homebrew is a temporary solution.

I think we need to document the need (for now) to install pkg-config with homebrew, but I'm a little bit concerned (without knowing the true root cause of your build issue) if that is the proper way to handle this in the long term. Especially when it seems that many Mac users don't need to homebrew install pkg-config to get spack stack to build.

For this PR, it might be better to write a note after the list of packages to be installed with homebrew to let the user know that they can try installing pkg-config with homebrew if their libevent build does not succeed. Ie, treat it more like a workaround instead of a hard requirement (at this point). Would this be agreeable?

ericlingerfelt:
Hi @srherbener, I completely agree with your approach to issue. Thanks!

ericlingerfelt:
Closing as taken care of in https://github.com/JCSDA/spack-stack/pull/726.

PR #690, opened 2023-07-26
Add documentation for package testing/env chaining

### LINKED ISSUE #680: Add environment chaining features to our Spack extensions
Per our earlier discussion about plans to support updates between releases, we should modify our `create env` and `setup-meta-modules` commands to support chained environments. Specifically, for the former, we can add a `--upstream=/path/to/upstream/spack-stack-1.4.1/envs/unified-env/install` that adds an `upstreams:` section to spack.yaml. We could then either modify `setup-meta-modules` or add a new command which copies over module files from the upstream environment so that they can all be accessed in one place. To handle the situation where, say, we want to try a new version number of ESMF and all of its dependencies, we could make a list of packages consisting of the root spec(s) for the downstream environment (e.g., ESMF) and the output of `spack dependents` (e.g., MAPL, etc.) to clear out those now-obsolete modulefiles and avoid confusion/conflicts.

AlexanderRichert-NOAA:
### Summary

This PR adds documentation for testing packages through environment chaining.

### Testing

n/a

### Applications affected

n/a

### Systems affected

n/a

### Dependencies

https://github.com/JCSDA/spack/pull/298 (+update submodule pointer)

### Issue(s) addressed

Fixes #680

### Checklist


srherbener (commented):
Thanks for adding this! Depending on what you decide to do about my comment in the spack PR (https://github.com/JCSDA/spack/pull/298#pullrequestreview-1548256429) you might want to add a note on what to do if the `--upstream` option fails and you want to re-build your new environment (netcdf-test in your example).

AlexanderRichert-NOAA:
@srherbener thanks, I added a note to the documentation talking about the --upstream option and those warnings.

srherbener (approved):
Looks great! Thanks for adding the note about the `--upstream` option.

AlexanderRichert-NOAA:
FYI it was necessary to use checkout v3 in order to get it to use the submodule from my fork.

climbfuji:
I think this PR should have also updated section 8 and section 11.5.1 @AlexanderRichert-NOAA ?

PR #694, opened 2023-07-27
Add running show_duplicate_packages.py utility after concretize step

### LINKED ISSUE #692: Add instructions to run show_duplicate_packages.py after the concretize step in NewSiteConfigs.rst
It is a good practice to run the show_duplicate_packages.py after concretize to find and clean up duplicated packages.

srherbener:
### Summary

This adds instructions in the NewSiteConfigs.rst file of the documentation that advises the user to save the output of concretize in a log file and then run show_duplicate_packages.py before running install.

### Testing

CI testing

I ran sphinx-build on my local fork to verify that the rendering worked properly.

### Applications affected

None - this is a documentation only update

### Systems affected

None - documentation update

### Dependencies

None

### Issue(s) addressed

Fixes #692 

### Checklist


doc/source/NewSiteConfigs.rst
```diff
@@ -258,9 +258,15 @@ Remember to activate the ``lua`` module environment and have MacTeX in your sear
 
 10. Process the specs and install
 
+It is recommended to save the output of concretize in a log file and inspect that log file using the show_duplicate_packages.py utility.
+This is done to find and fix duplicate package specifications before running the install and discovering the duplicate packages hours later.
+There a few acceptable (i.e., by design) duplicate packages which are: crtm, fms, mapl and esmf.
```
AlexanderRichert-NOAA: Are we still maintaining duplicates of mapl and esmf?

doc/source/NewSiteConfigs.rst
```diff
@@ -258,9 +258,15 @@ Remember to activate the ``lua`` module environment and have MacTeX in your sear
 
 10. Process the specs and install
 
+It is recommended to save the output of concretize in a log file and inspect that log file using the show_duplicate_packages.py utility.
+This is done to find and fix duplicate package specifications before running the install and discovering the duplicate packages hours later.
```
AlexanderRichert-NOAA: Rather than saying "before running the install and discovering..."  maybe we add that duplicates can cause issues at the module creation step (I think apart from that it's generally obvious why we wouldn't want duplicates).

doc/source/NewSiteConfigs.rst
srherbener: I'm not sure, and I'm happy to remove mapl and esmf if we are not supporting duplicates of those anymore. @mathomp4 do you know the answer to this question? Thanks!

AlexanderRichert-NOAA: Based on the unified-dev template it doesn't look like we're adding the extra esmf/mapl versions anymore. Alternatively, we could generalize this to something along the lines of "Note that in the unified environment, there may be deliberate duplicates; consult the specs in spack.yaml to determine which ones are desired."

doc/source/NewSiteConfigs.rst
srherbener: Thanks for this feedback. I think your suggestion is great so I made this change.

doc/source/NewSiteConfigs.rst
srherbener: Great idea! I've made these changes. Thanks

AlexanderRichert-NOAA (approved):

ulmononian (approved):

PR #732, opened 2023-08-22
Allow show_duplicate_packages.py to ignore dup's from different compilers

### LINKED ISSUE #724: show_duplicates utility not working correctly when multiple compilers are used
**Describe the bug**
The util/show_duplicates.py utility doesn't work correctly if multiple compilers are used on a system for one environment, e.g. through the use of compiler matrices in spack.yaml. It reports duplicates because it doesn't know that %intel and %gcc are different things, for example.

A workaround is to use something like this:
```
> cat log.concretize.unified-env.001 | grep '%intel' | util/show_duplicate_packages.py -d
```

**To Reproduce**
Run `util/show_duplicate_packages.py` on a system with multiple compilers for unified-env

**Expected behavior**
Now false duplicates reported

**System:**
Discovered on Hercules, but applies on all systems with two or more compilers

**Additional context**
n/a

AlexanderRichert-NOAA:
### Summary

This PR adds an option to show_duplicate_packages.py that allows it to ignore duplicates that come from different compilers.

### Testing

Tested by hand; updated util_tests.sh (used in CI)

### Applications affected

n/a

### Systems affected

CI

### Dependencies

none

### Issue(s) addressed

Fixes #724

### Checklist

### DIFF
--- a/doc/source/NewSiteConfigs.rst
+++ b/doc/source/NewSiteConfigs.rst
@@ -527,11 +527,12 @@ It is recommended to increase the stacksize limit by using ``ulimit -S -s unlimi
 It is recommended to save the output of concretize in a log file and inspect that log file using the :ref:`show_duplicate_packages.py <Duplicate_Checker>` utility.
 This is done to find and eliminate duplicate package specifications which can cause issues at the module creation step below.
 Note that in the unified environment, there may be deliberate duplicates; consult the specs in spack.yaml to determine which ones are desired.
+See the :ref:`documentation <Duplicate_Checker>` for usage information including command line options.
 
 .. code-block:: console
 
    spack concretize 2>&1 | tee log.concretize
-   util/show_duplicate_packages.py -d log.concretize
+   util/show_duplicate_packages.py -d [-c] log.concretize
    spack install [--verbose] [--fail-fast]
 
 13. Create tcl module files (replace ``tcl`` with ``lmod`` if you have manually installed lmod)
--- a/doc/source/NewSiteConfigs.rst
+++ b/doc/source/NewSiteConfigs.rst
@@ -277,11 +277,12 @@ Remember to activate the ``lua`` module environment and have MacTeX in your sear
 It is recommended to save the output of concretize in a log file and inspect that log file using the :ref:`show_duplicate_packages.py <Duplicate_Checker>` utility.
 This is done to find and eliminate duplicate package specifications which can cause issues at the module creation step below.
 Note that in the unified environment, there may be deliberate duplicates; consult the specs in spack.yaml to determine which ones are desired.
+See the :ref:`documentation <Duplicate_Checker>` for usage information including command line options.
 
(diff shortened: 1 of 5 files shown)


climbfuji (commented):
Is there a reason why you don't want to make this behavior (ignore duplicates from different compilers) the default or the only option?

AlexanderRichert-NOAA:
I can go either way on whether it's the default, so say the word and I'm happy to change it. Initially I figured users would grep as needed, but I'd like to leave the option to use the current behavior 1) because some (meta)packages, namely jedi-tools, should only be built with one or another compiler and we would want to catch it if they weren't, and 2) I wanted to leave room for more exotic situations where, say, an intel-based package depended on a gcc-based package and we would not want to gcc-based dependency to get duplicated.

climbfuji:
> I can go either way on whether it's the default, so say the word and I'm happy to change it. Initially I figured users would grep as needed, but I'd like to leave the option to use the current behavior 1) because some (meta)packages, namely jedi-tools, should only be built with one or another compiler and we would want to catch it if they weren't, and 2) I wanted to leave room for more exotic situations where, say, an intel-based package depended on a gcc-based package and we would not want to gcc-based dependency to get duplicated.

Good points! Let's leave it as is then, we just need to update the documentation (either in this PR or a follow-up PR), it appears in several places:
```
> grep -Rie show_dupl doc
doc/source/NewSiteConfigs.rst:It is recommended to save the output of concretize in a log file and inspect that log file using the :ref:`show_duplicate_packages.py <Duplicate_Checker>` utility.
doc/source/NewSiteConfigs.rst:   util/show_duplicate_packages.py -d log.concretize
doc/source/NewSiteConfigs.rst:It is recommended to save the output of concretize in a log file and inspect that log file using the :ref:`show_duplicate_packages.py <Duplicate_Checker>` utility.
doc/source/NewSiteConfigs.rst:   util/show_duplicate_packages.py -d log.concretize
doc/source/Utilities.rst:show_duplicate_packages.py
doc/source/Utilities.rst:The utility located at util/show_duplicate_packages.py parses the output of ``spack concretize`` and detects duplicates. Usage is as follows:
doc/source/Utilities.rst:   spack concretize | ${SPACK_STACK_DIR}/util/show_duplicate_packages.py
doc/source/Utilities.rst:   ${SPACK_STACK_DIR}/util/show_duplicate_packages.py log.concretize
doc/source/PreConfiguredSites.rst:   spack concretize | ${SPACK_STACK_DIR}/util/show_duplicate_packages.py -d log.concretize
```

climbfuji (approved):
I pushed the change you made for the macos new site config for Linux as well. I checked the rendered documentation and it looks good! Thanks very much.

climbfuji:
I will merge this after https://github.com/JCSDA/spack-stack/pull/722 (so that I don't have to rerun the CI tests there; this PR here is already out of date and will require CI reruns anyway).

Issue #746, opened 2023-08-28
can we include build instructions in the readme?


edwardhartnett:
In the readme it says to look at the readthedocs for install instructuons.

When I look there I get info on modules and how to work with spack-stack on NOAA R&D machines. But no instructions on how to install on a new machine.

So firstly, how do I install spack-stack on my Linux box.

Secondly, can we include those instructions in the README?

climbfuji:
Firstly: look for section https://spack-stack.readthedocs.io/en/latest/NewSiteConfigs.html

Secondly: I am against that, because I don't want to duplicate information and it's a lot that really belongs into the comprehensive documentation.

If we add more links in the README (for fully supported sites, see ... ; for installing spack-stack on a new system see ...) would that work?

edwardhartnett:
To summarize discussion from email, we agree that outside users (i.e. scientists outside NOAA/UCAR) should use spack-stack to install the UFS.

That being the case, we need to provide better instructions, and they should be in the README. Users will, in general, not read much documentation before they get confused and give up, telling everyone they know that spack-stack is terrible and not worth the effort. Such negative talk can really impact funding and use by other projects, just as it can for commercial projects.

Since we want spack-stack to succeed, we will need to support multiple versions of the documentation, including a simple way to get started, and of course the more complex documentation needed to understand all the details.

Since installation process should not change much from release to release, this does not represent an undue maintenance burden. But whatever burden it does represent, is necessary for project success so must be endured.

edwardhartnett:
Using the current instructions I tried this command but it is not working for me:

```
ed@Pooh-Bah:~/spack-stack$ spack stack create env --site linux.default --template unified-env-all --name unified-env.mylinux

Configuring basic directory information ...
  ... script directory: /home/ed/spack-stack/spack/lib/jcsda-emc/spack-stack/stack
  ... base directory: /home/ed/spack-stack/spack/lib/jcsda-emc/spack-stack
  ... spack directory: /home/ed/spack-stack/spack
==> Error: Template: "unified-env-all" does not exist

```

climbfuji:
Where did you get the template name from?

edwardhartnett:
From the readthedocs document. THat's the command that is given here: https://spack-stack.readthedocs.io/en/latest/NewSiteConfigs.html

climbfuji:
A typo in the documentation! The macOS section has it right, template is `unified-dev`
```
[...]

PR #755, opened 2023-08-29
setup.sh, doco changes

### LINKED ISSUE #597: Reset $PATH before installing
I didn't realize until recently that Spack doesn't touch the $PATH variable when generating a build environment, so theoretically what's in the user's path can affect installations. We could set the PATH variable in each site's compilers.yaml to ensure consistency of installations. In most cases, it will probably just be '/usr/bin:/bin' or similar (Spack adds its own compiler paths, etc. onto this setting).

AlexanderRichert-NOAA:
### Summary

This PR:
1. Adds a '-i' option to setup.sh which attempts to reset $PATH to the system default. I believe we should try it out for our 1.5.0 release installations, and if it goes well, add it to the documentation.
2. Updates the overview documentation to clearly spell out the main use cases and where to go in the documentation for each (as well as a note on when to use spack-stack vs. Spack). It also adds a new section that is essentially the old quick start guide. I reordered a few sections; I'm not married to the current order but it feels like the logical order as far as introducing the various use cases in the Overview section.
3. Makes a few miscellaneous documentation changes (typos, stylistic tweaks, etc.).

### Testing

- For setup.sh changes, I tested concretizing and installing packages on Orion, Gaea C4, Hera, and Acorn. On each of Orion and Acorn, I ran one UFS WM RT.
- Doco updates tested with sphinx locally.

### Applications & systems affected

None directly affected, as I intend for us to further test 'setup.sh -i' where the new functionality is off by default (and I think it's fine for it to remain that way).

### Dependencies

none

### Issue(s) addressed

Addresses #746
Fixes #597

### Checklist

### DIFF
--- a/doc/source/CreatingEnvironments.rst
+++ b/doc/source/CreatingEnvironments.rst
@@ -5,7 +5,7 @@ Creating and installing new environments
 
 The following instructions are a simplified version of the instructions found in :numref:`Sections %s <NewSiteConfigs_macOS>` and :numref:`%s <NewSiteConfigs_Linux>`, and should work for systems where spack-stack already has a site configuration, or where minimal configuration is necessary. For more detailed instructions, including how to generate new site configurations and more details of when and how to use external packages, see :numref:`Sections %s <NewSiteConfigs_macOS>` and :numref:`%s <NewSiteConfigs_Linux>`. To chain a new Spack environment to an existing one, such as to test a new package version based on dependencies already installed in the upstream environment, see :numref:`%s <Add_Test_Packages>`.
 
-Note that items in "<item>" should be replaced with the appropriate values (site names, etc.). "<site name>" and "<template name>" must match the names of directories found under 'configs/sites/' and 'configs/templates/', respectively; <environment name> is a user-chosen name for the Spack environment. <path to spack-stack directory> is the root directory created by cloning the spack-stack GitHub repository, which contains configs/, doc/, setup.sh, util/, etc.
+Note that items in "<>" should be replaced with the appropriate values (site names, etc.). *<site name>* and *<template name>* must match the names of directories found under 'configs/sites/' and 'configs/templates/', respectively; *<environment name>* is a user-chosen name for the Spack environment. *<path to spack-stack directory>* is the root directory created by cloning the spack-stack GitHub repository, which contains 'configs/', 'doc/', 'setup.sh', 'util/', etc.
 
 .. code-block:: console
 
--- a/doc/source/CreatingEnvironments.rst
+++ b/doc/source/CreatingEnvironments.rst
@@ -5,7 +5,7 @@ Creating and installing new environments
 
 The following instructions are a simplified version of the instructions found in :numref:`Sections %s <NewSiteConfigs_macOS>` and :numref:`%s <NewSiteConfigs_Linux>`, and should work for systems where spack-stack already has a site configuration, or where minimal configuration is necessary. For more detailed instructions, including how to generate new site configurations and more details of when and how to use external packages, see :numref:`Sections %s <NewSiteConfigs_macOS>` and :numref:`%s <NewSiteConfigs_Linux>`. To chain a new Spack environment to an existing one, such as to test a new package version based on dependencies already installed in the upstream environment, see :numref:`%s <Add_Test_Packages>`.
 
-Note that items in "<>" should be replaced with the appropriate values (site names, etc.). *<site name>* and *<template name>* must match the names of directories found under 'configs/sites/' and 'configs/templates/', respectively; *<environment name>* is a user-chosen name for the Spack environment. *<path to spack-stack directory>* is the root directory created by cloning the spack-stack GitHub repository, which contains 'configs/', 'doc/', 'setup.sh', 'util/', etc.
+Note that items in "<>" should be replaced with the appropriate values (site names, etc.), and items in '[]' are optional. *<site name>* and *<template name>* must match the names of directories found under 'configs/sites/' and 'configs/templates/', respectively; *<environment name>* is a user-chosen name for the Spack environment. *<path to spack-stack directory>* is the root directory created by cloning the spack-stack GitHub repository, which contains 'configs/', 'doc/', 'setup.sh', 'util/', etc.
 
 .. code-block:: console
 
--- a/doc/source/CreatingEnvironments.rst
+++ b/doc/source/CreatingEnvironments.rst
@@ -1,0 +1,33 @@
+.. _CreatingEnvironment:
+
+Creating and installing new environments
+****************************************
+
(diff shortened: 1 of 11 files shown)


doc/source/CreatingEnvironments.rst
```diff
@@ -0,0 +1,33 @@
+.. _CreatingEnvironment:
+
+Creating and installing new environments
```
AlexanderRichert-NOAA, line 3: FYI @edwardhartnett

doc/source/CreatingEnvironments.rst
```diff
@@ -0,0 +1,33 @@
...
+.. code-block:: console
+
+   # To generate a new spack-stack directory structure, run 'git clone --recurse-submodules https://github.com/JCSDA/spack-stack',
+   # optionally with, e.g., '-b release/1.4.1' to specify the version
+
+   . <path to spack-stack directory>/setup.sh
+
+   spack stack create env --site <site name> --template <template name> --name <environment name>
+
+   cd ${SPACK_STACK_DIR}/envs/<environment name>/
```
climbfuji, line 19: The NewSiteConfig instructions don't descend into the env directory but instead use `spack env activate -p envs/<environment name>/`.

Might be better to be consistent.

doc/source/CreatingEnvironments.rst
```diff
@@ -0,0 +1,33 @@
...
+
+Note that items in "<>" should be replaced with the appropriate values (site names, etc.), and items in '[]' are optional. *<site name>* and *<template name>* must match the names of directories found under 'configs/sites/' and 'configs/templates/', respectively; *<environment name>* is a user-chosen name for the Spack environment. *<path to spack-stack directory>* is the root directory created by cloning the spack-stack GitHub repository, which contains 'configs/', 'doc/', 'setup.sh', 'util/', etc.
+
+.. code-block:: console
+
+   # To generate a new spack-stack directory structure, run 'git clone --recurse-submodules https://github.com/JCSDA/spack-stack',
```
climbfuji, line 12: I think in other places we've used `--recursive``, maybe better to use one or the other (I don't care which) for consistency and to avoid confusion

doc/source/CreatingEnvironments.rst
```diff
@@ -0,0 +1,33 @@
...
+
+   cd ${SPACK_STACK_DIR}/envs/<environment name>/
+
+   spack env activate .
+
+   # If not using an existing site configuration, you may wish to modify config files and/or populate them using commands
+   # such as 'spack external find' and 'spack compiler find'.
+   # See https://spack-tutorial.readthedocs.io/en/latest/tutorial_configuration.html
+
+   spack concretize |& tee log.concretize
```
climbfuji: Again, in other places we used (I believe)
```
spack install [--verbose] [--fail-fast] 2>&1 | tee log.install
```

doc/source/NewSiteConfigs.rst
```diff
@@ -514,7 +514,7 @@ It is recommended to increase the stacksize limit by using ``ulimit -S -s unlimi
 11. Edit site config files and common config files, for example to remove duplicate versions of external packages that are unwanted, add specs in ``envs/unified-env.mylinux/spack.yaml``, etc.
 
 .. warning::
-   **Important:** Remove any external ``cmake@3.20`` package from ``envs/unified-env.mylinux/site/packages.yaml``. It is in fact recommended to remove all versions of ``cmake`` up to ``3.20``. Further, on Red Hat/CentOS, remove any external curl that might have been found.
+   Remove any external ``cmake@3.20`` package from ``envs/unified-env.mylinux/site/packages.yaml``. It is in fact recommended to remove all versions of ``cmake`` up to ``3.20``. Further, on Red Hat/CentOS, remove any external curl that might have been found.
```
climbfuji: We should be able to use the new-ish
```
spack external find --scope system --exclude ... --exclude ...
```
syntax in the copy & paste instructions to achieve the same?

doc/source/Overview.rst
```diff
@@ -22,9 +22,16 @@ spack-stack is maintained by:
...
+If you wish to **access an existing spack-stack installation** on a given system, such as to load packages and compile and run a UFS or JEDI application, see the instructions in :numref:`Section %s <UsingSpackEnvironments>`. Ready-to-use spack-stack installations are available on a number of HPC and cloud platforms. A list of those platforms, along with special instructions and caveats for each, can be found in :numref:`Section %s <Preconfigured_Sites>`.
 
-If you are a developer looking to build and run software locally using spack-stack, or if you are looking to install spack-stack on a new platform you can follow the instructions in :numref:`Section %s <NewSiteConfigs>`.
+If you wish to quickly **create a new spack-stack environment (stack installation)**, either on a personal machine or on a supported platform where you are the maintainer, see :numref:`Section %s <CreatingEnvironment>`.
```
climbfuji: At some point we want to refer to the "Maintainers" section that contains detailed instructions for maintainers of tier-1 platforms (or advanced users who want to replicate/update a tier-1 site config).

setup.sh
```diff
@@ -1,9 +1,21 @@
+# Use '-i' when sourcing this file to attempt to reset $PATH to system defaults (i.e., based on /etc/profile),
+# then load any modules (e.g., python) critical for loading and working with Spack.
+# This may have unwanted/unexpected behavior on some systems, so use carefully.
+if [ "$1" == "-i" ]; then
```
climbfuji, line 4: Intersting approach. Let's try this out for the upcoming release. I always thought that we can trim down the instructions for the tier-1 platforms ("load these modules for building spack environments") substantially, which falls into the same category as what you are trying to do here.

doc/source/CreatingEnvironments.rst
AlexanderRichert-NOAA: What would you think of moving to putting log.{concretize,install} under the environment itself? That way the root dir doesn't get cluttered, especially if we install additional environments.

doc/source/CreatingEnvironments.rst
AlexanderRichert-NOAA: 'recursive' is I believe getting deprecated, so I've changed the other instances for consistency.

doc/source/CreatingEnvironments.rst
AlexanderRichert-NOAA: updated

doc/source/CreatingEnvironments.rst
climbfuji: Hmm, it's super convenient to never have to `cd` elsewhere except the spack-stack top-level directory. I also found it useful to see what kind of envs exist and what has been installed. Anyway, no strong preference, but it will require some training since I've more or less hardcoded those steps in my finger muscles (and in my scripts)!

doc/source/NewSiteConfigs.rst
AlexanderRichert-NOAA: updated

doc/source/CreatingEnvironments.rst
AlexanderRichert-NOAA: I have a moderate preference toward putting the logs in their corresponding env directories for organizational reasons, but no strong preference on the cd. Would you prefer, say, `... | tee envs/unified-env/log.concretize`?

climbfuji: I'll leave it up to you - whatever is easier for everyone and less typing.

doc/source/Overview.rst
AlexanderRichert-NOAA: Just added a reference, see what you think

climbfuji: Yes, that's great, thanks very much!

climbfuji (approved):

climbfuji:
Thanks for making these updates!

Issue #757, opened 2023-08-31
Adding and removing several environmental variables for METplus and MET modules


weiwilliam:
I'm using METplus-4.1.1/MET-10.1.1 under spack-stack-1.4.1 to calculate scores.
Adding and removing several environmental variables will be very helpful to use it later.

1. Removing MET_BASE 
With $MET_BASE exported, METplus/MET will encounter the error below
```
OUTPUT:
WARNING:
WARNING: get_filenames() -> can't stat "/data/prod/jedi/spack-stack/spack-stack-1.4.1/envs/unified-env/install/intel/2021.5.0/met-10.1.1-oy2shu6/table_files"
WARNING:
WARNING:
WARNING: get_filenames() -> can't stat "/data/prod/jedi/spack-stack/spack-stack-1.4.1/envs/unified-env/install/intel/2021.5.0/met-10.1.1-oy2shu6/table_files"
WARNING:
ERROR  :
ERROR  : recursive_envs() -> unable to open input file "/data/prod/jedi/spack-stack/spack-stack-1.4.1/envs/unified-env/install/intel/2021.5.0/met-10.1.1-oy2shu6/config/ConfigConstants"
ERROR  :
```
This was discussed on METplus discussion last year.
[METplus Discussion#1810](https://github.com/dtcenter/METplus/discussions/1810#discussion-4398627)
By unsetting MET_BASE after loading modules, it can successfully generate statistics files.

2. Adding MET_INSTALL_PATH (or something similar)
By exporting this path, the 'MET_INSTALL_DIR' in main.conf for run_metplus.py can be pointed to this variable by {ENV[MET_INSTALL_PATH]}.
For example, MET_INSTALL_PATH can be '/data/prod/jedi/spack-stack/spack-stack-1.4.1/envs/unified-env/install/intel/2021.5.0/met-10.1.1-oy2shu6' on S4.

3. Adding METPLUS_PATH
Similar to point 2, the PARM_BASE in main.conf can be pointed to {ENV[METPLUS_PATH]}.
METPLUS_PATH can be '/data/prod/jedi/spack-stack/spack-stack-1.4.1/envs/unified-env/install/intel/2021.5.0/metplus-4.1.1-l75c2q5'

Then the conf files for METplus can be prepared without hard-coded library paths.

climbfuji:
[...]

PR #760, opened 2023-09-05
Enable MET graphics and other variants, enable `gsi-env` in `unified-dev` template

### LINKED ISSUE #689: [INSTALL] MET with graphics variant (or more)
Note: Please search to see if an issue already exists for the software you are requesting to be installed.

**Which software in the stack would you like installed?**

MET

**What is the version/tag of the software?**

11.0.2 at least

**What compilation options would you like set?**

```
+python +grib2 +graphics
```
Or maybe even:
```
+python +grib2 +graphics +lidar2nc +modis
```

**Which machines would you like to have the software installed?**

NCCS Discover

**Any other relevant information that we should know to correctly install the software??**

You are the experts and @AlexanderRichert-NOAA seems to have all variants I need in there.

**Additional context**

I work with a colleague at GMAO who is looking into and using MET, and he builds his own (which as we all know is a bit of "fun"). But I thought, hey, MET must be in spack-stack, and it is and I found a dev install with MET 11.0.2. But per the [`packages.yaml`](https://github.com/JCSDA/spack-stack/blob/develop/configs/common/packages.yaml) it's built as:
```yaml
    met:
      version: [11.0.2]
      variants: +python +grib2
```

On the other hand, my user here built with:

```
--enable-grib2 --enable-modis --enable-mode_graphics --enable-lidar2nc --enable-python
```

Now, he *knows* he needs the graphics option. He's not sure about the other two, but I suppose if it's possible to enable all the variants, that's "safest"

### LINKED ISSUE #739: For spack-stack@1.5.0: enable gsi meta-package in unified-dev template
All in the title.

See above

**Additional context**
n/a

climbfuji:
### Summary

1. Add MET variants `+graphics +lidar2nc +modis` to CI tests and to Discover site config.
2. Enable `gsi-env` in `unified-dev` template
3. Added `pic` variant to a number of packages in `configs/common/packages.yaml`, required to build MET and its dependencies on Ubuntu CI with the Intel compiler. Credits for these changes go to @eap who first proposed them, but the PR went stale and never got merged (https://github.com/JCSDA/spack-stack/pull/625). I don't think that adding these variants hurts building on systems/in environments were static libraries are used, but maybe @AlexanderRichert-NOAA can test building the static ufs-weather-model environment (which, by the way, needs updating before the release) on Acorn?

### Testing

- [x] CI
- [x] @climbfuji's macOS

### Applications affected

CI, Discover when building unified-dev

### Systems affected

CI, Discover

### Dependencies

- [x] waiting on https://github.com/JCSDA/spack/pull/316

### Issue(s) addressed

Resolves https://github.com/JCSDA/spack-stack/issues/739
Resolves https://github.com/JCSDA/spack-stack/issues/689

### Checklist
- [yes+no] These changes have been tested on the affected systems and applications (not yet on Discover, but in CI and on @climbfuji's macOS; if testing on DIscover requires any site-specific updates, these will be made on the release branches as part of the roll-out).

### DIFF
--- a/.github/workflows/macos-ci-aarch64.yaml
+++ b/.github/workflows/macos-ci-aarch64.yaml
@@ -92,6 +92,9 @@ jobs:
           spack config add "packages:all:compiler:[apple-clang@14.0.0]"
           sed -i '' "s/\['\%aocc', '\%apple-clang', '\%gcc', '\%intel'\]/\['\%apple-clang'\]/g" $ENVDIR/spack.yaml
 
+          # Add additional variants for MET packages, different from config/common/packages.yaml
+          spack config add "packages:met:variants:+python +grib2 +graphics +lidar2nc +modis"
+
           # Concretize and check for duplicates
           spack concretize 2>&1 | tee log.concretize.apple-clang-14.0.0
           ${SPACK_STACK_DIR}/util/show_duplicate_packages.py -d log.concretize.apple-clang-14.0.0 -i fms -i crtm
--- a/.github/workflows/macos-ci-x86_64.yaml
+++ b/.github/workflows/macos-ci-x86_64.yaml
@@ -85,6 +85,9 @@ jobs:
           spack config add "packages:all:compiler:[apple-clang@14.0.0]"
           sed -i '' "s/\['\%aocc', '\%apple-clang', '\%gcc', '\%intel'\]/\['\%apple-clang'\]/g" $ENVDIR/spack.yaml
 
+          # Add additional variants for MET packages, different from config/common/packages.yaml
+          spack config add "packages:met:variants:+python +grib2 +graphics +lidar2nc +modis"
+
           # Concretize and check for duplicates
           spack concretize 2>&1 | tee log.concretize.apple-clang-14.0.0
           ${SPACK_STACK_DIR}/util/show_duplicate_packages.py -d log.concretize.apple-clang-14.0.0 -i fms -i crtm
--- a/.github/workflows/ubuntu-ci-x86_64.yaml
+++ b/.github/workflows/ubuntu-ci-x86_64.yaml
@@ -123,6 +123,9 @@ jobs:
           spack config add "packages:all:compiler:[intel@2022.1.0]"
           sed -i "s/\['\%aocc', '\%apple-clang', '\%gcc', '\%intel'\]/\['\%intel'\]/g" $ENVDIR/spack.yaml
 
+          # Add additional variants for MET packages, different from config/common/packages.yaml
+          spack config add "packages:met:variants:+python +grib2 +graphics +lidar2nc +modis"
+
           # Concretize and check for duplicates
           spack concretize 2>&1 | tee log.concretize.intel-2022.1.0
           ${SPACK_STACK_DIR}/util/show_duplicate_packages.py -d log.concretize.intel-2022.1.0 -i fms -i crtm
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,9 +1,11 @@
 [submodule "spack"]
   path = spack
-  #url = https://github.com/spack/spack
-  #branch = develop
-  url = https://github.com/jcsda/spack
-  branch = jcsda_emc_spack_stack
+  ##url = https://github.com/spack/spack
+  ##branch = develop
+  #url = https://github.com/jcsda/spack
+  #branch = jcsda_emc_spack_stack
+  url = https://github.com/climbfuji/spack
+  branch = feature/met_graphics
 [submodule "doc/CMakeModules"]
   path = doc/CMakeModules
   url = https://github.com/noaa-emc/cmakemodules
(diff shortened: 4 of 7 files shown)


AlexanderRichert-NOAA:
I'd been meaning to follow up on the MET graphics/lidar/modis question. This adds a number of dependencies, and I'm still not clear on whether it was needed by the original user, or if that original user's needs were based on a UFS/JEDI application. Would it be good enough to have the functionality in the recipes in case users want to build their own MET, but not enable it in the unified environment unless/until it's needed for UFS or JEDI?

climbfuji:
> I'd been meaning to follow up on the MET graphics/lidar/modis question. This adds a number of dependencies, and I'm still not clear on whether it was needed by the original user, or if that original user's needs were based on a UFS/JEDI application. Would it be good enough to have the functionality in the recipes in case users want to build their own MET, but not enable it in the unified environment unless/until it's needed for UFS or JEDI?

I am pretty sure that JEDI will want those options soon, too. We just started adding MET to our workflow and experimenting with its features. But things like modis and lidar are likely going to be needed. I've got it all working now, and the additional dependencies aren't that many. But I'd also like to hear from @mathomp4 on what he thinks.

climbfuji:
We could, for the time being, enable these variants only in CI so that at least we know it's working?

natalie-perlin (commented):
Looks good to me!
As to the MET package, how the modules are formed, is it possible to add another env. variable? For example, MET_INSTALL_DIR variable is one of the commonly used in MET package; it is identical to MET_ROOT but MET_INSTALL_DIR is hard-coded into the package, so it needs to be set in either case.
Currently installed MET packages used with SRW also have another variable defined in the modulefile, MET_BIN_EXEC. This variable, however, may not be required for MET, and set for convenience with the use with SRW workflow.

climbfuji:
> Looks good to me! As to the MET package, how the modules are formed, is it possible to add another env. variable? For example, MET_INSTALL_DIR variable is one of the commonly used in MET package; it is identical to MET_ROOT but MET_INSTALL_DIR is hard-coded into the package, so it needs to be set in either case. Currently installed MET packages used with SRW also have another variable defined in the modulefile, MET_BIN_EXEC. This variable, however, may not be required for MET, and set for convenience with the use with SRW workflow.

Thanks for reviewing. I will let @weiwilliam and @AlexanderRichert-NOAA weigh in on which variables should be set or not, since they recently discussed this in detail (in https://github.com/JCSDA/spack-stack/issues/757).

mathomp4:
I mean, I only asked if you could add the variants as a local user here builds MET with them. If spack-stack can't, I suppose the user can keep building by hand. But we'd love if you could!

climbfuji:
> I mean, I only asked if you could add the variants as a local user here builds MET with them. If spack-stack can't, I suppose the user can keep building by hand. But we'd love if you could!

I think we can, the question is whether we make them the default everywhere or just in CI + Discover until JCSDA needs them everywhere (if that is the case).

mathomp4:
> I think we can, the question is whether we make them the default everywhere or just in CI + Discover until JCSDA needs them everywhere (if that is the case).

Ah. Okay. Well, I suppose that if/once it's on Discover, I can have my user try it out with his workflow.

weiwilliam:
@climbfuji @natalie-perlin 
As I know, MET_INSTALL_DIR is defined when the spack-stack (at least JCSDA fork) installs METplus.
https://github.com/JCSDA/spack/blob/a041f7d05c59a90c2a37b281ac73052c9786a130/var/spack/repos/builtin/packages/metplus/package.py#L54-L56
```
        metplus_config.filter(
            r"MET_INSTALL_DIR = /path/to", "MET_INSTALL_DIR = {}".format(met_prefix)
        )
```
So, it means users don't need to specify the MET_INSTALL_DIR in main.conf anymore.
It is in the default.conf under $metplus_ROOT/parm/metplus_conf folder.

@natalie-perlin is MET_BIN_EXEC pointed to run_metplus.py?
It could be helpful to ensure the use of the correct python script but not required because the $metplus_ROOT/ush will be prepended in the $PATH.
Then direct running run_metplus.py should always use the right one.

@climbfuji For the application I have (GridStat), the current set of env variables is sufficient to use. I'm not sure if there are other env variables needed for other applications (I believe it's not).

climbfuji:
Thanks @weiwilliam. @natalie-perlin looks like everything is set correctly?

climbfuji:
@AlexanderRichert-NOAA Please let me know if you would like me to revert the default variants for MET and enable them only in CI and on Discover for now, or leave as is. Thanks!

AlexanderRichert-NOAA:
@climbfuji if you think those variants are likely to be used for JEDI then it's no problem. For me the main thing is that our general approach is to not needlessly add more dependencies, but I'm not worried about this particular instance.

climbfuji:
I think a compromise is the best solution for now; build in CI and on Discover (this is also where the JEDI folks exploring MET are working mostly, because they deal with geos, aerosol and all that atmospheric composition stuff). I'll make the change!

climbfuji:
@AlexanderRichert-NOAA @natalie-perlin @mathomp4 @ulmononian Ready for your final review.

AlexanderRichert-NOAA (approved):

PR #320, opened 2022-08-26
Update GNU container recipe, add Clang container

climbfuji:
## Description

1. Update ubuntu-gnu-openmpi docker recipe:
    - use the default `gnu@9.4.0` compiler instead of `gnu@10.0.0` - there were problems with both of them in the container build
    - use a few more external packages such as `qt@5` and `wget`
    - set environment variable `TZ` to avoid the container build hanging indefinitely in the final phase when installing `tzdata`
    - do not strip binaries after build to avoid warnings at end of build process
    - Use new functionality of extra instructions in pre-build and pre-final phases
2. Add ubuntu-clang-mpich docker recipe, almost identical to ubuntu-gnu-openmpi with minimal necessary changes
3. Add `README.md` in `configs/containers` to list the specs to install for each spack-stack release and application

**Notes**
- CI tests passed for https://github.com/NOAA-EMC/spack-stack/pull/320/commits/5f89cd003c7360c73a09e66f12d9e20610197aaf (this includes installing `py-f90nml` and `py-xarray`)
- Also fixed two small errors in `doc/source/Platforms.rst`
- Waiting on https://github.com/NOAA-EMC/spack/pull/162 (merged)
- Waiting on https://github.com/NOAA-EMC/spack/pull/169 (merged)
- Fixes https://github.com/NOAA-EMC/spack/issues/163

PR #764, opened 2023-09-06
Update srw-dev template specs.

### LINKED ISSUE #736: For spack-stack-1.5.0: update ufs-srw-app templates
## Description

The ufs-srw-app-* templates have fallen behind and will continue to do so, since we use unified-dev and skylab-dev on all systems.

## Describe the solution you'd like
Someone needs to go and update the versions/add missing packages in the templates before the 1.5.0 release to match what is built in the unified environment.

Additional context
n/a

ulmononian:
### Summary

update the package versions in the ufs-srw-dev template so that they match what is found in https://github.com/ufs-community/ufs-srweather-app/blob/develop/modulefiles/srw_common.lua. i'll note that the srw is lagging the weather model and `spack-stack/1.5.0` pretty signiciantly at this point, so while building from this template will enable testing of the current srw-dev branch w/ spack-stack, the srw will need to be updated to work w/ `spack-stack/1.5.0`

### Testing

template build on orion. 

### Applications affected

SRW

### Systems affected

not machine specific, so should not affect any

### Dependencies

[spack #320](https://github.com/JCSDA/spack/pull/320)

### Issue(s) addressed

Fixes #736 

### Checklist

### DIFF
--- a/configs/templates/ufs-srw-dev/spack.yaml
+++ b/configs/templates/ufs-srw-dev/spack.yaml
@@ -32,14 +32,14 @@ spack:
   - hdf5@1.10.6
   - netcdf-c@4.7.4
   - netcdf-fortran@4.5.4
-  - parallelio@2.5.3
+  - parallelio@2.5.7
   - esmf@8.3.0b09
-  - fms@2022.01
+  - fms@2022.04
   - bufr@11.7.0
   - bacio@2.4.1
-  - crtm@2.3.0
+  - crtm@2.4.0
   - g2@3.4.5
-  - g2tmpl@1.10.0
+  - g2tmpl@1.10.2
   - ip@3.3.3
   - sp@2.3.3
   - w3nco@2.4.1
@@ -54,5 +54,5 @@ spack:
   - wrf-io@1.2.0
   - ncio@1.1.2
   - gsi-ncdiag@1.1.1
-  - met@10.1.0
-  - metplus@4.1.0
+  - met@10.1.2
+  - metplus@4.1.3
--- a/configs/templates/ufs-srw-dev/spack.yaml
+++ b/configs/templates/ufs-srw-dev/spack.yaml
@@ -26,16 +26,16 @@ spack:
     unify: true
 
   specs:
-  - jasper@2.0.25
-  - zlib@1.2.11
+  - jasper@2.0.32
+  - zlib@1.2.13
   - libpng@1.6.37
-  - hdf5@1.10.6
-  - netcdf-c@4.7.4
-  - netcdf-fortran@4.5.4
-  - parallelio@2.5.7
-  - esmf@8.3.0b09
-  - fms@2022.04
-  - bufr@11.7.0
+  - hdf5@1.14.0
+  - netcdf-c@4.9.2
+  - netcdf-fortran@4.6.0
+  - parallelio@2.5.10
+  - esmf@8.4.2
+  - fms@2023.01
+  - bufr@12.0.0
   - bacio@2.4.1
   - crtm@2.4.0
   - g2@3.4.5
@@ -44,8 +44,8 @@ spack:
   - sp@2.3.3
   - w3nco@2.4.1
   - gftl-shared@1.5.0
-  - yafyaml@0.5.1
-  - mapl@2.22.0
+  - yafyaml@1.1.0
+  - mapl@2.35.2
   - nemsio@2.5.4
   - sfcio@1.4.1
   - sigio@2.3.2
@@ -54,5 +54,5 @@ spack:
   - wrf-io@1.2.0
   - ncio@1.1.2
   - gsi-ncdiag@1.1.1
-  - met@10.1.2
-  - metplus@4.1.3
+  - met@11.1.0
+  - metplus@5.1.0
--- a/.gitmodules
+++ b/.gitmodules
@@ -2,8 +2,8 @@
   path = spack
   #url = https://github.com/spack/spack
   #branch = develop
-  url = https://github.com/ulmononian/spack
-  branch = feature/update_srw_env
+  url = https://github.com/JCSDA/spack-stack
+  branch = jcsda_emc_spack_stack
 [submodule "doc/CMakeModules"]
   path = doc/CMakeModules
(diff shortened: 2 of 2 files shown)


configs/templates/ufs-srw-dev/spack.yaml
```diff
@@ -32,14 +32,14 @@ spack:
   - hdf5@1.10.6
   - netcdf-c@4.7.4
   - netcdf-fortran@4.5.4
-  - parallelio@2.5.3
+  - parallelio@2.5.7
```
climbfuji:
```suggestion
  - parallelio@2.5.10
```

configs/templates/ufs-srw-dev/spack.yaml
```diff
@@ -32,14 +32,14 @@ spack:
...
   - esmf@8.3.0b09
-  - fms@2022.01
+  - fms@2022.04
```
climbfuji:
```suggestion
  - fms@2023.01
```

configs/templates/ufs-srw-dev/spack.yaml
```diff
@@ -54,5 +54,5 @@ spack:
   - wrf-io@1.2.0
   - ncio@1.1.2
   - gsi-ncdiag@1.1.1
-  - met@10.1.0
-  - metplus@4.1.0
+  - met@10.1.2
```
climbfuji:
```suggestion
  - met@11.1.0
```

configs/templates/ufs-srw-dev/spack.yaml
```diff
@@ -54,5 +54,5 @@ spack:
...
+  - metplus@4.1.3
```
climbfuji:
```suggestion
  - metplus@5.1.0
```

configs/templates/ufs-srw-dev/spack.yaml
```diff
@@ -32,14 +32,14 @@ spack:
   - hdf5@1.10.6
```
climbfuji:
```suggestion
  - hdf5@1.14.0
```

climbfuji (dismissed):
I've started comparing versions, but I think what you'll need to do is concretize the unified-dev environment and then check/update all versions in the ufs-srw-dev template?

ulmononian:
i guess i am missing the point then. i thought this was supposed to match what is in the srw develop version? if this is supposed to just use the versions in spack-stack unified dev, then this template will just be many package versions ahead of the srw software itself.

climbfuji:
Likely I am missing the point ;-) You certainly know better, so please ditch all my suggestions. Can you test building with the versions you want?

ulmononian:
it may be a fundamental misunderstanding on my part of the purpose of these templates versus the virtual env packages (like we use in unified env that just use the most up to date package versions).

to me, it seems that if users want to transition to spack-stack unified dev, they should use the unified env we install and port the software. if they just want a template to build the packages they currently use, then use a template. otherwise, i don't think i am getting the point of the templates...why not just use the virtual envs then?

ulmononian:
frankly, i _feel_ that the template _should_ use the newest versions possible (like your review suggests) to drive transition of software to spack-stack, but it just comes back to purpose of template versus virtual env. package.

climbfuji:
Yes, I think that's what you will want to do in the future. Go with unified-dev on the HPCs. But, that may be overkill for someone who just wants to build the ufs-srw-app on a laptop or in a container. If you look at the skylab-dev template, that's how we address the problemf or JEDI: a stripped-down version of unified-dev that uses the same package versions as unified-dev but only includes the meta packages needed.

ulmononian:
> Yes, I think that's what you will want to do in the future. Go with unified-dev on the HPCs. But, that may be overkill for someone who just wants to build the ufs-srw-app on a laptop or in a container. If you look at the skylab-dev template, that's how we address the problemf or JEDI: a stripped-down version of unified-dev that uses the same package versions as unified-dev but only includes the meta packages needed.

ok i see. that makes sense to me. my remaining question then is this: a user can declare their environment name as `--unified-dev` but then just remove all the virtual env. packages listed in the unified-dev `spack.yaml` _except_ for, say, `ufs-srw-app`, so then only what is needed for the srw app is installed. is this not equivalent to what happens with the template, except that in the scenario i just described, the user gets all the latest package versions and not the static template versions? not sure what is preferable.

climbfuji:
> > Yes, I think that's what you will want to do in the future. Go with unified-dev on the HPCs. But, that may be overkill for someone who just wants to build the ufs-srw-app on a laptop or in a container. If you look at the skylab-dev template, that's how we address the problemf or JEDI: a stripped-down version of unified-dev that uses the same package versions as unified-dev but only includes the meta packages needed.
> 
> ok i see. that makes sense to me. my remaining question then is this: a user can declare their environment name as `--unified-dev` but then just remove all the virtual env. packages listed in the unified-dev `spack.yaml` _except_ for, say, `ufs-srw-app`, so then only what is needed for the srw app is installed. is this not equivalent to what happens with the template, except that in the scenario i just described, the user gets all the latest package versions and not the static template versions? not sure what is preferable.

It's a question of convenience. For ufs-srw-app, since it only needs one virtual-env, it may seem like an overkill to have a separate template. For skylab-dev, which combines several virtual-envs and a few additional fms/crtm versions, the burden would be too much for the JEDI developers to know what to remove from unified-dev.

ulmononian:
updated this template to reflect how the srw app env concretizes within the unified environment.

climbfuji:
@ulmononian Please update the spack submodule pointer and also pull in the latest jcsda/develop into your spack-stack branch. Thanks!

ulmononian:
> @ulmononian Please update the spack submodule pointer and also pull in the latest jcsda/develop into your spack-stack branch. Thanks!

done.

climbfuji (changes requested):
I am not showing the submodule pointer update to hash https://github.com/JCSDA/spack/commit/fedc6ad963cfb9f8295f708be73e215ff3a2b88e in your PR.

ulmononian:
ok i think it is done now.

climbfuji (approved):

PR #773, opened 2023-09-12
Restrict eckit version for intel 19 for acorn; remove acorn ext python

AlexanderRichert-NOAA:
### Summary

This PR sets eckit one version back for intel 19 because otherwise odc can't build. Also, it removes the external python on acorn. If any other systems have similar issues, we can move the eckit logic to common/packages.yaml.

### Testing

Installed with intel 2022 and 19 on Acorn, and the UFS WM control_CubedSphereGrid RT builds and runs successfully.

### Applications affected

UFS WM

### Systems affected

Acorn

### Dependencies

none

### Issue(s) addressed

none

### Checklist

### DIFF
--- a/configs/sites/acorn/packages.yaml
+++ b/configs/sites/acorn/packages.yaml
@@ -22,13 +22,6 @@
       externals:
       - spec: git-lfs@2.11.0
         modules: [git-lfs/2.11.0]
-    python:
-      buildable: false
-      externals:
-      - spec: python@3.8.6%intel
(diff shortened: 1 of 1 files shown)


climbfuji (approved):
It's fine to set eckit to an older version for now, until the new Intel compilers are used on Acorn.

We cannot change the version of eckit in general, version 24 is a hard requirement for JEDI and will thus also be a hard requirement on Acorn/WCOSS2 in the future (or whatever latest version will be used by then). eckit, fckit and atlas are to some extend co-developed by JCSDA because JEDI needs certain features, therefore we pretty much have to stick with the latest release every time.

I assume that the eckit failures on Acorn with Intel 19 are due to versions 1.24+ now requiring C++-17 standard.

PR #796, opened 2023-09-29
Lmod tcl config

### LINKED ISSUE #703: Warning: detected custom TCL modules configuration
**Describe the bug**
The latest pull from main Spack gifted us with an annoying "Warning: detected custom TCL modules configuration..." message.

**To Reproduce**
Install any of the spack-stack environments

**Expected behavior**
No more warning!

**System:**
Affects all systems

**Additional context**
Related to #50

Per @climbfuji's suggestion in #672, let's split the common/modules.yaml configs into two files, have `spack stack create env` automatically use lmod unless it belongs to a site where tcl is needed (plus have manual override option), and add a CI check that ensures that modules_lmod.yaml and modules_tcl.yaml match to the extent that they should. Alternatively, we could put the config in one template file and have `spack stack create env` generate modules.yaml dynamically depending on whether lmod or tcl is used.

AlexanderRichert-NOAA:
### Summary

This PR separates common/modules.yaml into tcl and lmod versions. Based on the corresponding change in the create env command (https://github.com/JCSDA/spack/pull/329), it will automatically detect for a given site whether to use lmod or tcl, with an option to override.

### Testing

Tested on personal machine (& CI)

### Applications affected

n/a

### Systems affected

All

### Dependencies

https://github.com/JCSDA/spack/pull/329

### Issue(s) addressed

Fixes #703, #50

### Checklist


srherbener (approved):
Changes look good. Again didn't test (relying on CI testing).

climbfuji (commented):
I like this! Can you also update the documentation and the unit tests, please?

.gitmodules
```diff
@@ -2,9 +2,11 @@
...
   #branch = develop
-  url = https://github.com/jcsda/spack
-  branch = jcsda_emc_spack_stack
+  url = https://github.com/AlexanderRichert-NOAA/spack
+  branch = lmod_tcl_config
 [submodule "doc/CMakeModules"]
   path = doc/CMakeModules
   url = https://github.com/noaa-emc/cmakemodules
   branch = develop
+[submodule "spack/"]
```
climbfuji: Lines 11-12 need to be removed

climbfuji (changes requested):
I ran the following on my macOS:
```
spack stack create env --name=test-tcl --template=unified-dev --site=macos.default --modulesys=tcl
```
This worked although it should have failed, because now `envs/test-tcl/common/packages.yaml` uses `tcl`, but `envs/test-tcl/site/packages.yaml` uses `lmod`. This won't work. We need to think about what the viable options should be for this utiity.

I would also like to suggest that, after the code changes are finalized the documentation is updated in the different sections.

But overall neat, and thanks for adding the unit tests in the spack submodule.

AlexanderRichert-NOAA:
I'm not sure I follow? The idea is that tcl/lmod will be set automatically, but I added the '--modulesys' argument for users with weird use cases who want to do their own thing (with the expectation that they will need to do whatever they need to do to make their specific case work). That said, I'd be just as happy leaving that option out, setting it only automatically, and any user who wants something different can copy/modify config files as needed.

climbfuji:
> I'm not sure I follow? The idea is that tcl/lmod will be set automatically, but I added the '--modulesys' argument for users with weird use cases who want to do their own thing (with the expectation that they will need to do whatever they need to do to make their specific case work). That said, I'd be just as happy leaving that option out, setting it only automatically, and any user who wants something different can copy/modify config files as needed.

Do you think this can wait until Thursday? It's probably way more efficient to talk about this in the spack-stack meeting.

AlexanderRichert-NOAA:
@climbfuji having it update the site config turned out to not be too difficult, but obviously let me know if I missed something. There's also a CI check that does a diff on the modules_{tcl,lmod}.yaml files. It looks like somehow my documentation update didn't get pushed previously but it should be there now.

climbfuji (commented):
I've tested this on my macOS and it works great. Will approve once the submodule pointer is updated etc. Thanks very much for this!

climbfuji:
I merged the spack PR, new hash for spack is https://github.com/JCSDA/spack/commit/0669acaacf0a1f0cc9f7fca0294458bf50418951

AlexanderRichert-NOAA:
Updated submodule

climbfuji (approved):

PR #805, opened 2023-10-03
Update `spack stack create ctr`: make specs configurable, remove warnings, add CI builds

### LINKED ISSUE #146: Add CI test for container build
Now that we can build containers, we should add CI tests for the container builds.

climbfuji:
### Summary

Changes for how we create containers, and CI testing for them.
1. With https://github.com/JCSDA/spack/pull/333, the `spack stack create ctr` command is updated, it takes four options (`--container`, `--specs`, optional `--dir`, optional `--overwrite`). This way, we don't need to manually add specs to the container recipes files, and we also get rid of annoying warnings from invalid options in the container `spack.yaml`.
2. Updated the documentation with these changes.
3. Created a CI test for container builds, because this takes so long I decided to run them on a schedule (this is UTC time, there are already nightly scheduled github actions at 2 UTC and 4 UTC for other JCSDA projects), a different container build / spec combination each day (only one spec so far --> just rotate the containers). CI can also be triggered for PRs via labels (but no choice of container or spec) or via `workflow_dispatch` (can choose spec, container). Because of the no-choice limitation for containers/specs for PR builds, I am not sure how valuable those are. Maybe `workflow_dispatch` is good enough? Keep it simple.

### Testing

- [x] Manually built containers on the Ubuntu CI instance first
- [x] CI build (using `docker-ubuntu-gnu-openmpi` with `jedi-ci`)

### Applications affected

JEDI CI container builds

### Systems affected

None

### Dependencies

- waiting on https://github.com/JCSDA/spack/pull/333

### Issue(s) addressed

Resolves https://github.com/JCSDA/spack-stack/issues/146

### Checklist


configs/containers/docker-ubuntu-clang-mpich.yaml
```diff
@@ -110,7 +109,7 @@ spack:
       os: ubuntu:20.04
       spack:
         url: https://github.com/jcsda/spack
-        ref: spack-stack-1.5.0
+        ref: jcsda_emc_spack_stack
```
climbfuji, line 112: An actual bug fix, this should not have been updated when `spack-stack-1.5.0` was merged into develop.

ulmononian (approved):

srherbener (approved):

PR #880, opened 2023-11-17
Fix Python ssl variant issues for several systems

### LINKED ISSUE #854: To resolve after release/1.5.1: Python variant +ssl issues on systems with old openssl/openssh
**Describe the bug**
This is a slightly complicated problem (to describe and even more to solve).

On systems with older `libssh` and `libssl`, we need to turn of the `+ssl` variant for Python to avoid (a) build errors for certain packages (e.g. `fckit`) and concretization issues (duplicate packages). The problem seems to be that none of the packages in our stack explicitly depends on `libssh` and that not all software correctly defines its dependencies on it.

For example, building `fckit` with a newer version of `openssl` than installed in the OS leads to an error in the linker step, originating from `libssh`, that certain `openssl` symbols (from the older `libssl` installation in the OS) are not found.

Python 3.10 apparently requires newer `openssl` versions for enabling its `+ssl` variant, therefore we can't have that option turned on for certain systems.

The workaround for release/1.5.1 was to turn off the `ssl` variant in the respective site configs (see https://github.com/JCSDA/spack-stack/pull/853).

**To Reproduce**
See above

**Expected behavior**
Hidden dependency on `openssh` package (`libssh`) identified and added to package recipe, so that both `openssl` and `openssh` are installed by spack if necessary and the `ssl` variant for Python can be turned back on.

**System:**
Errors seen on Derecho, Gaea C5, Narwhal so far

**Additional context**
n/a

climbfuji:
### Summary

This PR fixes the Python `~ssl` variant issues on several systems described in https://github.com/JCSDA/spack-stack/issues/854. The solution is to remove (where applicable) external `curl`, `openssh`, `openssl`, Python variant `~ssl`.

### Testing

See https://github.com/JCSDA/spack-stack/issues/854 for a list of systems and the testing status.

### Applications affected

n/a

### Systems affected

See https://github.com/JCSDA/spack-stack/issues/854

### Dependencies

n/a

### Issue(s) addressed

Resolves https://github.com/JCSDA/spack-stack/issues/854

### Checklist


climbfuji:
@AlexanderRichert-NOAA I only have Nautilus left to test and fix (doesn't build curl for some weird reason). I updated the emc-rhel site config as part of this PR, but I don't have access to your GFE laptops. Would it be possible for you to test building this PR? See also https://github.com/JCSDA/spack-stack/issues/854. Thanks very much in advance!

climbfuji:
The remaining system Nautilus needs a fix for curl. I found an old issue in spack mainline on this and decided to fix it first there (see https://github.com/spack/spack/pull/41380).

AlexanderRichert-NOAA:
I tested installing fckit and it built fine. To clarify, is python now supposed to use +ssl? As long as that's expected then we're good to go.

climbfuji:
> I tested installing fckit and it built fine. To clarify, is python now supposed to use +ssl? As long as that's expected then we're good to go.

That seems to be enough! Thanks for testing. I am going to finish off Nautilus and then ask for reviews.

climbfuji:
@AlexanderRichert-NOAA The build on Nautilus finished with my temporary/manual fix to curl. I am still figuring out the correct patch with the spack developers in https://github.com/spack/spack/pull/41380, but that shouldn't hold up this PR (we can cherry-pick the PR once it's merged).

AlexanderRichert-NOAA (approved):

ulmononian (approved):

Issue #705, opened 2023-08-09
Reduce dependency on hand-built packages


AlexanderRichert-NOAA:
**Is your feature request related to a problem? Please describe.**
The goal of spack-stack is to avoid the bad old days of manually configured package installations. Across the platforms supported by spack-stack, there are non-spack installations of packages including python, ecflow, openmpi, and mysql, which are relied on by spack-stack packages and/or the applications we support. These outside packages add complexity and risk (e.g., https://github.com/ufs-community/ufs-weather-model/pull/1707#issuecomment-1670487960), and the goal of spack-stack is to have a single, self-consistently built software stack.

**Describe the solution you'd like**
Let's reduce the dependence on external packages as much as possible. MySQL is probably the trickiest of them, and I don't know the full extent of the story behind having separate python installations, but ecflow and openmpi for example should not be difficult to incorporate into the unified environment.

climbfuji:
I agree with the intention of this issue. But let me note that `ecflow` is a bit tricky when using Intel. That's because it uses a boost `cereal` function which for many many years is known to not work with Intel (and apparently nobody cares). This is the only reason we build ecflow as an external package for spack, using the same gcc compiler (and building a static boost library) that gets used as backend for the Intel compiler.

AlexanderRichert-NOAA:
Noted. I've had some success with building ecflow with intel under spack (though in the distant past I've run into the issue you describe). I'll test it out on various systems when I have time.

climbfuji:
> Noted. I've had some success with building ecflow with intel under spack (though in the distant past I've run into the issue you describe). I'll test it out on various systems when I have time.

Great news! Not sure we can make it part of this release, given the short time we have and all the testing it needs, but even if not then definitely for the next one. That would be hugely helpful to have going forward. Thanks for looking into this.

AlexanderRichert-NOAA:
Quick update-- I've reproduced what I believe is the same issue with cereal/boost. The problem occurs not in running ecflow_server, but in running any ecflow_client command to communicate with it. I'll dig a little more, but for now the answer might be either to only use boost%gcc, or to not build ecflow with intel, as it doesn't look like the issue will be resolved (in one place, it appeared that it came down to interpretations of code standards, so intel classic may not ever support the relevant feature).

eap:
Are there any updates on this issue; how should we move forward? It seems there's also tension here with the notion that we can use platform libraries to make builds more rapid and smaller, is there any consensus on the better approach?

[...]

PR #884, opened 2023-11-21
November 2023 main Spack merge + updates

AlexanderRichert-NOAA:
### Summary

_I'll keep the notes for this update, which includes https://github.com/JCSDA/spack/pull/371, in this PR._

This PR pulls in Spack release 0.21.0. Additionally, I've reconciled NCEPLIBS recipes with those in the library repos themselves (very few changes), as well as added in changes to support NCO's stack. Ran style checks/fixes on all packages. This PR also includes a proposed solution for the MySQL/Boost problem (#705).

Misc. notes:
 - wgrib2 now creates limitation for gmake version; this may be fixable with patching (modeled on wgrib2@3.1.1+ which is appropriately corrected)
 - MySQL now has an option to use internal boost libraries (this approach could probably be improved by creating resources)
 - I added a fix for a random gmake build issue
 - I tweaked a few default version settings, including removing a few; obviously these can be re-added as needed but so far so good based on my testing.
 - I tested out the new `spack deconcretize` command. Pretty nifty :)
 - Still need external flex to avoid duplicates.
 - Adding `rust_bootstrap` variants to py-cryptography and py-setuptools-rust
 - Looks like py-numpy and py-setuptools versions should stay pinned

**Waiting on:**
 - https://github.com/spack/spack/pull/41207

### Testing

 - Tested concretizing and some building on PC.
 - Tested building NCO's stack (needed some changes, incl. supporting deprecated packages).
 - Tested building UFS WM & GW deps on Hera; no duplicates

Tests to do:
- [x] Do another round of diffing Spack core (non-recipe) changes to inventory our changes (@AlexanderRichert-NOAA; there appears to be one unintentional difference in lib/spack/spack/build_environment.py that can get fixed whenever)
- [x] Hercules (@climbfuji)
- [x] Orion (@ulmononian)
- [x] Discover (@climbfuji)
- [x] Casper (@climbfuji)
- [x] Derecho (@climbfuji)
- [x] Acorn (@AlexanderRichert-NOAA)
- [x] Gaea C5 (@RatkoVasic-NOAA)
- [x] Hera (@RatkoVasic-NOAA)
- [x] Jet (@RatkoVasic-NOAA)
- [x] Narwhal (@climbfuji)
- [x] Nautilus (@climbfuji)
- [x] S4 (@climbfuji)
- [x] AWS AMI RH JCSDA (@climbfuji)
- [x] AWS ParallelCluster JCSDA (@climbfuji)
- [ ] RDHPCS Cloud (ParallelWorks) EPIC
- [x] RDHPCS Cloud (ParallelWorks) JCSDA (deferred to https://github.com/JCSDA-internal/jedi-tools/issues/325)
- [x]  JCSDA CI container builds (@climbfuji)
    - [x] docker-ubuntu-clang-mpich
    - [x] docker-ubuntu-gnu-openmpi
    - [x] docker-ubuntu-intel-impi
- [x] EMC RHEL  (@AlexanderRichert-NOAA)

New site configs:
- [x]  @climbfuji's macOS Monterey with apple-clang@13.1.6 and openmpi@4.1.6

### Applications affected

All.

### Systems affected

All.

### Dependencies

- [x] https://github.com/JCSDA/spack/pull/371

### Issue(s) addressed

Addresses #705 (w.r.t. MySQL)

### Checklist


AlexanderRichert-NOAA:
@climbfuji have you run into `Undefined symbols for architecture arm64:` / `ld: symbol(s) not found for architecture arm64` (see https://github.com/JCSDA/spack-stack/actions/runs/6951315212/job/18913085682?pr=884)? It's failing on gsl, but it's the same package/version/spec as the [current one](https://github.com/JCSDA/spack-stack/actions/runs/6906644848/job/18792132704).

climbfuji:
> Details

That's new to me. If I had to guess I'd say that the binary cache needs to be cleared. Let me do that and rerun the failing tests.

climbfuji:
Same error when building from scratch.

climbfuji:
@AlexanderRichert-NOAA I am trying to fix the macOS and Ubuntu CI errors. Given the problems we are seeing I would like to have a good testing coverage across the various HPCs that we support, ideally we build this branch on all of the tier-1 platforms before merging?

AlexanderRichert-NOAA:
Yep agreed. I'll fill out the test list in a bit.

climbfuji:
After fixing gsl on macOS (see https://github.com/JCSDA/spack/pull/371#issuecomment-1823554867), I am now hitting an error in py-scipy.

climbfuji:
and errors in prod-util, and many other packages are following ...

AlexanderRichert-NOAA:
Looks like the prod-util one is because of a patch where it's been fixed in the code, so that should be an easy one.

climbfuji:
> Looks like the prod-util one is because of a patch where it's been fixed in the code, so that should be an easy one.

Yep, will push later. Problems with py-scipy, I somewhat suspect that it's the reversal for the MACOSX_DEPLOYMENT_SYSTEM variable that I just cherry-picked from spack mainline. Sigh.

climbfuji:
I just pushed two commits, one for prod-util and one for scipy (a bug introduced by the merge it seems):
```
commit c1b6bd5415f6cad49b61768417b122883031b3bd (HEAD -> nov2023_spackmerge, alex/nov2023_spackmerge)
Author: Dom Heinzeller <dom.heinzeller@icloud.com>
Date:   Wed Nov 22 16:10:42 2023 -0700

    Fix merge bug fix in var/spack/repos/builtin/packages/py-scipy/package.py

commit cd2ec2855da52081050f8bf66e185dde6c351dbf
Author: Dom Heinzeller <dom.heinzeller@icloud.com>
Date:   Wed Nov 22 15:09:49 2023 -0700

    Remove invalid/unnecessary patch apple-clang-13.0.0-stdlib.patch for prod-util 2.1.0
```
Please take a look. I don't think this solves the scipy 1.11.3 problems, but it's definitely needed (I compared to spack dev)

AlexanderRichert-NOAA:
Thanks, both changes look good.

climbfuji:
@AlexanderRichert-NOAA I was able to build unified-env on my macOS with the various bug fixes. I cleared the build cache on all three CI systems (necessary due to the macOS MACOSX_DEPLOYMENT_TARGET bug fix) and pushed an empty commit to trigger CI.

I am going to stop here and take off for the long weekend. Let's pick this up on Monday. Happy Thanksgiving!

AlexanderRichert-NOAA:
Sounds great. Have a good Thanksgiving weekend!

AlexanderRichert-NOAA:
I'm working on rust, it looks like the issue is that the trick of using an older py-cryptography doesn't work because all of openssl@1 is deprecated now.

AlexanderRichert-NOAA:
@climbfuji would you be open to us modifying awscli-v2 to add support for using the [precompiled version](https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html) with vendored dependencies? Rust seems to be a pretty massive dependency just for this one little thing, but I also don't want to use a deprecated openssl indefinitely if we can help it. If yes, then I can work on that and keep going with the ubuntu CI if you can tackle the macOS CI (though right now it looks like rust is the issue for both).

climbfuji:
> @climbfuji would you be open to us modifying awscli-v2 to add support for using the [precompiled version](https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html) with vendored dependencies? Rust seems to be a pretty massive dependency just for this one little thing, but I also don't want to use a deprecated openssl indefinitely if we can help it. If yes, then I can work on that and keep going with the ubuntu CI if you can tackle the macOS CI (though right now it looks like rust is the issue for both).

I am fine with that approach. Not having to build rust sounds like a good idea, and awscli is just a simple package from which we use binaries, not link against it. FWIW, rust built on my mac, but let's try to avoid it if we can.

AlexanderRichert-NOAA:
@climbfuji FYI the macos-ci-x86_64-build timed out previously after 6 hours. Looking at the current run, it seems that some of the stage times are particularly long, so I wonder if the runner is having internet issues? In any case do you mind keeping an eye on that?

climbfuji:
> @climbfuji FYI the macos-ci-x86_64-build timed out previously after 6 hours. Looking at the current run, it seems that some of the stage times are particularly long, so I wonder if the runner is having internet issues? In any case do you mind keeping an eye on that?

I'll need to hop onto the machine and do one build out of CI to create a new binary cache. I'll check if there are issues with getting data, or the disk running full.

climbfuji:
You got all the CI tests to pass! I think we should get #880 merged before we start testing on the HPCs.

climbfuji:
As for #880, this still needs a fix for Derecho (yep, once again) and a test on Nautilus (is under maintenance until today).

climbfuji:
@AlexanderRichert-NOAA I updated your spack branch from jcsda_emc_spack_stack, updated your spack-stack branch from develop, and committed the submodule pointer update. I think we can start testing on the HPCs?

configs/common/packages.yaml
```diff
@@ -82,7 +82,7 @@
       #version: ['2023.01']
       #variants: precision=32,64 +quad_precision +gfs_phys +openmp +pic constants=GFS build_type=Release
       version: ['2023.02.01']
```
climbfuji, line 85:
```suggestion
      version: ['2023.03']
```

climbfuji: Should we also delete the two commented lines directly above?

AlexanderRichert-NOAA:
:cowboy_hat_face: Yep sounds good

AlexanderRichert-NOAA:
@climbfuji it occurs to me that base-env is still specifying `zlib` as a dependency, but every zlib-using package is using `zlib-api` where `zlib-ng` is always being selected as the provider. This could lead to inconsistencies in downstream applications, and we probably don't need both. Would it make sense to just go all in with zlib or zlib-ng (either way, we should add it to the list of providers in common/packages.yaml)? Maybe it makes the most sense to use zlib for now, since that's what we currently use, then test upgrading to zlib-ng later. If you agree I'll update the config.

climbfuji:
> @climbfuji it occurs to me that base-env is still specifying `zlib` as a dependency, but every zlib-using package is using `zlib-api` where `zlib-ng` is always being selected as the provider. This could lead to inconsistencies in downstream applications, and we probably don't need both. Would it make sense to just go all in with zlib or zlib-ng (either way, we should add it to the list of providers in common/packages.yaml)? Maybe it makes the most sense to use zlib for now, since that's what we currently use, then test upgrading to zlib-ng later. If you agree I'll update the config.

That sounds like a good plan.

I ran into problems with the container bulds because it says that ecmwf-atlas doesn't have a variant `ectrans` anymore (it's called `trans` in spack develop):
```
> git diff jcsda/jcsda_emc_spack_stack var/spack/repos/builtin/packages/ecmwf-atlas
diff --git a/var/spack/repos/builtin/packages/ecmwf-atlas/package.py b/var/spack/repos/builtin/packages/ecmwf-atlas/package.py
index d720ed1ff0..44cdd7adc2 100644
--- a/var/spack/repos/builtin/packages/ecmwf-atlas/package.py
+++ b/var/spack/repos/builtin/packages/ecmwf-atlas/package.py
@@ -52,9 +52,8 @@ class EcmwfAtlas(CMakePackage):
     variant("openmp", default=True, description="Use OpenMP")
     depends_on("llvm-openmp", when="+openmp %apple-clang", type=("build", "run"))
     variant("shared", default=True, description="Build shared libraries")
-
-    variant("ectrans", default=False, description="Enable ectrans", when="@0.31.0:")
-    depends_on("ectrans@1.1.0:", when="+ectrans")
+    variant("trans", default=False, description="Enable trans")
+    depends_on("ectrans@1.1.0:", when="@0.31.0: +trans")
     variant("eigen", default=True, description="Enable eigen")
     depends_on("eigen", when="+eigen")
     variant("fftw", default=True, description="Enable fftw")
@@ -73,9 +72,9 @@ def cmake_args(self):
             "-DPYTHON_EXECUTABLE:FILEPATH=" + self.spec["python"].command.path,
         ]
         if self.spec.satisfies("@0.31:0.34"):
-            args.append(self.define_from_variant("ENABLE_TRANS", "ectrans"))
+            args.append(self.define_from_variant("ENABLE_TRANS", "trans"))
         if self.spec.satisfies("@0.35:"):
-            args.append(self.define_from_variant("ENABLE_ECTRANS", "ectrans"))
+            args.append(self.define_from_variant("ENABLE_ECTRANS", "trans"))
             args.append(self.define_from_variant("ENABLE_TESSELATION", "tesselation"))
         if "~shared" in self.spec:
             args.append("-DBUILD_SHARED_LIBS=OFF")
```
We'll need to update the following two files with the variants (no need to re-test building on the systems we already tested, ectrans built there fine every time). Do you want me to push these updates?
```
> git diff configs/
diff --git a/configs/common/packages.yaml b/configs/common/packages.yaml
index df4d820..a98ba2b 100644
--- a/configs/common/packages.yaml
+++ b/configs/common/packages.yaml
@@ -52,7 +52,7 @@
       variants: linalg=eigen,lapack compression=lz4,bzip2
     ecmwf-atlas:
       version: ['0.35.0']
-      variants: +fckit +ectrans +tesselation +fftw
+      variants: +fckit +trans +tesselation +fftw
     ectrans:
       version: ['1.2.0']
       variants: ~mkl +fftw
diff --git a/configs/containers/specs/jedi-ci.yaml b/configs/containers/specs/jedi-ci.yaml
index 3cc655d..eb53ea3 100644
--- a/configs/containers/specs/jedi-ci.yaml
+++ b/configs/containers/specs/jedi-ci.yaml
@@ -1,7 +1,7 @@
   ### spack-stack-1.5.1 / skylab-6.1.0 containers for fv3-jedi and mpas-jedi (but not for ufs-jedi)
   specs: [base-env@1.0.0, jedi-base-env@1.0.0, ewok-env@1.0.0, jedi-fv3-env@1.0.0,
     jedi-mpas-env@1.0.0, bacio@2.4.1, bison@3.8.2, bufr@12.0.1, ecbuild@3.7.2, eccodes@2.27.0, ecflow@5,
-    eckit@1.24.4, ecmwf-atlas@0.35.0 +fckit +ectrans +tesselation +fftw, fiat@1.2.0, ectrans@1.2.0 +fftw,
+    eckit@1.24.4, ecmwf-atlas@0.35.0 +fckit +trans +tesselation +fftw, fiat@1.2.0, ectrans@1.2.0 +fftw,
     eigen@3.4.0, fckit@0.11.0, fms@release-jcsda, g2@3.4.5, g2tmpl@1.10.2, gftl-shared@1.6.1,
     gsibec@1.1.3, hdf@4.2.15, hdf5@1.14.0, ip@4.3.0, jasper@2.0.32, jedi-cmake@1.4.0,
     libpng@1.6.37, nccmp@1.9.0.1, netcdf-c@4.9.2, netcdf-cxx4@4.3.1,
```

AlexanderRichert-NOAA:
Yes please go ahead with that, sorry I missed those. I'll add the zlib update momentarily (worked fine on Acorn just now).

climbfuji:
> Yes please go ahead with that, sorry I missed those. I'll add the zlib update momentarily (worked fine on Acorn just now).

Done, thanks for the quick reply. I am surprised that the `configs/common/packages.yaml` variant `ectrans` didn't lead to an error, but probably because these are just "soft requirements". The explicit request for the variant in the container spec did trigger the error.

climbfuji:
The container build also doesn't like the precision variant for `sp`:
```
   echo '    sp:' &&   echo '      version: [2.3.3]' &&   echo '      variants: precision=4,d,8'
```
-->
```
1. Cannot set variant 'precision' for package 'sp' because the variant condition cannot be satisfied for the given spec. You could consider setting `concretizer:unify` to `when_possible` or `false` to allow multiple versions of some packages.
```
That's strange ... do we need some extra quotes?

climbfuji:
I think we need to update the version of `sp` in configs/common/packages.yaml for the precision variant to be valid. 2.5.0 ?

AlexanderRichert-NOAA:
Yep that sounds good.

climbfuji:
> Yep that sounds good.

Done

climbfuji:
Error on Narwhal on the home stretch, for mapl with GNU (worked with Intel):
```
work1/heinzell/spack-stack-nov2023_spackmerge/cache/build_stage/spack-stage-mapl-2.40.3-bpfups3oqez3cqbg5wqu4zn6ofoxalty/spack-src'

1 error found in build log:
     22    -- Cray Programming Environment 2.7.19 C
     23    -- Detecting C compiler ABI info
     24    -- Detecting C compiler ABI info - done
     25    -- Check for working C compiler: /p/work1/heinzell/spack-stack-nov2023_spackmerge/spack/lib/spack/env/gcc/gcc - skipped
     26    -- Detecting C compile features
     27    -- Detecting C compile features - done
  >> 28    CMake Error at ESMA_cmake/esma.cmake:5 (esma_check_install_prefix):
     29      Unknown CMake command "esma_check_install_prefix".
     30    Call Stack (most recent call first):
     31      CMakeLists.txt:29 (include)
     32
     33
     34    -- Configuring incomplete, errors occurred!

See build log for details:
  /p/work1/heinzell/spack-stack-nov2023_spackmerge/cache/build_stage/spack-stage-mapl-2.40.3-bpfups3oqez3cqbg5wqu4zn6ofoxalty/spack-build-out.txt
```
I can look into it later but if @mathomp4 has an idea right away (hopefully!) ...

AlexanderRichert-NOAA:
Noted. That looks suspiciously like an error I ran into ages ago involving the esma_cmake resource and the caching thereof.

mathomp4:
That macro is fairly benign:

https://github.com/GEOS-ESM/ESMA_cmake/blob/main/esma_support/esma_check_install_prefix.cmake

it just checks to make sure you aren't installing to `/usr/local`, you don't have a comma in the directory path[^1], and where you are installing, you have write permissions. 

[^1]: That one was a fun bug to figure out.

climbfuji:
> Noted. That looks suspiciously like an error I ran into ages ago involving the esma_cmake resource and the caching thereof.

Hmm, but given that it installed fine a few hours beforehand with Intel that seems rather unlikely. And there's no difference between jcsda_emc_spack_stack mapl and this PR, and it previously built just fine. Grrrrrr...

AlexanderRichert-NOAA:
Try relocating the cached version so it redownloads? But don't delete it, in case I'm wrong :)

climbfuji:
> Try relocating the cached version so it redownloads? But don't delete it, in case I'm wrong :)

Solution, a lot easier: run `spack clean -a` once, try, it fails. Wait a few hours, run `spack clean -a` again, try, it installs just fine.

configs/common/packages.yaml
climbfuji: Note that we don't have to rerun tests for this, currently `unified-dev` contains a whole list of fms versions, including `2023.02.01` and `2023.03`.

climbfuji (commented):
I am happy with all of this, and very thankful to @AlexanderRichert-NOAA for shepherding these PRs through. I will approve once the spack PR is merged and the submodule updated as usual.

climbfuji:
spack PR was merged, new hash is https://github.com/JCSDA/spack/commit/8683cc4f1a42329cab38d2fd607476e202ce50d1

configs/common/packages.yaml
climbfuji: We can make these changes in one of the next PRs when we update to 2023.04.

configs/common/packages.yaml
```diff
@@ -152,6 +156,8 @@
       version: ['5.1.0']
     mpich:
       variants: ~hwloc +two_level_namespace
+    mysql:
+      variants: +download_boost
```
climbfuji, line 160: I am not sure how relevant this is at the moment, given that we install mysql as external dependency on all systems, but it won't hurt either.

configs/common/packages.yaml
```diff
@@ -237,8 +241,7 @@
       version: ['1.5.8']
       variants: ~mpi
     py-numpy:
-      version: ['1.22.3']
-      variants: +blas +lapack
```
climbfuji, line 241: @AlexanderRichert-NOAA Do you know what the impact of removing line 241 is?

climbfuji (commented):
I went ahead and updated the `spack` submodule / reverted `.gitmodules`. Only one question left from my side, other than that it's good to go!

natalie-perlin:
@AlexanderRichert-NOAA @climbfuji - 
let me update or include README.md files for noaa-<cloud> platforms

climbfuji:
> @AlexanderRichert-NOAA @climbfuji - let me update or include README.md files for noaa- platforms

@natalie-perlin How quickly can you do that? If you need a bit longer, we can make it a small follow-up PR (since no testing needed for adding/updating README.md files).

configs/common/packages.yaml
AlexanderRichert-NOAA: Those variants don't exist anymore, the recipe now always builds with blas and lapack.

climbfuji: Perfect, thanks.

climbfuji (approved):

climbfuji:
Merging this now to avoid container build problems (since the spack PR was merged this morning).

PR #891, opened 2023-12-05
Add utility for single-node parallel installations

AlexanderRichert-NOAA:
### Summary

This PR adds a utility for doing single-node parallel installations using backgrounded instances of `spack install`. Mostly suitable for small installations or interactive batch jobs; for larger installations using a batch job along the lines of what we have for Acorn probably makes more sense.

### Testing

Tested on personal machine.

### Applications affected

none

### Systems affected

none

### Dependencies

none

### Issue(s) addressed

none

### Checklist

### DIFF
--- a/util/parallel_install.sh
+++ b/util/parallel_install.sh
@@ -1,0 +1,46 @@
+#!/usr/bin/bash
+
+# This script launches backgrounded instances of the 'spack install' command
+# for an active Spack environment ($SPACK_ENV must be set). The first two
+# arguments are required, specifying the number of instances and number of
+# threads per instance, respectively. Remaining arguments are passed to the
+# 'spack install' command. It is strongly recommended to only run this utility
+# in a concretized environment. The script can be sourced so that the
+# backgrounded jobs are associated with the current shell environment (strongly
+# recommended when running inside a 'screen' or 'tmux' session).
+#
+# NOTE: If you are installing on a shared system (i.e., an HPC platform),
+# especially if not running through a job scheduler with specified resources,
+# be respectful of system resources and usage policies.
+#
+# *Depending on available resources*, for building the entire spack-stack
+# unified environment, 2-4 instances with 4-8 cores each is generally
+# reasonable. Six or more instances is generally overkill, as is more than 8
+# cores per package build.
+#
+# Example usage:
+#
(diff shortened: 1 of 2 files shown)


climbfuji (commented):
What is the return code from this script? Is it /= 0 if one or more of the parallel install commands fail?

AlexanderRichert-NOAA:
Based on the `wait` command, the return code is the most recent one, so we probably want something more robust than that. I'll play it around with it some time. This definitely doesn't need to be there for 1.6.0, just a nice to have.

srherbener (approved):
Thanks for adding this utility. It looks quite useful!

util/parallel_install.sh
```diff
@@ -0,0 +1,55 @@
...
+
+argtext="Run as '$0 N M', where N=# of instances, M=# of threads per instance"
+n_instances=${1?"$argtext"}
+n_threads=${2?"$argtext"}
+shift 2
+
+if [[ ! " $(seq -s ' ' 10) " =~ " $n_instances " ]]; then echo "Invalid number of instances" ; exit 1; fi
+if [[ ! " $(seq -s ' ' 64) " =~ " $n_threads " ]]; then echo "Invalid number of threads" ; exit 1; fi
+
+echo "Installing with $n_instances instances and ${n_threads} threads in environment '${SPACK_ENV?"SPACK_ENV not set!"}'"
```
climbfuji: Do you want to abort if SPACK_ENV is not set, i.e. make it a requirement?

climbfuji:
Also add to doc/source/?

util/parallel_install.sh
AlexanderRichert-NOAA: I think that's what this will do? If not then I can explicitly set `set -a` or whatever to fail on non-zero.

climbfuji: `set -e` I think - I was confused because the notes above say "it is strongly recommended" to do this in an active env.

climbfuji (approved):
Thanks, looks good to me!

ulmononian (approved):

climbfuji:
I resolved the conflict, once the readthedocs page is rendered I'll take a look, if it looks good I'll merge

PR #896, opened 2023-12-07
Adding README.md to all noaa-* cloud site directories; modify noaa-azure/packages.yaml

natalie-perlin:
### Summary

- Adds identical README.md files to all ./configs/sites/noaa-* directories.
- Modifies a packages.yaml file in ./configs/sites/noaa-azure/packages.yaml to add ncurses@5.9-14.20130511 that is pre-installed when running /contrib/admin/basic-setup.sh. 

### Testing

These changes for GCP/AWS/Azure were tested when installing spack-stack-1.5.0, 1.5.1.

### Systems affected

NOAA CSP (noaa-aws, noaa-gcp, noaa-azure)

### Issue(s) addressed

Resolves issues with building a required package ncurses on NOAA Azure. Errors 

### Checklist


configs/sites/noaa-aws/README.md
```diff
@@ -124,3 +133,19 @@ make install 2>&1 | tee log.install
 cd /contrib/spack-stack/modulefiles
 mkdir ecflow
 # Create the modulefile from the template in doc/modulefile_templates
+
+############## Steps to perform when starting a new cluster ##############
```
climbfuji, line 128: Can lines 137 to 151 go elsewhere in EPIC land? The site config doesn't distinguish between JCSDA and EPIC or any other PW project. We (JCSDA) don't have `/contrib/EPIC/bin` and we don't need it. You could put this in another admin script like `/contrib/admin/epic_setup.sh` and call both through the PW cluster configuration (web page).

configs/sites/noaa-azure/README.new
```diff
@@ -0,0 +1,151 @@
+This README provides step by step instructions for installing the basic packages
```
climbfuji: Why is this file called README.new - the others are called `README.md`?

configs/sites/noaa-aws/README.md
climbfuji: Or maybe have a `configs/sites/noaa-common/` folder with a README.md, a README_epic_addon.md and a README_jcsda_addon.md?

climbfuji (commented):
Replicating these README.md files means making identical changes three times. Is that what we want? We should also think about how we deal with JCSDA and EPIC specific items (see my other comments).

configs/sites/noaa-azure/README.new
natalie-perlin: It was definitely a typo. Will correct.

configs/sites/noaa-aws/README.md
natalie-perlin: Thank you for your comments and suggestions!
I could then suggest making README.md files cloud platform-specific (ncurses to be pre-installed for Azure only).

The path for binaries /contrib/EPIC/bin could be taken out in general, if not used by anybody else in PW.  It contains aws-cli, among other packages, which is useful for installing from mirrors and pulling data from the AWS S3.

Could other directions, such as shown below, still be useful for other PW users, who may want to use build spack-stack or a new environment, or for building/running an application based on the installed spack-stack?

```
module unuse /opt/cray/craype/default/modulefiles
module unuse /opt/cray/modulefiles
module use /contrib/spack-stack/modulefiles
module load cmake/3.27.2  # needed for building spack-stack
module load ecflow/5.8.4
module load mysql/8.0.31
module load git-lfs/2.4.1
```

climbfuji: We typically advise users to type this themselves, thus it would be inconsistent with the other platforms, bit I think it's ok to keep
```
module unuse /opt/cray/craype/default/modulefiles
module unuse /opt/cray/modulefiles
module use /contrib/spack-stack/modulefiles
module load cmake/3.27.2  # needed for building spack-stack
module load ecflow/5.8.4
module load mysql/8.0.31
module load git-lfs/2.4.1
```
Note that awscli-v2 is part of spack-stack, therefore no need to install separately. Just load the module and use as is.

natalie-perlin: Thank you!  Oh, the aws-cli was needed to install a new spack stack from a mirror and for pulling data from S3, and so it was before any module was built.

climbfuji (approved):

PR #940, opened 2024-01-10
Fix CI tests on R&D ParallelCluster

### LINKED ISSUE #939: Nightly builds on JCSDA R&D AWS ParallelCluster with Intel broken
**Describe the bug**
Seems that since last week when the `metis` PR was merged, the nightly builds on the JCSDA R&D cluster (AWS ParallelCluster) are broken: https://github.com/JCSDA/spack-stack/actions/runs/7454696094/job/20282497526 - duplicate packages in the concretization step.

**To Reproduce**
See https://github.com/JCSDA/spack-stack/actions/runs/7454696094/job/20282497526

**Expected behavior**
No duplicate packages, build completes

**System:**
AWS ParallelCluster used as JCSDA R&D cluster with Intel compiler

**Additional context**
n/a

climbfuji:
### Summary

This is more of a workaround than a good solution. By looping over the two compilers, reconcretizing and installing, I can get around the concretizer problems that are described in https://github.com/JCSDA/spack-stack/issues/939.

I tested this manually on AWS ParallelCluster. Since it runs via `workflow_dispatch`, I can't run the CI tests from my fork/branch. We'll have to merge and check afterwards if it works.

### Testing

See above

### Applications affected

None

### Systems affected

AWS ParallelCluster CI runs only

### Dependencies

none

### Issue(s) addressed

Resolves https://github.com/JCSDA/spack-stack/issues/939

### Checklist

### DIFF
--- a/.github/workflows/ubuntu-rnd-x86_64.yaml
+++ b/.github/workflows/ubuntu-rnd-x86_64.yaml
@@ -83,7 +83,7 @@ jobs:
             cat spack.yaml
 
             # Concretize and check for duplicates
-            spack concretize 2>&1 | tee log.concretize.${ENVNAME}.${compiler}
+            spack concretize --force --fresh 2>&1 | tee log.concretize.${ENVNAME}.${compiler}
             ${SPACK_STACK_DIR}/util/show_duplicate_packages.py -d log.concretize.${ENVNAME}.${compiler} -i fms -i crtm -i esmf -i mapl
 
             # Update spack source cache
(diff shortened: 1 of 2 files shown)


srherbener (approved):
I think the code looks good. It seems to be capturing what should have been run before this update. I guess we will see with the testing. Thanks!

AlexanderRichert-NOAA (approved):
If it runs and works then this is fine. Would it make sense to use `strategy:matrix:...` to avoid all the indentation, and also allow gcc and intel to run in parallel? I have no problem with the current approach though.

climbfuji:
> If it runs and works then this is fine. Would it make sense to use `strategy:matrix:...` to avoid all the indentation, and also allow gcc and intel to run in parallel? I have no problem with the current approach though.

That's a good idea. I'll make this change in a follow-up PR. That will also provide a chance to fix issues that we can only find out by merging this PR.

PR #958, opened 2024-01-18
Acorn config update

AlexanderRichert-NOAA:
### Summary
This PR adds gcc 10.2.0 to the intel compiler build environment on Acorn (as we do for other systems), and also pins several packages versions for intel 19 on Acorn to avoid C++-related build failures. I have created separate intel 19 and 2022 environment on Acorn under spack-stack-1.6.0 (due to an unrelated issue having to do with py-contourpy), and those directories are now reflected in the documentation.

### Testing
Tested on Acorn; these modifications are reflected in the directories shown in the updated documentation.

### Applications affected
all

### Systems affected
Acorn only

### Dependencies
none

### Issue(s) addressed
--

### Checklist


climbfuji (approved):
Makes sense to me, thanks for the updates!

climbfuji:
@Hang-Lei-NOAA Can you take a look at this PR please? Thanks!

Hang-Lei-NOAA:
@Alexander Richert - NOAA Affiliate ***@***.***> I
remembered that the gcc/11.2.0 is on acorn, why not pin gcc/11.2.0?
This commit itself should be fine to add gcc in the environment.

AlexanderRichert-NOAA:
For some reason not all of the GCC installations work in this context, 10.2.0 is the only one for each where I was able to get it to build the whole stack.

Hang-Lei-NOAA (approved):

PR #978, opened 2024-01-31
Add utility for verifying that version/variant/externals are respected

AlexanderRichert-NOAA:
### Summary

This PR adds a utility for verifying the concretized environments respect the configured version (common/packages.yaml), variant (common/packages.yaml), and external (site/packages.yaml) settings.

### Testing

Tested on personal machine and in util_tests.sh/CI. Will explore adding to build CI.

### Applications affected

none

### Systems affected

Should be useful on all systems, though as currently written, it might involve ignoring certain packages on certain systems.

### Dependencies

none

### Issue(s) addressed

### Checklist


AlexanderRichert-NOAA:
@climbfuji I decided not to put it in the main CI since I think the utility of it is more for telling an installer "hey just so you know," but I'm open to adding it. I think it would take a bit of work since when I tried it there were maybe a dozen or so packages that didn't match the configuration (but I don't know how much of that is intentional).

climbfuji (commented):
This is a handy tool, thanks for coming up with this. Two questions:

1. How are variant mismatches for externals handled? I believe spack concretize just uses the variants specified for the external when it uses it, correct? (same for version)
2. Would it be better to use a repetitive `-i` (ignore) flag for ignoring certain packages like `esmf`? This way it would be more consistent with the `show_duplicate_packages.py` script, and it would also be more intuitive. The current syntax is more commonly used to check for just this package, not all others except that?

AlexanderRichert-NOAA:
1. Correct (i.e., we don't need a separate check because the concretized spec will be constrained by the external spec).
2. Done. The ubuntu ci is failing with some kind of git/runner-related glitch but otherwise I think the updated code works.

climbfuji (approved):
Thanks for considering my comments and suggestions!

I restarted the failed CI run and enabled auto-merge

AlexanderRichert-NOAA:
@climbfuji cool. FYI I (finally) added the util-tests as required just now.

PR #979, opened 2024-01-31
No szip for WCOSS2 hdf5 & netcdf

AlexanderRichert-NOAA:
No szip for WCOSS2 hdf5 & netcdf.

### DIFF
--- a/configs/common/packages.yaml
+++ b/configs/common/packages.yaml
@@ -114,7 +114,7 @@
       variants: +external-xdr ~fortran ~netcdf
     hdf5:
       version: ['1.14.0']
-      variants: +hl +fortran +mpi ~threadsafe +szip
+      variants: +hl +fortran +mpi ~threadsafe ~szip
     ip:
       version: ['4.3.0']
       variants: precision=4,d,8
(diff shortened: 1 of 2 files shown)


climbfuji:
We don't need `szip` anywhere - can you instead remove it entirely (common / packages.yaml)?

AlexanderRichert-NOAA:
@edwardhartnett @junwang-noaa is it correct to say we don't need szip support for hdf5/netcdf anywhere? I think we discussed this some time ago but I don't remember what the conclusion was.

edwardhartnett:
Correct.

Both HDF5 and netCDF will build with or without szip (a.k.a. szlib). So unless we have some group that is demanding to use szip, or we have some szip-packed netCDF data we need to work with, we don't have to build it.

Both zlib and especially zstandard are better choices (faster, more compressive), so there's no reason for anyone to use this archaic compression.

climbfuji:
Thanks Ed, that was my understanding as well.

climbfuji:
Looks good! I'll take a look at the CI tests to check that it indeed built hdf5 and netcdf-c w/o szip support.

climbfuji (approved):

PR #1031, opened 2024-03-18
Document workaround for apple-clang@15.0.0 linker/loader issues.

### LINKED ISSUE #971: Implement workaround for apple-clang@15.x.y linker/loader issue

Apple has an issue with the linker/loader (ld) that is shipped with apple-clang%15.x compilers of which can be worked around by requesting the older linker/loader functionality via the `-Wl,-ld_classic` compiler option. Apple is aware of this issue and has an internal ticket: FB13208302: "Building ffmpeg 6.0 with shared libraries broken with Xcode 15's default linker"

For some examples, see:
  https://github.com/spack/spack/pull/40187 (ffmpeg build)
  https://github.com/spack/spack/pull/42264 (hdf5 build)

This workaround also fixes the issue with apple-clang%15.x reported in #829 which reported errors during the configure step with the openmpi build.

There seems to be number of packages impacted by the apple-clang%15.x linker/loader issue: hdf5, openmpi, ffmpeg so far.

Option 2 has the advantage that we are only fixing known problems, and if it turns out to be a handful of packages it's really not that difficult to remove later on.

I thing there are two options:
1. Globally set `LDFLAGS="-Wl,-ld_classic"` when compiling with apple-clang%15.x to address anticipated issues
2. Set `LDFLAGS="-Wl,-ld_classic"` on a case-by-case basis as issues are found

Option 1 has the advantage that it is easy to remove the workaround once apple has fixed the issue. But has the disadvantage that we are changing build process for a lot of packages that are not broken.

Option 2 has the advantage that we are only fixing the packages that are known to be broken, and if there are not too many of those it will be easy to remove the work around when apple fixes their linker/loader issue.

I'm leaning toward option 2 now. I'm a little bit concerned about the global setting in option 1.

What does everyone think?

**Additional context**

srherbener:
### Summary

Until Apple fixes their new linker/loader associated with apple-clang@15.0.0 compilers, we need to use a manual workaround to get the spack-stack build to complete on MacOS platforms. See #971 for details. This PR adds documentation showing the user how to manually insert the workaround.

### Testing

I tested this on my Mac using Sphinx to build the modified documentation. It appears to be rendering the new instructions properly, but we'll see for sure with the CI testing.

### Applications affected

This primarily impacts UFS and JEDI applications, however the impact is very limited. This is documentation only change for a workaround that is only necessary on apple-clang%15.0.0 or newer compilers.

### Systems affected

MacOS platforms when using apple-clang@15.0.0 or newer compilers

### Dependencies

None

### Issue(s) addressed

Resolves #971

### Checklist

### DIFF
--- a/doc/source/NewSiteConfigs.rst
+++ b/doc/source/NewSiteConfigs.rst
@@ -268,7 +268,7 @@ Remember to activate the ``lua`` module environment and have MacTeX in your sear
       extra_rpaths: []
 
 .. note::
-  Apple is aware of this issue and working on a solution, so this is a temparary workaround that will be removed once the linker/loader issues are repaired.
+  Apple is aware of this issue (Apple ticket number FB13208302) and working on a solution, so this is a temporary workaround that will be removed once the linker/loader issues are repaired.
 
 6. Do **not** forget to unset the ``SPACK_SYSTEM_CONFIG_PATH`` environment variable!
 
--- a/doc/source/NewSiteConfigs.rst
+++ b/doc/source/NewSiteConfigs.rst
@@ -244,6 +244,32 @@ Remember to activate the ``lua`` module environment and have MacTeX in your sear
 
    spack compiler find --scope system
 
+.. note::
+  When using apple-clang@15.0.0 (or newer) compilers, you need to manually add the following ldflags spec in the `site/compilers.yaml` file.
+  There are known issues with new features in the Apple linker/loader that comes with the 15.0.0 compiler set, and this change tells the linker/loader to use its legacy features which work fine.
+
+.. code-block:: yaml
+  :emphasize-lines: 9,10
+
+  compilers:
+  - compiler:
+      spec: apple-clang@=15.0.0
+      paths:
+        cc: /usr/bin/clang
+        cxx: /usr/bin/clang++
+        f77: /opt/homebrew/bin/gfortran-12
+        fc: /opt/homebrew/bin/gfortran-12
+      flags:
+        ldflags: '-Wl,-ld_classic'         # Add this ldflags spec
+      operating_system: sonoma
+      target: aarch64
+      modules: []
+      environment: {}
+      extra_rpaths: []
+
+.. note::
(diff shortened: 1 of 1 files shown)


srherbener:
I could not figure out a way to get a `spack config ...` command to add the ldflags setting that the instructions in this change direct the user to do. If anyone has a suggestion of how to do this with a spack command, that would be very much appreciated. Thanks!

doc/source/NewSiteConfigs.rst
```diff
@@ -521,6 +547,7 @@ It is recommended to increase the stacksize limit by using ``ulimit -S -s unlimi
    If the environment will be used to run JCSDA's JEDI-Skylab experiments using R2D2 with a local MySQL server, run the following command:
 
 .. code-block:: console
+
```
srherbener, line 550: This is unrelated to the issue being addressed, but the sphinx-build process I was using to test the rendering threw an error in this code-block without the blank line.

doc/source/NewSiteConfigs.rst
```diff
@@ -244,6 +244,32 @@ Remember to activate the ``lua`` module environment and have MacTeX in your sear
...
+      flags:
+        ldflags: '-Wl,-ld_classic'         # Add this ldflags spec
+      operating_system: sonoma
+      target: aarch64
+      modules: []
+      environment: {}
+      extra_rpaths: []
+
+.. note::
+  Apple is aware of this issue and working on a solution, so this is a temparary workaround that will be removed once the linker/loader issues are repaired.
```
climbfuji:
```suggestion
  Apple is aware of this issue and working on a solution, so this is a temporary workaround that will be removed once the linker/loader issues are repaired.
```
Can you add a sentence here mentioning the Apple ticket, please?

srherbener: Thanks for the quick review! I added the apple ticket number and fixed the indicated typo.

climbfuji (approved):

PR #1077, opened 2024-04-16
Update path to find qt@5

svahl991:
### Summary

The homebrew path to qt version 5 that is currently in the instructions was not found on a new system, resulting in qt not being found by the `find` command. Updating the path.

### Testing

Instructions tested on new M2 Mac with fresh homebrew install.


climbfuji (approved):
Thanks!

climbfuji:
Actually maybe it's best to add a note before that updated `spack external find` command to instruct users to check if the path exists?

svahl991:
> Actually maybe it's best to add a note before that updated `spack external find` command to instruct users to check if the path exists?

Done.

climbfuji (approved):

Issue #1130, opened 2024-06-03
[INSTALL]: spack-stack on Gaea C6


BijuThomas-NOAA:
### Package name

spack-stack

### Package version/tag

1.6.0

### Build options

current

### Installation timeframe

HAFS team would like to install UFS weather model and HAFS on Gaea C6 sooner.

BijuThomas-NOAA:
We request to install spack-stack libraries on Gaea C6 for us to build and run the HAFS system

AlexanderRichert-NOAA:
I'll work on a C6 config based on 1.6.0 then work on adding to develop for 1.8.0.

AlexanderRichert-NOAA:
@BijuThomas-NOAA can you test with /autofs/ncrc-svm1_proj/epic/spack-stack/spack-stack-1.6.0/envs/unified-env-c6/install/modulefiles/Core and let me know if there are any issues?

BijuThomas-NOAA:
@AlexanderRichert-NOAA Thank you. I will test HAFS system on Gaea and let you know.

BijuThomas-NOAA:
@AlexanderRichert-NOAA I see the following error: 
```
module purge
module load /autofs/ncrc-svm1_proj/epic/spack-stack/spack-stack-1.6.0/envs/unified-env-c6/install/modulefiles/Core
module load PrgEnv-intel/8.3.3
module load stack-intel/2023.2.0
```
_Lmod has detected the following error:  The following module(s) are unknown: "PrgEnv-intel/8.5.0" "intel-classic/2023.2.0" "craype/2.7.30"_

Am I loading modules correctly?

AlexanderRichert-NOAA:
The following works for me on gaea60:
```console
module use /autofs/ncrc-svm1_proj/epic/spack-stack/spack-stack-1.6.0/envs/unified-env-c6/install/modulefiles/Core
module load stack-intel
module load stack-cray-mpich
```

I'm having some trouble getting ecmwf-atlas, seemingly related to the wrong libstdc++ being used, but otherwise things seem to be going okay...

climbfuji:
@BijuThomas-NOAA Is this now working for you?

BijuThomas-NOAA:
The build is successful. Need more time to run HAFS.

climbfuji:
@BijuThomas-NOAA Were you able to run HAFS with this installation of spack-stack?

BijuThomas-NOAA:
[...]

PR #1132, opened 2024-06-06
Add gaea c6 config

AlexanderRichert-NOAA:
### Summary

Adding a gaea-c6 site config.

Note the interesting compiler flags. We want the Intel compilers to be aware of a recent-ish version of GCC for the usual reasons. On C6, GCC 12 and 13 are installed as 'gcc-12' 'gcc-13 'gfortran-12' etc., and using `-gcc-name` and `-gxx-name` appear to be the preferred ways of pointing to non-standard GCC executable names. That said, when I do that, ecmwf-atlas won't compile unless I also add `-static-libstdc++`, so I've added that. The alternatives would be to install our own GCC (my least favorite option), try a different version (not clear that this would work, like if we had the admins install GCC 11), or create aliases for the compilers so they use the regular names, then add that path to $PATH. For the last option, it would avoid the need for the `-static-libstdc++` flag, but it's kind of an ugly solution.

### Testing

Tested building 1.6.0 unified env, as well as concretizing and installing some packages based on develop.

### Applications affected

all

### Systems affected

gaea-c6

### Dependencies

none

### Issue(s) addressed

Related to #1130

### Checklist

### DIFF
--- a/configs/sites/gaea-c6/compilers.yaml
+++ b/configs/sites/gaea-c6/compilers.yaml
@@ -18,7 +18,7 @@ compilers:
       fflags: "-gcc-name=/usr/bin/gcc-12"
     environment:
       set:
-        # OpenSUSE on Gaea C5 sets CONFIG_SITE so
+        # OpenSUSE on Gaea C6 sets CONFIG_SITE so
         # Automake-based builds are installed in lib64
         # which confuses some packages.
         CONFIG_SITE: ''
--- a/configs/sites/gaea-c6/compilers.yaml
+++ b/configs/sites/gaea-c6/compilers.yaml
@@ -1,0 +1,25 @@
+compilers:
+- compiler:
+    spec: intel@2023.2.0
+    paths:
+      cc: cc
+      cxx: CC
+      f77: ftn
+      fc: ftn
+    operating_system: sles15
+    modules:
+    - PrgEnv-intel/8.5.0
+    - intel-classic/2023.2.0
+    - craype/2.7.30
+    - libfabric/1.20.1
+    flags:
+      cflags: "-gcc-name=/usr/bin/gcc-12"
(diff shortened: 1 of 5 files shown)


configs/sites/gaea-c6/compilers.yaml
```diff
@@ -0,0 +1,25 @@
...
+    - intel-classic/2023.2.0
+    - craype/2.7.30
+    - libfabric/1.20.1
+    flags:
+      cflags: "-gcc-name=/usr/bin/gcc-12"
+      cxxflags: "-gxx-name=/usr/bin/g++-12 -gcc-name=/usr/bin/gcc-12 -static-libstdc++"
+      fflags: "-gcc-name=/usr/bin/gcc-12"
+    environment:
+      set:
+        # OpenSUSE on Gaea C5 sets CONFIG_SITE so
```
climbfuji:
```suggestion
        # OpenSUSE on Gaea C6 sets CONFIG_SITE so
```
?

configs/sites/gaea-c6/modules.yaml
```diff
@@ -0,0 +1,7 @@
+modules:
+  default:
+    enable::
+    - lmod
+    lmod:
+      include:
+      - python
```
climbfuji, line 7: Not in this PR, but just as a note. We should at some point remove "Python" from the default exclude lists, and then remove all the manual include settings (basically each site has this now). This is a legacy setting from the early days of spack-stack when we didn't build Python but used an external Python instead.

climbfuji (approved):
Thanks for this! I actually like your solution for pointing to the correct GCC compiler, and the static libstdc++ isn't too bad, or? You basically resolved one or two open issues related to ecmwf-atlas not compiling with Intel when using gcc-12 as backend :-)

AlexanderRichert-NOAA:
@climbfuji I swear I wasn't trying to fix anything... ;)

PR #1138, opened 2024-06-12
Overhaul packages.yaml

AlexanderRichert-NOAA:
### Summary

This PR updates common/packages.yaml to remove unused entries, unpin versions that don't need to be pinned, and apply hard requirements as much as possible.

### Status:
99% of the entries now use `require:`.
 ```
bacio:
  version: ['2.4.1']
```
becomes
```
bacio:
  require: '@2.4.1'
```
Then if someone wants to use a different version in their own environment (or site config), they have to override it (note the ::):
```
bacio::
  require: '@2.5.0'
```

### Testing
_Make sure to check for duplicates!_
 - Tested on Alex's PC
 - Concretizes without duplicates on Hera (required '::' for ecflow entry in site config)
 - [x] Hera/Intel/UFSWM (@AlexanderRichert-NOAA)
 - [x] Hera/Intel/build full unified env (@AlexanderRichert-NOAA)
 - [x] Hera/GCC/concretization (@AlexanderRichert-NOAA)
 - [x] Gaea C5/Intel/UFSWM (@AlexanderRichert-NOAA) - ran into an unrelated issue with mapl ^cray-mpich, captured separately in #1153
 - [x] Hercules/Intel/concretization (@AlexanderRichert-NOAA)
 - [x] Hercules/GCC/concretization (@AlexanderRichert-NOAA)
 - [x] Jet/Intel/concretization (@AlexanderRichert-NOAA)
 - [x] Jet/GCC/concretization (@AlexanderRichert-NOAA)

### Applications affected

All

### Systems affected

All

### Dependencies

none

### Issue(s) addressed

Fixes #1115

### Checklist
- [x] This PR addresses one issue/problem/enhancement, or has a very good reason for not doing so.
- [ ] These changes have been tested on the affected systems and applications.
- [ ] All dependency PRs/issues have been resolved and this PR can be merged.

PR #1159, opened 2024-06-25
Overhaul packages.yaml (@climbfuji's addons)

climbfuji:
### Summary

This PR should get merged into https://github.com/JCSDA/spack-stack/pull/1138 (or it can replace #1138, since it is based on it), but for the purpose of testing in CI I am creating a PR here.

The changes on top of #1138 are:
- Unpin a few Python packages, pin a few others to resolve duplicated packages for the unified environment and the neptune standalone environment on Nautilus (w/ Intel + MKL); also works on @climbfuji's Dell laptop for both templates
- Update Nautilus site config (this site is special, since it uses MKL instead of openblas/fftw as provider for blas, lapack, fftw-api)

### Testing

- Concretized both environments on Nautilus with Intel - no unwanted duplicates
- Concretized both environments on @climbfuji's Dell laptop with GCC - no unwanted duplicates
- **MISSING** CI testing
### Applications affected

See #1138

### Systems affected

See #1138

### Dependencies

n/a

### Issue(s) addressed

See #1138

### Checklist


climbfuji:
Not yet ready for prime time. Getting lots of build errors on nautilus, seem to be related to the new setuptools or pip

climbfuji:
@AlexanderRichert-NOAA When I restrict py-numpy to 1:25 or earlier, everything builds fine on my end and in CI. Do you want to check out these additions here on top of your PR?

configs/common/packages.yaml
```diff
@@ -72,232 +65,196 @@
...
+    # also check Nautilus site config when making changes here
     py-numpy:
-      require: ['@1.22.3']
+      require: '@:1.25'
     py-pandas:
-      variants: +excel
-    # Pin py-poetry-core to avoid duplicate Python packages
-    py-poetry-core:
-      require: '@1.8.1'
+      require: '+excel'
```
AlexanderRichert-NOAA, line 218: We still need to pin the py-pybind11 to avoid duplicates on Hera when compiling with %gcc:
```
    py-pybind11:
      require: '@2.11.0'
```

climbfuji: Sigh, ok. Let me add that back in. Is that because of the old gcc 9.2.0?

AlexanderRichert-NOAA: Yes-- I tried concretizing with 13.2.0 without the pinned py-pybind11 version and it worked fine with no duplicates.

AlexanderRichert-NOAA:
I think we should lobby the Spack devs to see about a concretization option that _really_ unifies everything, including build-only dependencies, which as far as I can tell is the root of all these issues (i.e., Spack doesn't count build-only dependencies as duplicates for unify:true or unify:when_possible).

configs/common/packages.yaml
climbfuji: We could consider making that a hera + jet site requirement if it turns out to be a problem on other platforms (I am trying CI now, next my laptop, then Nautilus).

climbfuji:
> I think we should lobby the Spack devs to see about a concretization option that _really_ unifies everything, including build-only dependencies, which as far as I can tell is the root of all these issues (i.e., Spack doesn't count build-only dependencies as duplicates for unify:true or unify:when_possible).

That's a good point. Why don't you join the (almost) weekly spack user's meeting (Wed 10am MT/12pm ET) - quite often there is a number of spack developers on the call (for example today).

climbfuji:
Ubuntu CI runner had a full disk again. I fixed it and submitted a PR to a JCSDA repo to avoid this problem in the future. Hopefully the current Intel build will go through smoothly, then I will rerun GNU.

climbfuji:
@AlexanderRichert-NOAA - https://github.com/AlexanderRichert-NOAA/spack-stack/pull/3

climbfuji:
Closing, no longer needed.

PR #1186, opened 2024-07-09
Separate site configs into tier-1 and tier-2 platforms

climbfuji:
### Summary

This PR splits up the site configs into subdirectories `tier1` and `tier2`. This is done because the number of sites is getting longer, and it makes it easier to identify which ones are fully supported and which aren't. Further, we can then add more site configs in `tier2`, for example as needed by spack-stack developers. (As a follow up PR, I am going to submit my laptop site config, which has gotten increasingly complicated with multiple compilers etc.)

Also, changes to `configs/sites/**` are added to the GitHub actions ignore list for the three self-hosted runner setups.

### Testing

- [x] Tested manually on my laptop
- [ ] CI

### Applications affected

n/a

### Systems affected

All

### Dependencies

none

### Issue(s) addressed

none

### Checklist


AlexanderRichert-NOAA:
What happens if I have sites with the same name under tier1 and tier2? Will it pick up on the first one it finds? It doesn't matter, it's just for my future reference.

AlexanderRichert-NOAA (approved):

climbfuji:
> What happens if I have sites with the same name under tier1 and tier2? Will it pick up on the first one it finds? It doesn't matter, it's just for my future reference.

tier1 takes precedence over tier2 - is that what you would like to have? or error out?

AlexanderRichert-NOAA:
Yep that's fine as it is

PR #1189, opened 2024-07-10
Add site config for blackpearl

climbfuji:
### Summary

Adding the site config for @climbfuji's development system that's gotten too complicated (because of multiple compilers etc) to generate it from `linux.default` every time.

There are minor modifications to `configs/common/packages.yaml` that are needed to avoid duplicate packages, and I believe those will be necessary for other systems moving to the `oneapi` (Intel LLVM) compilers (at least `icx`, `icpx`) in the future. (Otherwise, I'd not bother you reviewing this PR).

### Testing

Tested locally

### Applications affected

n/a

### Systems affected

n/a

### Dependencies

n/a

### Issue(s) addressed

n/a

### Checklist

### DIFF
--- a/configs/common/packages.yaml
+++ b/configs/common/packages.yaml
@@ -58,12 +58,12 @@
         - any_of: ['fflags="-fp-model precise" cxxflags="-fp-model precise"']
           when: "%intel"
           message: "Extra ESMF compile options for Intel"
-        - any_of: ['']
-          when: "%gcc"
-          message: "Extra ESMF compile options for GCC"
-        - any_of: ['']
-          when: "%apple-clang"
-          message: "Extra ESMF compile options for GCC"
+        #- any_of: ['']
+        #  when: "%gcc"
+        #  message: "Extra ESMF compile options for GCC"
+        #- any_of: ['']
+        #  when: "%apple-clang"
+        #  message: "Extra ESMF compile options for GCC"
     fckit:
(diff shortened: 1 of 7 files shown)


climbfuji:
@AlexanderRichert-NOAA I need to merge this PR to proceed with a follow-up on setting the compiler. Would you mind taking a quick look, please?

configs/common/packages.yaml
```diff
@@ -58,12 +58,12 @@
...
-          message: "Extra ESMF compile options for GCC"
-        - any_of: ['']
-          when: "%apple-clang"
-          message: "Extra ESMF compile options for GCC"
+        #- any_of: ['']
+        #  when: "%gcc"
+        #  message: "Extra ESMF compile options for GCC"
+        #- any_of: ['']
+        #  when: "%apple-clang"
+        #  message: "Extra ESMF compile options for GCC"
```
AlexanderRichert-NOAA, line 66: What's the story here? Are these lines just unnecessary?

AlexanderRichert-NOAA (approved):

configs/common/packages.yaml
climbfuji: Yes, they are unnecessary.

climbfuji: The only compiler that needs extra flags is intel, and that item only gets added "when: "%intel" (should also have oneapi in the future)

PR #1192, opened 2024-07-10
Remove unused options from "spack stack" extension and add logic to set compiler for "spack stack create env"

### LINKED ISSUE #1126: Error executing setup-meta-modules
**Describe the bug**
Building the devel branch of the stack (as of 05/22/2024) with Intel then GCC results in a "Multiple version matches" error when it tries to create the meta modules.

**To Reproduce**
**Note:** I have forced the use of Open MPI 4.1.6 by requesting it in `site/packages.yaml` to facilitate using TotalView (the 5.x line has removed the MPIR process acquisition interface required by TotalView).

Build the stack with both Intel and GCC compilers then try to execute `spack stack setup-meta-modules` in either environment and you will get output similar to below:

```
$ spack stack setup-meta-modules
Configuring basic directory information ...
  ... script directory: /opt/src/spack-stack/spack-ext/lib/jcsda-emc/spack-stack/stack
  ... base directory: /opt/src/spack-stack/spack-ext/lib/jcsda-emc/spack-stack
  ... spack directory: /opt/src/spack-stack/spack
Configuring active spack environment ...
  ... environment directory: /opt/src/spack-stack/envs/devel20240523_intel
Parsing spack environment main config ...
  ... install directory: /opt/spack-stack
Parsing spack environment modules config ...
  ... configured to use lmod modules
  ... module directory: /opt/spack-stack/modulefiles
Parsing spack environment package config ...
  ... list of possible compilers: '['intel@2021.8.0', 'gcc', 'clang', 'oneapi', 'xl', 'nag', 'fj', 'aocc']'
  ... list of possible mpi providers: '['openmpi@4.1.6', 'openmpi', 'mpich']'
['intel', 'openmpi', 'gcc']
 ... stack compilers: '{'intel': ['2021.8.0'], 'gcc': ['11.4.1']}'
==> Error: Multiple version matches for openmpi@4.1.6 in /opt/spack-stack/modulefiles/openmpi: ['4.1.6-2edufp4', '4.1.6-6gupjjp']
```

Curiously, If I temporarily move `4.1.6-2edufp4`, for example, and run the `spack stack setup-meta-modules` it gets further (far enough to make the stack loadable), but still errors like below so there are no `stack-openmpi` or `stack-python` modules created.

```
$ spack stack setup-meta-modules
Configuring basic directory information ...
  ... script directory: /opt/src/spack-stack/spack-ext/lib/jcsda-emc/spack-stack/stack
  ... base directory: /opt/src/spack-stack/spack-ext/lib/jcsda-emc/spack-stack
  ... spack directory: /opt/src/spack-stack/spack
Configuring active spack environment ...
  ... environment directory: /opt/src/spack-stack/envs/devel20240523_intel
Parsing spack environment main config ...
  ... install directory: /opt/spack-stack
Parsing spack environment modules config ...
  ... configured to use lmod modules
  ... module directory: /opt/spack-stack/modulefiles
Parsing spack environment package config ...
  ... list of possible compilers: '['intel@2021.8.0', 'gcc', 'clang', 'oneapi', 'xl', 'nag', 'fj', 'aocc']'
  ... list of possible mpi providers: '['openmpi@4.1.6', 'openmpi', 'mpich']'
['intel', 'openmpi', 'gcc']
 ... stack compilers: '{'intel': ['2021.8.0'], 'gcc': ['11.4.1']}'
 ... stack mpi providers: '{'openmpi': {'4.1.6-6gupjjp': {'intel': ['2021.8.0']}}}'
  ... core compilers: ['gcc@4.6']
Preparing meta module directory ...
  ... meta module directory : /opt/spack-stack/modulefiles/Core
Creating compiler modules ...
  ... configuring stack compiler gcc@11.4.1
  ... ... CC  : /usr/bin/gcc
  ... ... CXX : /usr/bin/g++
  ... ... F77 : /usr/bin/gfortran
  ... ... FC' : /usr/bin/gfortran
  ... ... COMPFLAGS:
  ... ... MODULELOADS:
  ... ... MODULEPREREQS:
  ... ... MODULEPATH  : /opt/spack-stack/modulefiles/gcc/11.4.1
  ... writing /opt/spack-stack/modulefiles/Core/stack-gcc/11.4.1.lua
  ... configuring stack compiler intel@2021.8.0
  ... ... CC  : /opt/sw/apps/intel/oneapi/compiler/2023.0.0/linux/bin/intel64/icc
  ... ... CXX : /opt/sw/apps/intel/oneapi/compiler/2023.0.0/linux/bin/intel64/icpc
  ... ... F77 : /opt/sw/apps/intel/oneapi/compiler/2023.0.0/linux/bin/intel64/ifort
  ... ... FC' : /opt/sw/apps/intel/oneapi/compiler/2023.0.0/linux/bin/intel64/ifort
  ... ... COMPFLAGS: setenv("CFLAGS", "-diag-disable=10441")
setenv("CXXFLAGS", "-diag-disable=10441")
setenv("FFLAGS", "-diag-disable=10441")
  ... ... MODULELOADS: load("intel/2021.8.0")
  ... ... MODULEPREREQS: prereq("intel/2021.8.0")
  ... ... ENVVARS  : prepend_path("LD_LIBRARY_PATH", "/opt/sw/apps/intel/oneapi/compiler/2023.0.0/linux/compiler/lib/intel64_lin")
  ... ... MODULEPATH  : /opt/spack-stack/modulefiles/intel/2021.8.0
  ... writing /opt/spack-stack/modulefiles/Core/stack-intel/2021.8.0.lua
==> Error: Package with matching name is incompatible: ordereddict([('version', ['4.1.6']), ('variants', '~internal-hwloc +two_level_namespace')])
```

**Expected behavior**
`spack stack setup-meta-modules` should complete without errors

**System:**
Linux Workstation running Rocky 9. Intel is 2023.0.0 (2021.8.0 legacy compilers (icc,icpc,ifort)) and GCC is 11.4.1 system package manager install.

**Additional context**
This appears to be somewhat recent and is possibly due to the fact that spack now adds the shortened hash to the openmpi (and presumably other MPI implementations) version sub-directories. In previous versions, `/path/to/spack-stack/modulefiles/openmpi` would contain version number (_i.e._ 4.1.6) folders. However, now the hash is added to the folder name so instead of `/path/to/spack-stack/modulefiles/openmpi` containing a single 4.1.6 folder with `intel` and `gcc` sub-directories, there is a `4.1.6-<intel_built_hash>` and a `4.1.6-<gcc_built_hash>` folder.

### LINKED ISSUE #1167: Only use one (principal) compiler per enviroment
The spack-stack developers decided to limit environments to one compiler only. Strictly, that is one principal compiler as the default for all packages, with the option to use a fallback compiler for certain packages that are incompatible.

We can achieve this either via
```
packages:all:prefer:['%oneapi']
```
or
```
packages:all:require:['%oneapi']
```
with additional "requires" for individual packages to override. This is what E4S does, and it gives you more contro (see https://github.com/spack/spack/blob/develop/share%2Fspack%2Fgitlab%2Fcloud_pipelines%2Fstacks%2Fe4s-oneapi%2Fspack.yaml#L64)l. The downside is more settings in our configs.

We also want to add the option to set the compiler as part of `spack stack create env` and remove the compiler matrices that we use in some of our templates. We also need to update the CI scripts and instructions accordingly.

**Additional context**
n/a

climbfuji:
### Summary

This PR removes unused options for the "spack stack" extension in an order to clean up the amount of code and facilitate adding the new capability described below.

The new capability requires the user to define a compiler `--compiler` when running `spack stack create env`. It further updates the logic for creating common and site environment spack config files for modules and packages. For both common and site configs (in `configs/common/` and `configs/sites/tier?/*/`), the following is true:
1. For `packages`: If both `packages.yaml` and `packages_${COMPILER}.yaml` exist in the same directory, merge the two (`packages_${COMPILER}.yaml` overrides `packages.yaml` if necessary) and write to the environment's common or site config as `packages.yaml`. See for example the updated `blackpearl` site config in this PR. If only one of them exists, copy that file to the destination.
2. For `modules`: If both `modules.yaml` and `packages_${LUA_OR_LMOD}.yaml` exist in the same directory, merge the two (`packages_${LUA_OR_LMOD}.yaml` overrides `modules.yaml` if necessary) and write to the environment's common or site config as `modules.yaml`. This is currently not the case, but maybe we can consolidate some of the contents into one default `modules.yaml` file?

In the environment `spack.yaml`, update the compiler definition line for the matrix (if applicable) and set `packages:all:require:['%COMPILER']`. This requires that packages that must be built with other compilers (e.g. `gcc-runtime`) have their own `require: '%COMPILER'` setting to overwrite this. This is already the case for `gcc-runtime` in `configs/common/packages.yaml`. The same may be needed for other `runtime` packages for each of the compilers in the future, and for packages like `bison` that don't build with `oneapi` for example. I am basically following what E4S does.

### Testing

- [x] Created unified-env and empty environments with both `gcc` and `oneapi` on `blackpearl`, concretized and verified that the blas/lapack/fftw-api providers are correct, the correct compilers are being used, and that no conflicts are reported (except the known/greenlighted ones).
- [x] CI (this represents testing new site configs for different compilers and OSs)
- [x] Built `neptune-env` on Nautilus with `oneapi@2024.1.2` using the existing site config (i.e. no changes necessary). Then built `NEPTUNE` with the same compiler and ran basic tests.
- [x] Built intel-impi container with jedi-ci specs

### Applications affected

None

### Systems affected

None

### Dependencies

- [x] waiting on https://github.com/JCSDA/spack/pull/452

### Issue(s) addressed

Resolves https://github.com/JCSDA/spack-stack/issues/1167
Resolves https://github.com/JCSDA/spack-stack/issues/1126

### Checklist


climbfuji:
@AlexanderRichert-NOAA I am keeping this in draft for the moment, but I would really like to ask you for your feedback as I've been rather aggressive with removing unused options in our `spack stack` extension.

The new feature to set the compiler in `spack stack create env` works very well for as far as I have tested.

climbfuji:
Of course, I am going to fix the unit tests and update the documentation. But I won't do that until we've agreed on the actual changes.

AlexanderRichert-NOAA:
Okay, thanks for the headsup. So far so good I think. I don't really use any of those, and it seems like they're things that can be dealt with manually by copying files around. If you can think of an exception to that let me know.

climbfuji:
Thanks @AlexanderRichert-NOAA. This is all looking pretty good. I'll finish the GitHub actions, test container builds and also building locally on an NRL system, push the documentation updates and open it up for reviews. Would be great to get this merged before I go on leave end of this week (for three weeks).

climbfuji:
@AlexanderRichert-NOAA This is now ready for review (and additional testing if you feel inclined to do so)

climbfuji:
@RatkoVasic-NOAA @natalie-perlin FYI

srherbener (approved):
I've tested on Discover and my Mac. Discover successfully made it through the spack-stack build. My Mac made it through concretize and failed with the `musl` issue during install. I think the Mac issue is limited to apple-clang%15 and not related to this PR, so I'm happy to approve. Thanks!

climbfuji:
Thanks so much for testing @srherbener - @AlexanderRichert-NOAA any chance you can take a look at this PR in the next week or two?

AlexanderRichert-NOAA:
Sorry @climbfuji this got stuck in the hopper. I'm going to test it out momentarily...

spack-ext/lib/jcsda-emc/spack-stack/stack/stack_env.py
```diff
@@ -232,12 +253,22 @@ def write(self):
                 original_sections[key] = copy.deepcopy(section)
 
         # Commonly used config settings
-        if self.compiler:
-            compiler = "packages:all::compiler:[{}]".format(self.compiler)
-            spack.config.add(compiler, scope=env_scope)
-        if self.mpi:
-            mpi = "packages:all::providers:mpi:[{}]".format(self.mpi)
-            spack.config.add(mpi, scope=env_scope)
+        compiler = f"packages:all:require:['%{self.compiler}']"
```
AlexanderRichert-NOAA: This doesn't work. When used in conjuction with, e.g., `packages:gcc-runtime:require:'%gcc'`, this general setting doesn't get respected. In a test concretization just now, hdf5 and other packages are getting spec'd with %gcc. This is something I've run into previously. The solution appears to be to use `prefer` instead of `require` here (but continue to use `require` for individual packages like gcc-runtime).

AlexanderRichert-NOAA (changes requested):

spack-ext/lib/jcsda-emc/spack-stack/stack/stack_env.py
AlexanderRichert-NOAA: Actually, I think we can remove the `packages:gcc-runtime:require:'%gcc'` from common/packages.yaml because that constraint is already present in the gcc-runtime recipe itself.

climbfuji: This is supposed to work (and it did for me). E4S does the same, use require self.compiler forces using the compiler you want for all packages, but individual packages can overwrite. Which is what we want - we WANT gcc-runtime to be built with gcc always, regardless of what the principal compiler for the eng is.

climbfuji: @AlexanderRichert-NOAA See here for the E4S config: https://github.com/spack/spack/blob/develop/share%2Fspack%2Fgitlab%2Fcloud_pipelines%2Fstacks%2Fe4s-oneapi%2Fspack.yaml#L64

AlexanderRichert-NOAA: Steps to reproduce on my CentOS 8 PC (commit 445aca64a55c562db6a90913e075318ff4d692ae):
```console
 $ . setup.sh
 $ spack stack create env --name test-env --template unified-dev --site linux.default --compiler oneapi
 $ spack env activate envs/test-env
 $ spack compiler find # detect oneapi
 $ spack compiler list
==> Available compilers
-- gcc centos8-x86_64 -------------------------------------------
gcc@8.5.0

-- oneapi centos8-x86_64 ----------------------------------------
oneapi@2024.0.0
 $ spack concretize | grep %gcc
 -   shcc26y          ^gcc-runtime@8.5.0%gcc@8.5.0 build_system=generic arch=linux-centos8-zen
 -   xbichoa              ^hdf5@1.14.3%gcc@8.5.0~cxx+fortran+hl~ipo~java~map+mpi+shared~subfiling~szip+threadsafe+tools api=default build_system=cmake build_type=Release generator=make patches=82088c8 arch=linux-centos8-zen
 -   ktbqw7d                  ^cmake@3.27.9%gcc@8.5.0~doc+ncurses+ownlibs build_system=generic build_type=Release arch=linux-centos8-zen
 -   r43pbyy          ^bison@3.8.2%gcc@8.5.0~color build_system=autotools arch=linux-centos8-zen
 -   r43pbyy                      ^bison@3.8.2%gcc@8.5.0~color build_system=autotools arch=linux-centos8-zen
...
```

I can also reproduce this problem on Hera. If I add oneapi to my config and concretize, bison, hdf5, and python end up being spec'd with '%intel'. If I take intel out of the available compilers, then those packages end up as %gcc. I think somehow the %oneapi conflict in the bison recipe leads to the general requirement being nullified.

climbfuji: Interesting. If you look at the packages in common/packages.yaml that I set to require `%gcc`, then you see that this is bison and everything that bison depends on. hdf5 shouldn't be affected at all, and neither should be cmake. Weird, because I don't have this problem on my Oracle Linux 9. What did you say works for you? I unfortunately don't have my Dell with me on vacation, therefore I won't be able to test it there (but I can test on Ubuntu 22)

climbfuji: We should probably report this to spack as a bug as well?

AlexanderRichert-NOAA: I'll retest with 'prefer' and file a bug.

climbfuji: I am testing the "prefer" version too. So far looking ok on macOS with apple-cland and on Ubuntu with oneapi (plus the CI tests - just need to verify it used the compilers as intended).

AlexanderRichert-NOAA:
Looks like py-torch specifically wants sleef@3.5.1_2020-12-22, but oneapi support wasn't added until 3.6. Maybe remove the ai piece for now?

configs/templates/unified-dev/spack.yaml
```diff
@@ -42,7 +42,8 @@ spack:
     - [$packages]
     - [$compilers]
     exclude:
-    # py-torch in ai-env doesn't build with Intel
+    # py-torch in ai-env doesn't build with Intel; deps don't work for oneapi
     - ai-env%intel
+    - ai-env%oneapi
     # jedi-tools doesn't build with Intel
     - jedi-tools-env%intel
```
climbfuji, line 49:
```suggestion
    - jedi-tools-env%intel
    - jedi-tools-env%oneapi
```

climbfuji:
@AlexanderRichert-NOAA CI tests now pass (I turned off the env chaining for the new oneapi workflow). Can you please review again?

AlexanderRichert-NOAA (approved):

PR #1193, opened 2024-07-11
Re-enable cylc testing in skylab-dev in GitHub actions

### LINKED ISSUE #1162: Turn off cylc variant for ewok-env by default, but make sure we test building cylc regularly
The cylc packages introduce a lot of Python dependencies that make concretization and building difficult and lengthy in the unified environment. Nobody so far is using the spack-stack cylc installation.

NRL has an interest in using cylc from spack-stack, but doesn't need the full unified environment. We want to make sure that we continue testing the build of the cylc packages, so maybe it's possible to add it to the neptune environment (as a variant or as default). Or found another way (a chained environment tested in CI?).

**Additional context**
n/a

climbfuji:
### Summary

Unpin `py-jinia`, re-enable `cylc` variant for `skylab-dev` template, build both `skylab-dev` and `unified-dev` in GitHub actions on Ubuntu with GCC.

### Testing

- [x] CI

### Applications affected

n/a

### Systems affected

n/a

### Dependencies

n/a

### Issue(s) addressed

Resolves https://github.com/JCSDA/spack-stack/issues/1162

### Checklist


.github/workflows/ubuntu-ci-x86_64-gnu.yaml
```diff
@@ -53,88 +53,99 @@ jobs:
...
-
-          # Remove buildcache config settings
-          spack config remove config:install_tree:padded_length
 
-          # Next steps: synchronize source and build cache to a central/combined mirror?
-
-          # Cleanup
-          spack clean -a
-          spack env deactivate
+          declare -a TEMPLATES=("unified-dev" "skylab-dev")
```
climbfuji, line 57: There are no code changes between lines 66 and 146 - they are the same as lines 56 to 137 on the left side. The only diff is the indentation for the loop over templates.

AlexanderRichert-NOAA (approved):

climbfuji:
Thanks for reviewing the PR @AlexanderRichert-NOAA - glad to see that we don't need to pin `py-jinja` at all.

AlexanderRichert-NOAA:
Yeah- I had left it since it's one of those build deps that pops up all over the place so it might need to be pinned again in the future, but I think commenting it is the way to go.

climbfuji:
> Yeah- I had left it since it's one of those build deps that pops up all over the place so it might need to be pinned again in the future, but I think commenting it is the way to go.

Rahul is also checking whether 3.0.3 works with wxflow - if it does, we can go back to that version, too.

Issue #1199, opened 2024-07-11
[INSTALL]: Update to wgrib2 3.3.0?


climbfuji:
### Package name

wgrib2

### Package version/tag

3.3.0 - https://github.com/NOAA-EMC/wgrib2/releases/tag/v3.3.0

### Build options

Need to be checked - this is a new build system

### Installation timeframe

See https://github.com/NOAA-EMC/wgrib2/releases/tag/v3.2.0 - is it possible to include this in spack-stack-1.8.0?

### Other information

My hope is that the new cmake build system will solve the constant problems we've had with building wgrib2.

### WCOSS2

- [X] Check this box if and only if your package should be installed on WCOSS2 Cactus and Dogwood (all spack-stack packages will be installed on Acorn). If not, you may disregard the rest of the items below and submit this request.

### WCOSS2: General questions

Already approved for wcoss2, and developed by NOAA-EMC

### WCOSS2: Technical & security review list

- [ ] The code is mature, stable, and production ready
- [ ] The code is does not and cannot use the internet, and does not contain URLs (http, https, ftp, etc.) except in comments
- [ ] The package does not contain prebuilt binary files that have not been approved by NCO security review
- [ ] The code has no publicly disclosed cybersecurity vulnerabilities and exposures (search https://cve.mitre.org/cve/)
- [ ] The code is not prohibited by DHS, DOC, NOAA, or NWS
- [ ] The code comes from a trusted source. Trusted sources include other NWS, NOAA, or DOC, agencies, or other Federal agencies that operate at a FISMA high or equivalent level. Additionally, trusted sources could be third-party agencies through which there is an existing SLA on file (such as RedHat).
- [ ] The code is actively maintained and supported (it continues to get updates, patches, etc.)
- [ ] The code is not maintained by a private entity operating in a foreign country (if it is, make a note below)
- [ ] There is sufficient documentation to support maintenance
- [ ] There are no known security vulnerabilities or weaknesses
- [ ] Installing and running the code does not require privileged processes/users
- [ ] There are no software dependencies that are unapproved or have security concerns (if there are, make a note below)
- [ ] There are no concerns related to SA, SI, and SC NIST control families

AlexanderRichert-NOAA:
[...]

Issue #967, opened 2024-01-25
wgrib2 2.0.8/3.1.1 do not build with Intel oneAPI LLVM compilers


climbfuji:
**Describe the bug**
```
==> Installing wgrib2-2.0.8-izv2torii27jtoc2eqvf3pu5khxb2gfj [74/362]
==> No binary for wgrib2-2.0.8-izv2torii27jtoc2eqvf3pu5khxb2gfj found: installing from source
==> Error: UnboundLocalError: local variable 'comp_sys' referenced before assignment

/home/ubuntu/sandpit/spack-stack/spack/var/spack/repos/builtin/packages/wgrib2/package.py:160, in setup_build_environment:
        157        elif self.spec.compiler.name in ["gcc", "clang", "apple-clang"]:
        158            comp_sys = "gnu_linux"
        159
  >>    160        env.set("COMP_SYS", comp_sys)
```

This is one of many problems when trying to build wgrib2 (2.0.8 or 3.1.1) with the oneAPI compilers (`icx`, `icpx`). Note that with the release of oneAPI 2024.0, Intel has removed the classic C/C++ compilers and for now the best/vetted combination is `icx/icpx/ifort`.

We therefore need a solution for `wgrib2` asap - as a workaround, I will remove `wgrib2` from the stack if the Intel oneAPI compilers are used.

**To Reproduce**
Build with Intel oneAPI LLVM compilers (`icx`, `icpx`)

**Expected behavior**
It recognizes the compiler and builds

**System:**
New AWS ParallelCluster using Ubuntu 22.04 and Intel oneAPI 2024.0.2 (there are no classic C/C++ compilers anymore)

**Additional context**
@AlexanderRichert-NOAA @edwardhartnett @aerorahul At first I thought I could fix this along the way because it looked like a simple problem, but solving one problem led to another and I don't have the bandwidth to work on this. I'll therefore remove `wgrib2` from the environment if the LLVM/oneAPI compilers are used so that I can continue building and testing (there are other package failures, too) until someone has come up with a solution for `wgrib2`.

climbfuji:
A workaround was put in place to skip wgrib2 as dependency with oneapi in all our virtual packages. Hopefully, we'll have an all-new wgrib2 package soon with a cmake build system that also builds with oneapi. This may be the case for the wgrib2@3.4.0 release (release 08/15/2024) - tbd.

edwardhartnett:
[...]

Issue #912, opened 2023-12-18
Move to Intel oneapi (LLVM) compilers for C, C++


climbfuji:
**Is your feature request related to a problem? Please describe.**

From @michalakes:
> With its 2024 release of OneAPI, Intel is deprecating its "classic" versions of Intel Fortran (ifort) and Intel C/C++ in favor of it's new LLVM-based versions. The C/C++ classic compilers are already deprecated in the 2024 release of OneAPI and classic Fortran will be deprecated by the end of 2024.

>The announcements are here:
>https://community.intel.com/t5/Intel-oneAPI-Data-Parallel-C/DEPRECATION-NOTICE-Intel-Fortran-Compiler-Classic-ifort/td-p/1545790
>https://community.intel.com/t5/Intel-oneAPI-Data-Parallel-C/REMOVAL-NOTICE-Intel-C-Compiler-Classic/td-p/1545804

Based on what I have found so far, we need the 2024 oneapi release, because Python 3.9 and later don't built with oneapi compilers up to the latest 2023 release:
- https://github.com/python/cpython/issues/106424
- https://github.com/spack/spack/pull/40536

While at it, I also came across an issue building bison with the oneapi compilers, adding it here as a reference/reminder:
- https://github.com/spack/spack/issues/37172

**Describe the solution you'd like**

**Note that we need oneapi@2024.2.0 or later because of bugsin previous releases**

1. Start with the NEPTUNE standalone environment on Nautilus
- [x] Have Intel oneAPI compilers installed on NAVY Nautilus
- [x] Nautilus: Switch site config to use this compiler's `icx` and `icpx` (but not yet `ifx`)
- [x] Make necessary updates to packages and spack configs so that `neptune-dev` builds
- [x] Nautilus: Build NEPTUNE standalone and run tests
2. For NOAA/UFS
- [x] Have Intel oneAPI compilers installed on a RDHPCS platform of choice
- [x] Locally, switch site config to use this compiler's `icx` and `icpx` (but not yet `ifx`; keep using `ifort`)
- [x] Make necessary updates to packages and spack configs so that `unified-dev` builds
- [x] Build ufs-weather-model and run tests
3. For JCSDA/jedi-bundle
- [x] Have Intel oneAPI compilers installed on a platform of choice
- [x] Locally, switch site config to use this compiler's `icx` and `icpx` (but not yet `ifx`; keep using `ifort`)
- [x] Make necessary updates to packages and spack configs so that `unified-dev` builds
[...]

PR #1200, opened 2024-07-12
More updates for Intel oneAPI and consolidation of common/modules*.yaml

climbfuji:
### Summary

This PR contains the following changes:

1. Based on the functionality to merge `modules_${LMOD_OR_TCL}.yaml` with `modules.yaml`, strip out the common settings from `moduels_lmod.yaml` and `modules_tcl.yaml` and put them in `modules.yaml`.

2. Additional updates for building the full unified environment with the oneAPI C/C++ compilers:
    - Add `gmake`, `libbsd`, `libm` to modules exclude list
    - Currently only in the `blackpearl` site config (once this all works, it can be moved to `common/packages_oneapi.yaml`): build `bison`, `gmake`, `libm`, `libbsd` with `gcc`. That's because `bison` doesn't build with `oneapi` (it builds but produces garbage, see https://github.com/spack/spack/issues/37172)

3. In `configs/templates/unified-dev/spack.yaml`: exclude `jedi-tools-env` and `ai-env` for `oneapi`, similar to what's done for `intel` (this was also done in a previous PR, therefore just cleaning up the comments in the file).

With these changes, the unified environment builds with `oneapi@2024.1.2` with the exception of `wgrib2` (see https://github.com/JCSDA/spack-stack/issues/967 and https://github.com/JCSDA/spack-stack/issues/1199)

### Testing

- [x]  CI
- [x] Blackpearl with `oneapi` and `gcc`

### Applications affected

n/a

### Systems affected

Nautilus and Blackpearl - these are the only sites that have `oneapi` compilers configured

### Dependencies

- [x] waiting on #1192

### Issue(s) addressed

More steps towards https://github.com/JCSDA/spack-stack/issues/912

### Checklist


srherbener (approved):

climbfuji:
@AlexanderRichert-NOAA I think this is ready for your review (and testing if inclined to do so)

configs/common/modules_tcl.yaml
```diff
@@ -35,18 +34,21 @@ modules:
       - freetype
       - gcc-runtime
       - gdbm
+      - gmake
```
AlexanderRichert-NOAA, line 37: Why is gmake blacklisted? This one I could sort of imagine developers wanting to access.

climbfuji: For one, if you blacklist something the modules that need it have the necessary paths in their own modulefiles. And also, even if not there is always a gmake on the system and in the default path. We need to exclude all modules from being generated that are created with a compiler different than the preferred compiler, or our module logic won't work.

AlexanderRichert-NOAA (approved):

PR #1208, opened 2024-07-19
Ensure that required tests always run.

eap:
This change adds "notest" variants of existing required tests. The purpose of this change is to ensure that required tests always run without forcing them to consume resources with a full build when this would be unnecessary. While the main tests use the `pull_request: ↳ paths-ignore:` directive to trigger the workflow, this abbreviated test uses the same file patterns but applies the inclusive `paths` directive.

### DIFF
--- a/.github/workflows/macos-ci-aarch64-notest.yaml
+++ b/.github/workflows/macos-ci-aarch64-notest.yaml
@@ -1,0 +1,24 @@
+name: macos-ci-aarch64-build
+on:
+  pull_request:
+    paths:
+      - 'configs/sites/**'
+      - 'doc/**'
+      - '**.md'
+      - '.github/ISSUE_TEMPLATE/*'
+      - '.github/workflows/macos-ci-aarch64-notest.yaml'
+      - '.gitignore'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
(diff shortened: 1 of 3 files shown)


climbfuji (commented):
Do we really want this? I prefer to know that tests didn't run when I go back and look at the commits that were merged. This sort of fakes that tests were run when they actually weren't.

eap:
@climbfuji 

Yeah, that's a valid question. The problem we have is that we require these tests for merging so any PR that only includes files matching the `paths-ignore` filter will not be mergable without an override. GitHub doesn't give any guidance on the proper course of action here and the advice I've seen seems split between 3 choices

1) Don't require tests that don't always need to run.

2) Run required tests on every PR even if no affected code is changed.

3) Make a workflow with the same name that executes against excluded paths.

I think #3 isn't as viable as I originally thought, it would probably work in a binary case but this PR shows the weakness; in case you have a test that mixes paths then the workflow run is a race condition. 

So... I don't know the best path here. I'm inclined against requiring tests that may not run but due to resource limits I really don't like activating all tests for no reason. We have plenty of PRs that only update docs.

climbfuji:
In this case I'd say let's do (2), because we have the runners up and running 24/7 anyway. And the rebuilds from the buildcache take only a few minutes.

eap:
moving the changes discussed here to https://github.com/JCSDA/spack-stack/pull/1210

Issue #1184, opened 2024-07-08
Run package tests with every deployment


AlexanderRichert-NOAA:
I suggest we add unit testing to our standard installation steps on supported platforms. This will help us catch various issues early, including platform config issues, portability problems, compiler issues, and so on. I'd love to have something in place for the 1.8.0 release, even if it's just for a handful of packages (NCEPLIBS would be a good place to start). We don't have run-time (as opposed to install-time) tests implemented for as far as I know any of our packages, so I think it makes sense to focus on install-time tests, though that's certainly up for discussion.

I'm thinking of changing our installation procedures to something to the effect of:
```console
# Set up environment as usual
spack concretize 2>&1 | tee log.concretize
# New step that will test packages specified in, say, setup.sh:
spack install --verbose --test root $SPACK_STACK_PACKAGES_TO_TEST 2>&1 | tee log.install_withtesting
# Install the rest of the stack as usual:
spack install --verbose 2>&1 | tee log.install
```

High-priority packages for EMC:
- MAPL
- ESMF

Here is a list of packages that can easily be tested right off the bat; I'll create a PR with setup.sh and doco updates:
- libpng
- libaec
- jasper
- scotch
- w3emc
- bacio
- g2
- g2c

Here are some packages that have testing but need some kind of fixing:
 - ip - failing for some reason, needs investigation
 - grib-util - needs small recipe fix in g2c so that utils are built
 - netcdf-c - recipe needs to be updated to ensure correct test target name is used
 - hdf5 - it works, but we should disable the parallel tests so that we don't run into trouble running with MPI until we're ready for that

climbfuji:
Let's talk about this at tomorrow's spack-stack meeting. I agree with you 100% that regular testing on tier-1 platforms is needed. We did have similar ideas and efforts in the past, and there are even open (and closed) issues for that.

edwardhartnett:
This is essential.

A recent example was the parallel I/O problems that absorbed a lot of debugging time, and ended up being a problem with HDF5-1.14.0. However, running the HDF5 tests showed the problem right away.

If the tests had been run at install, an expensive debugging effort would have been saved. Countless hours were spent checking various components before this was resolved. 

The entire IO stack needs to be tested on every single install.

NetCDF and HDF5 need to have their parallel tests run as well. I can work with some slurm expert to get these tests to run out of the box on future releases of netcdf-c, and potentially HDF5.

[...]

Issue #365, opened 2022-10-01
Use spack binary caches to greatly speed up the installation of spack environments


climbfuji:
**Is your feature request related to a problem? Please describe.**
Installing spack environments is slow, because everything gets compiled from source.

**Describe the solution you'd like**
`spack` binary caches allow caching installed packages locally and reuse them if the (unique) hash for the package is the same. This can greatly reduce the time to install environments, from hours to minutes, depending on how many packages have changed.

**Additional context**
The binary caches serve as a second local cache, in addition to the spack mirrors (see #364) to reduce network traffic and speed up the install process.

Test build caching based on 1.4.0 installations on the following systems (by creating the cache based on unified-env, install a new copy wherever, and time it):
- [ ] Acorn (@AlexanderRichert-NOAA) (time; location)
- [ ] ...

climbfuji:
We made significant progress on this in the first quarter of 2023. We are now using (simplified versions of) build caches for the CI tests running on self-hosted Github runners, which brought the build times down from over 6 hours to less than 30 minutes for the entire unified-dev environment.

There are a few issues that still need to be sorted out and that have to do with package relocation and the use of `config:install_tree:padded_length:VALUE`. We are avoiding this right now in the CI tests by always building in the same directory (i.e. using the same environment name) so that we don't have to pad the path when creating build caches.

I don't think it's realistic to have all this fleshed out so that we can roll it out on the HPCs and user systems, and have a hierarchical tree of binary caches in place (local, upstream on S3, ...). Therefore I defer completing this task to the next quarter / spack-stack release 1.4.0 by end of June 2023.

AlexanderRichert-NOAA:
@ulmononian @mark-a-potts per our discussion today can you test out creating/using build caches on, say, Hera and Cheyenne and add those systems to the list at the top of this issue? The test installation doesn't need to be in an official space for now, as I think the next step will be to nail down locations.

eap:
What is the status of this? We are using the binary cache in our test runner system, is the cache being used elsewhere?

I just want to understand how to move forward on this bug or if it should be tracked in a number of separate platform issues.

climbfuji:
This is going to get addressed in the next few weeks as part of our automation efforts - the issue here covers basically all supported platforms, not just the CI runner.

climbfuji:
Closed via #1213

PR #1213, opened 2024-07-30
Add weekly build/test scripts

AlexanderRichert-NOAA:
### Summary
This PR will add scripting for weekly unit testing and build caching of the unified env through Jenkins, cron jobs, etc.

Also a few updates for Acorn, including a change to `spack stack create env --compiler` that allows compiler version to be specified (`spack stack create env --compiler intel` -> `envs/myenv-intel`; `spack stack create env --compiler intel@2022` -> `envs/myenv-intel-2022`), which is needed to differentiate on ~~platforms~~ platform where more than one version of a compiler is installed.

### Testing/implementation by platform
On each system:
- Configure ShellSetup.sh (including setting centralized build_cache dir under spack-stack/ dir)
- Set up centralized source cache
- Get it running (Jenkins, cron, GHA); test installing a few packages
- Enable notifications (email, Slack, whatever)

Platforms (check when all of the above items are complete):
 - [x] Acorn - @AlexanderRichert-NOAA - cron
 - [x] Orion - @rickgrubin-tomorrow test with Jenkins on Orion
 - [x] Nautilus - @climbfuji - either cron or Github actions

### Applications affected
all

### Systems affected
all (eventually)

### Dependencies
none

### Issue(s) addressed
#1184, #365

### Checklist


AlexanderRichert-NOAA:
closing temporarily so I don't spam the CI with jobs...

AlexanderRichert-NOAA:
Cron entry for Acorn: `0 1 * * SUN /path/to/spack-stack-weekly-build/util/weekly_build/SpackStackBuildCache_AllSteps.sh $(date +\%y\%m\%d) /path/to/spack-stack/weekly_build acorn > /path/to/spack-stack/weekly_build/logs/cron.$(date +\%y\%m\%d).out 2>&1`

AlexanderRichert-NOAA:
@climbfuji what did you say would be a reasonable pad length for generating the build caches? Will that vary across systems and/or do we want whatever default we choose to be overridable?

climbfuji:
> @climbfuji what did you say would be a reasonable pad length for generating the build caches? Will that vary across systems and/or do we want whatever default we choose to be overridable?

I'd use 256 - works on macOS and Linux

rickgrubin-tomorrow:
Tried this (manually) for orion (using a personal workspace) without submitting a batch job; works as expected. 

I think that approach can be accommodated by Jenkins -- working on integrating something similar into a pipeline.

util/weekly_build/ShellSetup.sh
```diff
@@ -0,0 +1,95 @@
...
+      rc=$?
+      set -e
+      cat spack-build-cache-${RUNID}*
+      return $rc
+    }
+    PACKAGES_TO_TEST="libpng libaec jasper w3emc g2c"
+    PACKAGES_TO_INSTALL="ufs-weather-model-env global-workflow-env upp-env"
+    INSTALL_OPTS="-j6"
+    function alert_cmd {
+      mail -s 'spack-stack weekly build failure' alexander.richert@noaa.gov  < <(echo "Weekly spack-stack build failed in $1. Run ID: $RUNID")
```
BruceKropp-Raytheon: This appears to be a bug: '< <(echo...'

AlexanderRichert-NOAA: How so? Is this breaking things on other platforms?

util/weekly_build/4_SpackInstall.sh
```diff
@@ -0,0 +1,27 @@
...
+if [ -z $SETUPDONE ]; then . ShellSetup.sh $* ; fi
+
+cd $RUNDIR/${RUNID}
+
+set +x
+. setup.sh
+set -x
+
+INSTALL_OPTS="--show-log-on-error --fail-fast $INSTALL_OPTS"
+
```
BruceKropp-Raytheon: seeing a problem - infinite chain of links under ./envs/build-intel/install/__spack_path_placeholder__/ -> ....

AlexanderRichert-NOAA: I'm not sure I follow. INSTALL_OPTS can be set in ShellSetup.sh (for instance I'm using `-j6` on Acorn), so any platform-specific options will get added to the defaults here. Is there a platform where this is not working?

BruceKropp-Raytheon (commented):
found a few problems while running within a Jenkinsfile (see comments inline).

util/weekly_build/4_SpackInstall.sh
BruceKropp-Raytheon: if you can login to jenkins .. take a look here:
https://jenkins.epic.oarcloud.noaa.gov/job/spack-stack/job/nightly/7/execution/node/85/ws/orion-intel/weekly/envs/build-intel/

util/weekly_build/ShellSetup.sh
BruceKropp-Raytheon: All platforms are affected.  I've had to remove the space between '< <' ... then I get a log file output.

AlexanderRichert-NOAA: Hm, strange. What shell is that running in?

BruceKropp-Raytheon: correction ... for portability to sh, replace '< <' with a dollar-sign ...
`sed 's|< <| $|g' -i $(pwd)/ShellSetup.sh`

util/weekly_build/4_SpackInstall.sh
AlexanderRichert-NOAA: It says I don't have overall/read permissions.
It shouldn't be using links at all, so I think I'll need to see the output.

util/weekly_build/ShellSetup.sh
AlexanderRichert-NOAA: In Bash:
```bash
~ $ cat < <(echo blah)
blah
~ $ cat < $(echo blah)
-bash: $(echo blah): No such file or directory
```
Which shell are you using? If nothing else I can use a pipe instead, but I'm surprised it's causing issues.

util/weekly_build/4_SpackInstall.sh
BruceKropp-Raytheon: yeah. it might not be links .. just deep directory path to the .spack-db

orion-intel/weekly/envs/build-intel/install/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_/.spack-db/

Either way, I don't see any commands being executed beyond setup.sh

AlexanderRichert-NOAA: That's expected, that's part of the path padding for the build caching. It uses an extra long path in order to guarantee that there's space in, say, a shared library to search-and-replace the new installation path.

BruceKropp-Raytheon: okay. sure.
So what is supposed to occur during 4_SpackInstall.sh ?
I don't see much happening in the log output.

AlexanderRichert-NOAA: It should be fetching packages (`spack fetch`) then installing (`spack install`). I'm not sure why it wouldn't be showing output, maybe a slow network connection or filesystem? Where does it seem to be stuck?

BruceKropp-Raytheon: see details in slack.

util/weekly_build/ShellSetup.sh
BruceKropp-Raytheon: This file is sourced from each of the utility scripts here, which all run with /bin/bash .
However, in Jenkinfile pipeline, the shell is sh ... so this throws an error.

BruceKropp-Raytheon: Also, as an aside, I would reserve notifications for an outer CI mechanism.. rather than inside source scripts.

AlexanderRichert-NOAA: In the case of running this through a cron job, there is no outer mechanism. But needless to say this can be configured on each system as needed.

BruceKropp-Raytheon: okay.

util/weekly_build/4_SpackInstall.sh
BruceKropp-Raytheon: For Orion and Hercules, the problem I see may be related to them both being in the same filesystem, causing collision on pre-cleaning.  Trying a new way to overcome this,

AlexanderRichert-NOAA: Looking at the logs, it appears it's failing with `/usr/bin/tar: Relink '/apps/spack-managed/gcc-11.3.1/intel-oneapi-compilers-2023.1.0-sb753366rvywq75zeg4ml5k5c72xgj72/compiler/2023.1.0/linux/compiler/lib/intel64_lin/libimf.so' with '/lib64/libm.so.6' for IFUNC symbol 'sincosf'`.

AlexanderRichert-NOAA:
As a heads-up, I had a strange cmake failure when building hdf5 on Acorn that seems to come from the installation path padding. I reduced it from 256 to 200 and that fixed it; it was failing with a padded_length of 203.

climbfuji:
> As a heads-up, I had a strange cmake failure when building hdf5 on Acorn that seems to come from the installation path padding. I reduced it from 256 to 200 and that fixed it; it was failing with a padded_length of 203.

I've seen this before, and I believe I had it at 200 for that reason, too (in the JCSDA AWS pcluster nightly builds). We should set ot to that value instead of `padded_length: true` and add a check that the install path is no longer than that.

climbfuji:
I started working on this. There are a number of things that we need to update/make more flexible imo. I'll leave a few comments for starters, more will follow next week and maybe also a PR to your branch with some updates

util/weekly_build/SpackStackBuildCache_AllSteps.sh
```diff
@@ -0,0 +1,25 @@
...
+export SETUPDONE=YES
+
+function trap_and_run {
+  trap ERROR ERR
+  scriptname=$1
+  function ERROR { alert_cmd $scriptname ; exit 1;}
+  $*
+}
+
+trap_and_run ./1_DirectorySetup.sh $*
```
climbfuji: All these files should have two-digit numbers: 01_DirectorySetup.sh etc

util/weekly_build/ShellSetup.sh
```diff
@@ -0,0 +1,101 @@
...
+  discover17)
+    COMPILERS=${COMPILERS:-"intel gcc"}
+    BUILD_CACHE_DIR=${BUILD_CACHE_DIR:-/gpfsm/dswdev/jcsda/spack-stack/scu17/build_cache}
+    ;;
+  derecho)
+    COMPILERS=${COMPILERS:-"intel gcc"}
+    BUILD_CACHE_DIR=${BUILD_CACHE_DIR:-/glade/work/epicufsrt/contrib/spack-stack/derecho/build_cache}
+    ;;
+  acorn)
+    module load gcc/11.2.0 python/3.11.7
```
climbfuji: I think we should move those to separate preamble scripts. Not sure where to store those best, but there needs to be a structure like sites/tierN/platform/compiler since these preamble scripts are platform and compiler specific (at least for cray).

util/weekly_build/ShellSetup.sh
```diff
@@ -0,0 +1,101 @@
...
+  echo "Your run failed in $1. This is a placeholder alerting function. 'alert_cmd' should be defined for each system."
+}
+
+function spack_install_exe {
+  spack $* | tee -a log.install 2>&1
+}
+
+case $PLATFORM in
+  hercules)
+    COMPILERS=${COMPILERS:-"intel gcc"}
```
climbfuji: We need to document all environment variables that are used by these scripts.

util/weekly_build/README
```diff
@@ -0,0 +1 @@
+Concretization for deployments should be done with `spack concretize --fresh` so as to ensure that the stack would concretize the same way if the build cache weren't there.
```
climbfuji, line 1: Wondering if we should use something else than weekly, because I plan to do this nightly

util/weekly_build/4_SpackInstall.sh
```diff
@@ -0,0 +1,27 @@
...
+if [ -z $SETUPDONE ]; then . ShellSetup.sh $* ; fi
+
+cd $RUNDIR/${RUNID}
+
+set +x
+. setup.sh
+set -x
+
+INSTALL_OPTS="--show-log-on-error --fail-fast --no-cache $INSTALL_OPTS"
```
climbfuji: I would like to have an option to utilize the existing build cache (for faster rebuilds) or ignore it

util/weekly_build/2_GetSpackStackDevelop.sh
```diff
@@ -0,0 +1,13 @@
...
+if [ -z $SETUPDONE ]; then . ShellSetup.sh $* ; fi
+
+set -ex
+
+cd $RUNDIR
+if [ -d $RUNID ]; then
+  cd $RUNID
+  git pull
+else
+  git clone --recurse-submodules https://github.com/AlexanderRichert-NOAA/spack-stack -b weekly_build $RUNID
```
climbfuji: This needs to be configurable. On dod hpcmp machines, I need to point to the internal enterprise github (url and branch are different)

AlexanderRichert-NOAA: done

util/weekly_build/4_SpackInstall.sh
AlexanderRichert-NOAA: I added an option to not use the --no-cache flag so it'll go with the normal behavior (let me know if the --no-check-signature option is unneeded/unwanted).

util/weekly_build/README
AlexanderRichert-NOAA: Yeah it would make sense to use a more generic term... "scheduled_builds"? "automated_builds"? "build_cacher"? :shrug:

util/weekly_build/ShellSetup.sh
AlexanderRichert-NOAA: If it's okay I'd lean toward waiting to see how complicated the config for each site gets, then if it's more than a couple lines per site break it out into separate scripts.

util/weekly_build/SpackStackBuildCache_AllSteps.sh
AlexanderRichert-NOAA: Fine with me, I'm curious why though?

util/weekly_build/ShellSetup.sh
climbfuji: See readthedocs for Narwhal - different blocks for each compiler, both of them long

util/weekly_build/SpackStackBuildCache_AllSteps.sh
climbfuji: I see more than 9 of those in the future, and I would prefer to see them ordered correctly in bash etc

util/weekly_build/ShellSetup.sh
AlexanderRichert-NOAA: Done. I didn't break them into tier1/tier2 just to keep the logic simple.

util/weekly_build/4_SpackInstall.sh
climbfuji: That will still need the build-cache be added to the spack config, right?

util/weekly_build/ShellSetup.sh
climbfuji: I think that makes sense, we'll probably only do this for tier1 anyway

util/weekly_build/sites/acorn.sh
```diff
@@ -0,0 +1,25 @@
+module load gcc/11.2.0 python/3.11.7
```
climbfuji, line 1: Yes, I think that's great to split it up this way

util/weekly_build/4_SpackInstall.sh
AlexanderRichert-NOAA: Correct (or alternatively point it to $BUILD_CACHE_DIR as a mirror prior to installation)

climbfuji: Ok, so for my understanding, are you planning to permanently add the build-cache to the acorn site config and then use it or not with the --no-cache switch? If so, we could retrieve the build-cache directory from the spack config (if we agree on calling it local-binary in consistency with local-source for the source mirror, we could simply push to that as well).

AlexanderRichert-NOAA: Yep that's smart, let's do that :) I'll test that out shortly...

AlexanderRichert-NOAA: Would it make sense to have one directory on each system `.../spack-stack/cache/` and have one mirror? I think Spack will create 'build_cache' and '_source_cache' directories beneath that anyway, so I'm wondering if there's any downside to consolidating and calling them all 'local-cache' or 'spack-stack-cache'.

AlexanderRichert-NOAA: I tried it the consolidated way and like it (just pushed update), but I don't mind separate if you prefer.

climbfuji: I think it's ok as long as the location of the consolidated cache for source and binary is configurable and can be different that for the build stage. They may end up on slower nfs filesystems or other types of storage (s3) that are unsuitable for the build stage.

climbfuji:
It looks like I am not getting an error signaled to the Github runner when something goes wrong during spack install ... - I believe this is because `... 2>&1 | tee ...` is used.

AlexanderRichert-NOAA:
> It looks like I am not getting an error signaled to the Github runner when something goes wrong during spack install ... - I believe this is because `... 2>&1 | tee ...` is used.

That's... weird. That's with the default install function I take it? Either way, something we could consider is a separate command after the install step that confirms that everything got installed. It would also make it easier to run just the install step with PBS Pro because we wouldn't have to worry about error code propagation.

climbfuji:
> > It looks like I am not getting an error signaled to the Github runner when something goes wrong during spack install ... - I believe this is because `... 2>&1 | tee ...` is used.
> 
> That's... weird. That's with the default install function I take it? Either way, something we could consider is a separate command after the install step that confirms that everything got installed. It would also make it easier to run just the install step with PBS Pro because we wouldn't have to worry about error code propagation.

Seems to be redundant anyway:
```
function spack_install_exe {
  spack $* | tee -a log.install 2>&1
}
```
and then in 04...:
```
spack_install_exe install $INSTALL_OPTS $PACKAGES_TO_INSTALL 2>&1 | tee log.install
```

AlexanderRichert-NOAA:
Hm I thought I'd already fixed that. I'll remove the one from the 04_ script...

climbfuji:
> Hm I thought I'd already fixed that. I'll remove the one from the 04_ script...

I'd rather remove the function and keep things simple and explicit ...

AlexanderRichert-NOAA:
I wrote it this way to make it customizable by system, in particular to be able to use srun, qsub, or whatever else to install. If we call `spack install` directly in the 04_ script, then we can't do parallel builds, can't run MPI-based unit tests on some systems, and might violate system-specific policies regarding large jobs on login nodes. And on some systems we can't just run the whole workflow in a batch job on account of non-internet connected compute nodes. So I'm open to solutions as long as there's the ability to run srun, the parallel build utility, etc.

climbfuji:
> I wrote it this way to make it customizable by system, in particular to be able to use srun, qsub, or whatever else to install. If we call `spack install` directly in the 04_ script, then we can't do parallel builds, can't run MPI-based unit tests on some systems, and might violate system-specific policies regarding large jobs on login nodes. And on some systems we can't just run the whole workflow in a batch job on account of non-internet connected compute nodes. So I'm open to solutions as long as there's the ability to run srun, the parallel build utility, etc.

How would you overwrite this function in the code base? In the machine-specific file in weekly_build/sites/ ?

Good points ... although I would like us to try keeping things as simple as we can (maybe we can't here).

On another one, `spack fetch --missing` doesn't seem to work/maybe have a bug. `intel-oneapi-mpi` is clearly an external 
below, but spack is trying to fetch it anyway. Do we need this call? I never had to use it. If it's about creating a mirror, all we need is `spack mirror create -a -d /path/to/source/cache` after running the concretization step.
```
==> Concretized base-env%oneapi
 -   hz4kma7  base-env@1.0.0%oneapi@2024.1.2 build_system=bundle arch=linux-rhel8-zen3
 -   uqay63s      ^cmake@3.27.9%oneapi@2024.1.2~doc+ncurses+ownlibs build_system=generic build_type=Release arch=linux-rhel8-zen3
[e]  auporsh          ^gmake@4.2.1%gcc@11.2.1~guile build_system=generic patches=ca60bd9,fe5b60d arch=linux-rhel8-zen3
 -   ezwkt7d          ^ncurses@6.5%oneapi@2024.1.2~symlinks+termlib abi=none build_system=autotools patches=7a351bc arch=linux-rhel8-zen3
 -   mrepbct      ^curl@8.7.1%oneapi@2024.1.2~gssapi~ldap~libidn2~librtmp~libssh~libssh2+nghttp2 build_system=autotools libs=shared,static tls=openssl arch=linux-rhel8-zen3
 -   mtt35a6          ^nghttp2@1.57.0%oneapi@2024.1.2 build_system=autotools arch=linux-rhel8-zen3
[e]  rijnwys              ^diffutils@3.6%oneapi@2024.1.2 build_system=autotools arch=linux-rhel8-zen3
 -   yfv4f76          ^openssl@3.3.0%oneapi@2024.1.2~docs+shared build_system=generic certs=mozilla arch=linux-rhel8-zen3
 -   d42jjjk              ^ca-certificates-mozilla@2023-05-30%oneapi@2024.1.2 build_system=generic arch=linux-rhel8-zen3
 -   btagi6i          ^perl@5.38.0%oneapi@2024.1.2+cpanm+opcode+open+shared+threads build_system=generic patches=714e4d1 arch=linux-rhel8-zen3
 -   uq5rnva              ^berkeley-db@18.1.40%oneapi@2024.1.2+cxx~docs+stl build_system=autotools patches=26090f4,b231fcc arch=linux-rhel8-zen3
[e]  skah22p          ^pkgconf@1.4.2%oneapi@2024.1.2 build_system=autotools arch=linux-rhel8-zen3
[e]  iljymxt      ^git@2.31.1%oneapi@2024.1.2+man+nls+perl+subtree~svn~tcltk build_system=autotools arch=linux-rhel8-zen3
[e]  rb7dnia      ^glibc@2.28%oneapi@2024.1.2 build_system=autotools arch=linux-rhel8-zen3
 -   vgayy22      ^hdf5@1.14.3%oneapi@2024.1.2~cxx+fortran+hl~ipo~java~map+mpi+shared~subfiling~szip+threadsafe+tools api=default build_system=cmake build_type=Release generator=make patches=82088c8 arch=linux-rhel8-zen3
[e]  obrli5i          ^intel-oneapi-mpi@2021.12%oneapi@2024.1.2+envmods~external-libfabric~generic-names~ilp64 build_system=generic arch=linux-rhel8-zen3
[e]  rzjafji      ^intel-oneapi-runtime@2024.1.2%oneapi@2024.1.2 build_system=generic arch=linux-rhel8-zen3
 -   neptjsb      ^nccmp@1.9.0.1%oneapi@2024.1.2~ipo build_system=cmake build_type=Release generator=make arch=linux-rhel8-zen3
 -   petxke6      ^netcdf-c@4.9.2%oneapi@2024.1.2+blosc~byterange+dap~fsync~hdf4~jna~logging+mpi~nczarr_zip+optimize~parallel-netcdf+pic+shared~szip+zstd build_system=autotools patches=0161eb8 arch=linux-rhel8-zen3
 -   5srq3lt          ^bzip2@1.0.8%oneapi@2024.1.2~debug~pic+shared build_system=generic arch=linux-rhel8-zen3
 -   izfwrfa          ^c-blosc@1.21.5%oneapi@2024.1.2+avx2~ipo build_system=cmake build_type=Release generator=make arch=linux-rhel8-zen3
 -   rnczwav              ^lz4@1.9.4%oneapi@2024.1.2+pic build_system=makefile libs=shared,static arch=linux-rhel8-zen3
 -   ckyjvvp              ^snappy@1.1.10%oneapi@2024.1.2~ipo+pic+shared build_system=cmake build_type=Release generator=make arch=linux-rhel8-zen3
 -   6wnytum          ^libxml2@2.10.3%oneapi@2024.1.2+pic~python+shared build_system=autotools arch=linux-rhel8-zen3
 -   jylsau3          ^zstd@1.5.2%oneapi@2024.1.2+pic+programs build_system=makefile compression=none libs=shared,static arch=linux-rhel8-zen3
 -   b3eu7f7      ^netcdf-fortran@4.6.1%oneapi@2024.1.2~doc+pic+shared build_system=autotools arch=linux-rhel8-zen3
 -   h26amq7      ^parallel-netcdf@1.12.3%oneapi@2024.1.2~burstbuffer+cxx+fortran+pic+shared build_system=autotools arch=linux-rhel8-zen3
[e]  2on65s7          ^m4@1.4.18%oneapi@2024.1.2+sigsegv build_system=autotools patches=3877ab5,d66f3af,fc9b616 arch=linux-rhel8-zen3
 -   ityf6ea      ^parallelio@2.6.2%oneapi@2024.1.2+fortran~ipo~logging+mpi~ncint+pnetcdf+shared~timing build_system=cmake build_type=Release generator=make arch=linux-rhel8-zen3
 -   f6myyhi      ^py-pip@23.1.2%oneapi@2024.1.2 build_system=generic arch=linux-rhel8-zen3
 -   z4566jw          ^python-venv@1.0%oneapi@2024.1.2 build_system=generic arch=linux-rhel8-zen3
 -   ees4mz6      ^python@3.11.7%oneapi@2024.1.2+bz2+crypt+ctypes+dbm~debug+libxml2+lzma~nis~optimizations+pic~pyexpat+pythoncmd+readline+shared+sqlite3+ssl~tkinter+uuid+zlib build_system=generic patches=13fa8bf,b0615b2,ebdca64,f2fd060 arch=linux-rhel8-zen3
 -   ycou4wb          ^gdbm@1.23%oneapi@2024.1.2 build_system=autotools arch=linux-rhel8-zen3
[e]  nwslwzj          ^gettext@0.19.8.1%oneapi@2024.1.2+bzip2+curses+git~libunistring+libxml2+pic+shared+tar+xz build_system=autotools patches=9acdb4e arch=linux-rhel8-zen3
 -   icjbv3g          ^libffi@3.4.6%oneapi@2024.1.2 build_system=autotools arch=linux-rhel8-zen3
 -   vnucwch          ^libxcrypt@4.4.35%oneapi@2024.1.2~obsolete_api build_system=autotools patches=4885da3 arch=linux-rhel8-zen3
 -   jixpxbn          ^readline@8.2%oneapi@2024.1.2 build_system=autotools patches=bbf97f1 arch=linux-rhel8-zen3
 -   wbrrrsx          ^sqlite@3.43.2%oneapi@2024.1.2+column_metadata+dynamic_extensions+fts~functions+rtree build_system=autotools arch=linux-rhel8-zen3
 -   5oj72q2          ^util-linux-uuid@2.38.1%oneapi@2024.1.2 build_system=autotools arch=linux-rhel8-zen3
 -   5ug42r6          ^xz@5.4.6%oneapi@2024.1.2~pic build_system=autotools libs=shared,static arch=linux-rhel8-zen3
[e]  zvlyy5a      ^wget@1.19.5%oneapi@2024.1.2~libpsl~pcre~python+zlib build_system=autotools ssl=openssl arch=linux-rhel8-zen3
 -   neoop4r      ^zlib-ng@2.1.6%oneapi@2024.1.2+compat+new_strategies+opt+pic+shared build_system=autotools arch=linux-rhel8-zen3

+ return 0
[build-oneapi] [heinzell@nautilus03 build-oneapi]$ spack fetch --missing
==> Using cached archive: /p/work2/heinzell/spack-stack-nightly/20240925/cache/source_cache/_source-cache/archive/60/609a9b98572a6a5ea477f912cffb973109ed4d0a6a6b3f9e2353d2cdc048708e.tar.gz
==> Using cached archive: /p/work2/heinzell/spack-stack-nightly/20240925/cache/source_cache/_source-cache/archive/05/05bbd2b698e9cfbab477c33aa5e99b4975501835a41b7ca6ca71de03d8849e76.tar.bz2
==> Using cached archive: /p/work2/heinzell/spack-stack-nightly/20240925/cache/source_cache/_source-cache/archive/1e/1e3258453784d3b7e6cc48d0be087b168f8360b5d588c66bfeda05d07ad39ffd.tar.gz
==> Using cached archive: /p/work2/heinzell/spack-stack-nightly/20240925/cache/source_cache/_source-cache/archive/53/53e66b043322a606abf0087e7699a0e033a37fa13feb9742df35c3a33b18fb02.tar.gz
==> Using cached archive: /p/work2/heinzell/spack-stack-nightly/20240925/cache/source_cache/_source-cache/archive/5f/5fadcae90aa4ae041150f8e2d26c37d980522cdb49f923fc1e1b5eb8d74e71ad
==> Using cached archive: /p/work2/heinzell/spack-stack-nightly/20240925/cache/source_cache/_source-cache/archive/21/213ef58089d2f2c972ea353517dc60ec3656f050dcc027666e118b508423e517.tar.gz
==> Using cached archive: /p/work2/heinzell/spack-stack-nightly/20240925/cache/source_cache/_source-cache/archive/9d/9da50e155df72bce55cb69f51f1dbb4b62d23740fb99f6178bb27f22ebdf8a46.tar.gz
==> Using cached archive: /p/work2/heinzell/spack-stack-nightly/20240925/cache/source_cache/_source-cache/archive/0c/0cecb2ef0c67b166de93732769abdeba0555086d51de1090df325e18ee8da9c8.tar.gz
==> Using cached archive: /p/work2/heinzell/spack-stack-nightly/20240925/cache/source_cache/_source-cache/archive/ab/ab5a03176ee106d3f0fa90e381da478ddae405918153cca248e682cd0c4a2269.tar.gz
==> Using cached archive: /p/work2/heinzell/spack-stack-nightly/20240925/cache/source_cache/_source-cache/archive/74/74b1081d21fff13ae4bd7c16e5d6e504a4c26f7cde1dca0d963a484174bbcacd.tar.gz
==> Using cached archive: /p/work2/heinzell/spack-stack-nightly/20240925/cache/source_cache/_source-cache/archive/3f/3feb7171f16a84ee82ca18a36d7b9be109a52c04f492a053331d7d1095007c35.tar.gz
==> Using cached archive: /p/work2/heinzell/spack-stack-nightly/20240925/cache/source_cache/_source-cache/archive/bb/bbf97f1ec40a929edab5aa81998c1e2ef435436c597754916e6a5868f273aff7
==> Using cached archive: /p/work2/heinzell/spack-stack-nightly/20240925/cache/source_cache/_source-cache/archive/13/136d91bc269a9a5785e5f9e980bc76ab57428f604ce3e5a5a90cebc767971cc6.tar.gz
==> Using cached archive: /p/work2/heinzell/spack-stack-nightly/20240925/cache/source_cache/_source-cache/archive/a5/a5d504c0d52e2e2721e7e7d86988dec2e290d723ced2307145dedd06aeb6fef2.tar.gz
==> Using cached archive: /p/work2/heinzell/spack-stack-nightly/20240925/cache/source_cache/_source-cache/archive/09/09cdb287aa7a89148c1638dd20891fdbae08102cf433ef128fd345338aa237c7.tar.gz
==> Error: Unable to parse extension from https://registrationcenter-download.intel.com/akdlm//IRC_NAS/2021.12/l_mpi_oneapi_p_2021.11.0.49513_offline.sh.

If this URL is for a tarball but does not include the file extension
in the name, you can explicitly declare it with the following syntax:

    version('1.2.3', 'hash', extension='tar.gz')

If this URL is for a download like a .jar or .whl that does not need
to be expanded, or an uncompressed installation script, you can tell
Spack not to expand it with the following syntax:

    version('1.2.3', 'hash', expand=False)
```

climbfuji:
```
[build-oneapi] [heinzell@nautilus03 build-oneapi]$ spack mirror create -a -d /p/app/projects/NEPTUNE/spack-stack/source-cache
==> Skipping glibc@2.28%oneapi@2024.1.2 build_system=autotools arch=linux-rhel8-zen3/rb7dnia as it is an external spec.
==> Skipping gmake@4.2.1%gcc@11.2.1~guile build_system=generic patches=ca60bd9,fe5b60d arch=linux-rhel8-zen3/auporsh as it is an external spec.
==> Skipping intel-oneapi-runtime@2024.1.2%oneapi@2024.1.2 build_system=generic arch=linux-rhel8-zen3/rzjafji as it is an external spec.
==> Skipping diffutils@3.6%oneapi@2024.1.2 build_system=autotools arch=linux-rhel8-zen3/rijnwys as it is an external spec.
==> Skipping pkgconf@1.4.2%oneapi@2024.1.2 build_system=autotools arch=linux-rhel8-zen3/skah22p as it is an external spec.
==> Skipping git@2.31.1%oneapi@2024.1.2+man+nls+perl+subtree~svn~tcltk build_system=autotools arch=linux-rhel8-zen3/iljymxt as it is an external spec.
==> Skipping intel-oneapi-mpi@2021.12%oneapi@2024.1.2+envmods~external-libfabric~generic-names~ilp64 build_system=generic arch=linux-rhel8-zen3/obrli5i as it is an external spec.
==> Skipping m4@1.4.18%oneapi@2024.1.2+sigsegv build_system=autotools patches=3877ab5,d66f3af,fc9b616 arch=linux-rhel8-zen3/2on65s7 as it is an external spec.
==> Skipping gettext@0.19.8.1%oneapi@2024.1.2+bzip2+curses+git~libunistring+libxml2+pic+shared+tar+xz build_system=autotools patches=9acdb4e arch=linux-rhel8-zen3/nwslwzj as it is an external spec.
==> Skipping wget@1.19.5%oneapi@2024.1.2~libpsl~pcre~python+zlib build_system=autotools ssl=openssl arch=linux-rhel8-zen3/zvlyy5a as it is an external spec.
==> Adding package base-env@1.0.0 to mirror
==> Adding package cmake@3.27.9 to mirror
==> Adding package curl@8.7.1 to mirror
==> Adding package nghttp2@1.57.0 to mirror
==> Adding package openssl@3.3.0 to mirror
==> Adding package ca-certificates-mozilla@2023-05-30 to mirror
==> Adding package perl@5.38.0 to mirror
==> Adding package berkeley-db@18.1.40 to mirror
==> Adding package bzip2@1.0.8 to mirror
==> Adding package gdbm@1.23 to mirror
==> Adding package readline@8.2 to mirror
==> Adding package ncurses@6.5 to mirror
==> Adding package zlib-ng@2.1.6 to mirror
==> Adding package hdf5@1.14.3 to mirror
==> Adding package nccmp@1.9.0.1 to mirror
==> Adding package netcdf-c@4.9.2 to mirror
==> Adding package c-blosc@1.21.5 to mirror
==> Adding package lz4@1.9.4 to mirror
==> Adding package snappy@1.1.10 to mirror
==> Adding package zstd@1.5.2 to mirror
==> Adding package libxml2@2.10.3 to mirror
==> Adding package xz@5.4.6 to mirror
==> Adding package netcdf-fortran@4.6.1 to mirror
==> Adding package parallel-netcdf@1.12.3 to mirror
==> Adding package parallelio@2.6.2 to mirror
==> Adding package py-pip@23.1.2 to mirror
==> Adding package python@3.11.7 to mirror
==> Adding package libffi@3.4.6 to mirror
==> Adding package libxcrypt@4.4.35 to mirror
==> Adding package sqlite@3.43.2 to mirror
==> Adding package util-linux-uuid@2.38.1 to mirror
==> Adding package python-venv@1.0 to mirror
==> Summary for mirror in /p/app/projects/NEPTUNE/spack-stack/source-cache
==> Archive stats:
    30   already present
    0    added
    0    failed to fetch.
```

climbfuji:
Given what I wrote above, I tend to prefer having two separate explicit directories for the spack source and binary caches with pre-defined mirror names in the site config (local-source and local-binary). This way, we can leverage commands like `spack mirror create`, `spack buildcache update` with the names/paths of the mirrors and be done.

AlexanderRichert-NOAA:
> Given what I wrote above, I tend to prefer having two separate explicit directories for the spack source and binary caches with pre-defined mirror names in the site config (local-source and local-binary). This way, we can leverage commands like `spack mirror create`, `spack buildcache update` with the names/paths of the mirrors and be done.

I don't _think_ this approach would stop us from doing that, but I have no strong opinion either. I can revert that commit, or we can tweak it on a site-by-site basis (in which case please feel free to modify my branch directly).

AlexanderRichert-NOAA:
> How would you overwrite this function in the code base? In the machine-specific file in weekly_build/sites/ ?

Correct. See the acorn one as an example; when you define the function it will supersede the default one in ShellSetup.sh.

> Good points ... although I would like us to try keeping things as simple as we can (maybe we can't here).

I think we want and need this flexibility, but again if there's another way to do it I'm down. I agree this approach is a little ugly, but it was the only way I could think of to have flexibility across systems while also having a "standard" way of providing default and custom install arguments.

> On another one, `spack fetch --missing` doesn't seem to work/maybe have a bug. `intel-oneapi-mpi` is clearly an external below, but spack is trying to fetch it anyway. Do we need this call? I never had to use it. If it's about creating a mirror, all we need is `spack mirror create -a -d /path/to/source/cache` after running the concretization step.

That's super weird. `spack fetch` doesn't create a mirror per se, it just populates the source cache by downloading everything. This is needed on systems like Acorn where we'll build on non-internet connected compute nodes, but these scripts, including the fetch command, will run on an internet-connected node. It must be a bug... Does removing `--missing` make any difference? I don't think that option is strictly needed.

climbfuji:
> > How would you overwrite this function in the code base? In the machine-specific file in weekly_build/sites/ ?
> 
> Correct. See the acorn one as an example; when you define the function it will supersede the default one in ShellSetup.sh.
> 
> > Good points ... although I would like us to try keeping things as simple as we can (maybe we can't here).
> 
> I think we want and need this flexibility, but again if there's another way to do it I'm down. I agree this approach is a little ugly, but it was the only way I could think of to have flexibility across systems while also having a "standard" way of providing default and custom install arguments.
> 
> > On another one, `spack fetch --missing` doesn't seem to work/maybe have a bug. `intel-oneapi-mpi` is clearly an external below, but spack is trying to fetch it anyway. Do we need this call? I never had to use it. If it's about creating a mirror, all we need is `spack mirror create -a -d /path/to/source/cache` after running the concretization step.
> 
> That's super weird. `spack fetch` doesn't create a mirror per se, it just populates the source cache by downloading everything. This is needed on systems like Acorn where we'll build on non-internet connected compute nodes, but these scripts, including the fetch command, will run on an internet-connected node. It must be a bug... Does removing `--missing` make any difference? I don't think that option is strictly needed.

But we do have mirrors configured for each site and I don't understand why we don't use/update them through the weekly/nightly actions? There's no need for `spack fetch` if we use the mirrors and run a `spack mirror create -a -d` after the concretization step. This way it's always up to date and contains all the source codes needed to build the environments.

AlexanderRichert-NOAA:
I think functionally that will do the same thing, or at least, the same thing as what I want, so yeah that sounds good. I'll test it out and assuming it works I'll push the change.

AlexanderRichert-NOAA:
I think both `spack mirror create` (by default) and `spack fetch` (always) will fetch packages to whatever path is configured by `config:source_cache`. With `spack mirror create` one can specify a directory or mirror name, but in that case I don't seem to be able to give it a list of packages (which I need for Acorn), but then `spack fetch` obviously has a bug that you're running up against... Do you foresee any problem with using `spack mirror create --dependents ${PACKAGES_TO_INSTALL:---all}`? i.e., it'll default to `--all` if no packages are specified (which will probably be the case on most systems), and the packages will go into the path given by `config:source_cache`.

climbfuji:
> I think both `spack mirror create` (by default) and `spack fetch` (always) will fetch packages to whatever path is configured by `config:source_cache`. With `spack mirror create` one can specify a directory or mirror name, but in that case I don't seem to be able to give it a list of packages (which I need for Acorn), but then `spack fetch` obviously has a bug that you're running up against... Do you foresee any problem with using `spack mirror create --dependents ${PACKAGES_TO_INSTALL:---all}`? i.e., it'll default to `--all` if no packages are specified (which will probably be the case on most systems), and the packages will go into the path given by `config:source_cache`.

This could work, yes. I can try that out on Nautilus. Do you know if `spack clean -a` clears the source cache? I think it does from what I read. If so, then that's a big no for using source_cache.

AlexanderRichert-NOAA:
I believe so, specifically the -d. Could you use `spack clean -sfmp`?

climbfuji:
> I believe so, specifically the -d. Could you use `spack clean -sfmp`?

I am sorry, but I am going to push back here. `spack clean -a` is enshrined in everybody's memory, not just mine, because that's what they've been told for years.

I don't see the benefits of the proposed solution with cache directories that are somewhat "internal" to spack rather than making them explicit as mirrors (and I believe this is what e4s and others are doing).

There needs to be a clear separation also from an access perspective - the mirrors are maintained by the role account / spack admin user of that spack installation, the spack-internal caches are owned by whoever uses spack. Users who want to run their own spack installs and who are not spack-stack maintainers should be able to check out the site config and point to the existing source and binary mirrors, and they can't do that if the spack top-level cache directory points there (where would spack download an additional package to?).

Conversely, the nightly/weekly tests should be able to update those source and binary mirrors. Unless I am missing something entirely, of course!

AlexanderRichert-NOAA:
Okay I totally misunderstood the issue with `spack clean -a`, yes I agree that's a non-starter so nevermind that :upside_down_face: I agree with `config:source_cache:$SPACK_STACK_DIR` and `mirrors:local-whatever:...` for caches deployed by the spack-stack maintainers.
I'll see what the Spack folks say about specifying packages _and_ a mirror with `spack mirror create`, but worst case scenario I can just download everything every time including the stuff that I don't plan to install, or use a trimmed down template.

climbfuji:
Thanks very much, and sorry for not being clear about this. I'll also ask the spack maintainers about the `spack fetch --missing` issue.

climbfuji:
FWIW, I tried `spack fetch` without `--missing` and it gave the same error (tried to download intel-oneapi-mpi` even though it is an external package and was concretized as such).

climbfuji:
@AlexanderRichert-NOAA I use this in my testing in the 03_...  script:
```
  # Get path to local-source mirror from spack site config
  SOURCE_MIRROR=`spack mirror list | grep local-source`
  SOURCE_MIRROR=${SOURCE_MIRROR/local-source \[sb\] file:\/\//}
  # Download all source codes needed to build the environment
  spack mirror create -a -d ${SOURCE_MIRROR}
```
with output
```
+ SOURCE_MIRROR='local-source [sb] file:///p/app/projects/NEPTUNE/spack-stack/source-cache'
+ SOURCE_MIRROR=/p/app/projects/NEPTUNE/spack-stack/source-cache
+ spack mirror create -a -d /p/app/projects/NEPTUNE/spack-stack/source-cache
```

AlexanderRichert-NOAA:
@climbfuji I think we can just use the mirror name rather than a path? But if not, or if using the path as you're doing generalizes better, then let's do it that way for everything (maybe with the ability to override it for testing/debugging purposes).

climbfuji:
> @climbfuji I think we can just use the mirror name rather than a path? But if not, or if using the path as you're doing generalizes better, then let's do it that way for everything (maybe with the ability to override it for testing/debugging purposes).

Unfortunately, the mirror name doesn't work as far as I know - definitely a missing feature.

AlexanderRichert-NOAA:
That's weird. I'll try it again but otherwise I'll plan to do something along the lines of what you have there.

AlexanderRichert-NOAA:
@climbfuji I've updated the source cache logic to use `spack mirror create`, where both the mirror name/path and the packages can be customized, but by default it will run `spack mirror list`, grab the path for 'local-source', and use that as the directory argument for the mirror create command.

I also reverted the site configs back to using 'local-source', anticipating we'll add adding entries for 'local-binary'. It seems like the site mirrors.yaml's can maybe be trimmed down-- please that a look at the updated acorn/mirrors.yaml and let me know if there's any downside you can think of to the more condensed configuration (i.e., not having the separate fetch and push entries for each mirror).

configs/sites/tier1/acorn/mirrors.yaml
```diff
@@ -0,0 +1,3 @@
+mirrors:
+  local-source: file:///lfs/h1/emc/nceplibs/noscrub/spack-stack/cache
```
climbfuji, line 2: Yes, this looks good - as long as it works ;-)

util/weekly_build/05_BuildCache.sh
```diff
@@ -0,0 +1,18 @@
...
+
+set +x
+. setup.sh
+set -x
+
+for compiler in $COMPILERS; do
+  cd $RUNDIR/$RUNID/envs/build-${compiler/@/-}
+  spack env activate .
+  spack buildcache push --unsigned --force ${BUILD_CACHE} $PACKAGES_TO_INSTALL
+  spack buildcache rebuild-index ${BUILD_CACHE}
```
BruceKropp-Raytheon: Encountering an error here on Orion-intel

```
+ command spack buildcache push --unsigned --force /work/noaa/epic/role-epic/jenkins/workspace/spack-stack/build_auto/Orion-intel/build_cache
+ return 0
+ spack buildcache rebuild-index /work/noaa/epic/role-epic/jenkins/workspace/spack-stack/build_auto/Orion-intel/build_cache
+ : this is a shell function from: /work/noaa/epic/role-epic/jenkins/workspace/spack-stack/build_auto/Orion-intel/weekly/spack/share/spack/setup-env.sh
+ : the real spack script is here: /work/noaa/epic/role-epic/jenkins/workspace/spack-stack/build_auto/Orion-intel/weekly/spack/bin/spack
+ _spack_shell_wrapper buildcache rebuild-index /work/noaa/epic/role-epic/jenkins/workspace/spack-stack/build_auto/Orion-intel/build_cache
+ for var in DYLD_LIBRARY_PATH DYLD_FALLBACK_LIBRARY_PATH
+ eval 'if [ -n "${DYLD_LIBRARY_PATH-}" ]; then export SPACK_DYLD_LIBRARY_PATH=${DYLD_LIBRARY_PATH}; fi'
++ '[' -n '' ']'
+ for var in DYLD_LIBRARY_PATH DYLD_FALLBACK_LIBRARY_PATH
+ eval 'if [ -n "${DYLD_FALLBACK_LIBRARY_PATH-}" ]; then export SPACK_DYLD_FALLBACK_LIBRARY_PATH=${DYLD_FALLBACK_LIBRARY_PATH}; fi'
++ '[' -n '' ']'
+ '[' -n '' ']'
+ _sp_flags=
+ '[' '!' -z x ']'
+ '[' buildcache '!=' buildcache ']'
+ '[' -n '' ']'
+ '[' '' '!=' '' ']'
+ _sp_subcommand=
+ '[' '!' -z x ']'
+ _sp_subcommand=buildcache
+ shift
+ case $_sp_subcommand in
+ command spack buildcache rebuild-index /work/noaa/epic/role-epic/jenkins/workspace/spack-stack/build_auto/Orion-intel/build_cache
==> Warning: Encountered problem listing packages at file:///work/noaa/epic/role-epic/jenkins/workspace/spack-stack/build_auto/Orion-intel/build_cache/build_cache: [Errno 2] No such file or directory: '/work/noaa/epic/role-epic/jenkins/workspace/spack-stack/build_auto/Orion-intel/build_cache/build_cache'
==> Error: Unable to generate package index: Failed to get list of specs from file:///work/noaa/epic/role-epic/jenkins/workspace/spack-stack/build_auto/Orion-intel/build_cache/build_cache
++ ERROR
++ alert_cmd ./05_BuildCache.sh
++ echo 'Your run failed in ./05_BuildCache.sh. This is a placeholder alerting function. '\''alert_cmd'\'' should be defined for each system.'
Your run failed in ./05_BuildCache.sh. This is a placeholder alerting function. 'alert_cmd' should be defined for each system.
++ exit 1
```

AlexanderRichert-NOAA: It appears that the build cache is empty because nothing got installed... That's odd that it didn't fail on the installation step. Looking at envs/build-intel/log.install, it seems it tried to install gmake but then doesn't show any further output. Are you setting $PACKAGES_TO_INSTALL?

BruceKropp-Raytheon: Trying again with:
```
export BUILD_CACHE=${workspace}
mkdir -p ${BUILD_CACHE}/build_cache
```

climbfuji:
@AlexanderRichert-NOAA I am pulling in the latest updates from develop once more

rickgrubin-tomorrow:
Within a `Jenkins` pipeline:

Re-running everything after the recent merge, now I'm seeing a dreaded `Install tree root is too long` error:

```
[...]
17:10:48 ==> No patches needed for gmake
17:10:48 ==> gmake: Executing phase: 'install'
17:11:11 ==> Error: SbangPathError: Install tree root is too long. Spack cannot patch shebang lines when script path length (128) exceeds limit (127).
17:11:11   /work/noaa/epic/role-epic/jenkins/workspace/spack-stack/ufswm-test/Orion-intel/weekly/spack/user_cache/bootstrap/store/bin/sbang
[...]
17:11:11 ==> Warning: Skipping build of clingo-bootstrap-spack-wdkesgo7ubjc5jn7l2ee5s7awdugr6nr since gmake-4.4.1-e4ee4hnuu22kdvldnoenit3gsjbjaldl failed
17:11:11 ==> Warning: Skipping build of re2c-2.2-i4mprqgx2bz2zmsz3e3imasiyv5aycpc since gmake-4.4.1-e4ee4hnuu22kdvldnoenit3gsjbjaldl failed
17:11:11 ==> Error: cannot bootstrap the "clingo" Python module from spec "clingo-bootstrap@spack+python %gcc target=x86_64" due to the following failures:
17:11:11 github-actions-v0.5 raised SbangPathError: Install tree root is too long. Spack cannot patch shebang lines when script path length (128) exceeds limit (127).
[...]
```

Ironically, the build of `gmake-4.4.1` succeeded (per its build log).

Not clear on what I'm doing wrong / how to correct.

Will try taking this out of `Jenkins` and running in a shell (emulating a `cron` job).

climbfuji:
@rickgrubin-tomorrow The error message is right there:
```
github-actions-v0.5 raised SbangPathError: Install tree root is too long. Spack cannot patch shebang lines when script path length (128) exceeds limit (127).
```
You need to use a shorter path.

climbfuji (approved):
I am fine with merging this into develop in its current stage, it works perfectly for my github action runs on nautilus. Maybe we should all just get to the step where the environments build and then populate the cache. We can make adjustments and add more capabilities in subsequent PRs, for example for downstream application testing and package unit testing.

AlexanderRichert-NOAA:
If I can ever manage to log onto Acorn again I want to do another end-to-end test on there... But after that yes that's fine with me.

rickgrubin-tomorrow:
> @rickgrubin-tomorrow The error message is right there:
> 
> ```
> github-actions-v0.5 raised SbangPathError: Install tree root is too long. Spack cannot patch shebang lines when script path length (128) exceeds limit (127).
> ```
> 
> You need to use a shorter path.

I saw the message, and it's meaning was / is clear.

Curious as to why I didn't see this prior, and the only thing that changed was the merge. That said, I imagine that's not the culprit.

Thanks.

I'll have to get the pipleline to shorten:

```
SPACK_ENV=/work/noaa/epic/role-epic/jenkins/workspace/spack-stack/ufswm-test/Orion-intel/weekly/envs/build-unified-dev-intel

echo $SPACK_ENV | wc -c
     115
```

AlexanderRichert-NOAA:
I re-tested on Acorn after the last batch of updates and it was smooth sailing, so @climbfuji if you're still down to merge this sooner and build it out further over time, I'm fine with that.

climbfuji:
> I re-tested on Acorn after the last batch of updates and it was smooth sailing, so @climbfuji if you're still down to merge this sooner and build it out further over time, I'm fine with that.

Let's merge it after the spack dev update if that's ok

AlexanderRichert-NOAA:
spackautodoco

AlexanderRichert-NOAA:
@climbfuji can you check that the contents of .github/workflows/ look correct? Was deleting the ubuntu containers CI one of your modifications?

climbfuji:
> @climbfuji can you check that the contents of .github/workflows/ look correct? Was deleting the ubuntu containers CI one of your modifications?

Oh no - that must have slipped through. Can you restore the file `.github/workflows/ubuntu-ci-containers-x86_64.yaml` please?

configs/templates/neptune-dev/spack.yaml
```diff
@@ -9,7 +9,7 @@ spack:
   - compilers: ['%aocc', '%apple-clang', '%gcc', '%intel', '%oneapi']
   - packages:
       - neptune-env +espc        ^esmf@8.7.0
-      - neptune-python-env +xnrl ^esmf@8.7.0
+      #- neptune-python-env +xnrl ^esmf@8.7.0
```
climbfuji:
```suggestion
      - neptune-python-env +xnrl ^esmf@8.7.0
```

climbfuji: Found one more change that needs to be reverted, sorry

util/weekly_build/README
climbfuji: We can figure this out later, let's leave it as is for now.

AlexanderRichert-NOAA:
I just noticed that on Acorn, it was creating a directory called "file:" in $SPACK_ENV because of the way it handles the mirror path. I'm tweaking it to drop the "file://" part when it looks up the directory; @rickgrubin-tomorrow @climbfuji please re-test with this change and make sure I didn't break anything...

climbfuji:
> I just noticed that on Acorn, it was creating a directory called "file:" in $SPACK_ENV because of the way it handles the mirror path. I'm tweaking it to drop the "file://" part when it looks up the directory; @rickgrubin-tomorrow @climbfuji please re-test with this change and make sure I didn't break anything...

Yes, that's fine. I never used `file://` in the `spack mirror create` command.

climbfuji:
I just pushed develop into this branch. Once we have approval from EPIC (Rick and or Bruce), we can merge this.

AlexanderRichert-NOAA:
I need to make one or two last tweaks to the UWM test scripting then it'll be good to go on my end.

AlexanderRichert-NOAA:
@rickgrubin-noaa I made a few small changes, mainly a couple fixes, plus I'm putting ufs-weather-model under the env directory. This works on Acorn, so if this looks good to you and I didn't break anything for Orion, I'd like to push it as is then later on refine things like having it modify ufs-weather-model modulefiles so we don't need a fork and that sort of thing. Thanks again for adding the UWM testing-- once I added in the shell setup call because of using cron, it ran smoothly.

rickgrubin-noaa:
> @rickgrubin-noaa I made a few small changes, mainly a couple fixes, plus I'm putting ufs-weather-model under the env directory. This works on Acorn, so if this looks good to you and I didn't break anything for Orion, I'd like to push it as is then later on refine things like having it modify ufs-weather-model modulefiles so we don't need a fork and that sort of thing. Thanks again for adding the UWM testing-- once I added in the shell setup call because of using cron, it ran smoothly.

All changes are fine with me; nothing breaks on **orion**. I wasn't testing from cron, thus didn't notice the `shell setup` call missing, sorry.

I had considered modifying `ufs-weather-model` files on the fly (as it's a couple of `sed` statements that could be defined within a function in `sites/<site>.sh`) but had already implemented the _branch of `UFS WM`_ approach.

Please merge, thank you.

PR #1238, opened 2024-08-12
"Revert" Intel deprecation flag changes in spack's lib/spack/env/cc and pull in improved solution from spack mainline

climbfuji:
### Summary

This PR resolves https://github.com/JCSDA/spack-stack/issues/1230. Detailed information on what this PR does and why it is necessary is given in the issue. The improved solution does away with having to specify the `diag-disable` flags, since these are now set implicitly (and only during spack environment builds, not in downstream applications). Credits to Peter Scheibel from the spack core team for suggesting it and adding the spack unit tests to my implementation.

### Testing

- [x] CI (with `oneapi` using `icx`, `icpx`, `ifort` and with `intel` using `icc`, `icpc`, `ifort`)
- [x] Manual testing on Dom's development machine (blackpearl; with `oneapi@2024.1.2` using `icx`, `icpx`, `ifort`)

### Applications affected

None

### Systems affected

All using recent Intel compilers

### Dependencies

- [x]  waiting on https://github.com/JCSDA/spack/pull/460

### Issue(s) addressed

Resolves https://github.com/JCSDA/spack-stack/issues/1230

### Checklist
- [x] This PR addresses one issue/problem/enhancement, or has a very good reason for not doing so.
- [x] These changes have been tested on the affected systems and applications.
- [x] All dependency PRs/issues have been resolved and this PR can be merged.

PR #1244, opened 2024-08-15
Fix Hera Intel MPI config

AlexanderRichert-NOAA:
### Summary

The `prefix: /apps/oneapi` config for intel-oneapi-mpi for Hera got lost in the Intel MKL PR. This PR brings it back. Without it: `==> Error: RuntimeError: /apps/oneapi/mpi/2021.5.1/mpi/2021.5.1/include does not exist`

Also, flex doesn't build with external gettext, so this PR removes external gettext and flex for Hera.

### Testing

Tested on Hera.

### Applications affected

all

### Systems affected

Hera

### Dependencies

none

### Issue(s) addressed

none

### Checklist

### DIFF
--- a/configs/sites/tier1/hera/packages.yaml
+++ b/configs/sites/tier1/hera/packages.yaml
@@ -49,17 +49,9 @@ packages:
     externals:
     - spec: findutils@4.5.11
       prefix: /usr
-  flex:
-    externals:
-    - spec: flex@2.5.37+lex
-      prefix: /usr
   gawk:
     externals:
     - spec: gawk@4.0.2
-      prefix: /usr
-  gettext:
-    externals:
-    - spec: gettext@0.19.8.1
       prefix: /usr
   ghostscript:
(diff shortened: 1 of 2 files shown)


climbfuji (approved):
Oops. Thanks for fixing this. May be the same for Jet.

I am confused by the spack logic. According to spack, if prefix is present the module is ignored, correct? I was hoping the module was configured correctly.

AlexanderRichert-NOAA:
I'm not sure. I'm guessing it's a bug in Spack in terms of how it derives the prefix when using the module :shrug:

climbfuji:
Normally I would suggest force-merge, but since https://github.com/JCSDA/spack-stack/pull/1238 is running the CI tests for auto-merge and these take a bit, I'd like to wait. I can force merge after 1238 is in.

climbfuji:
oops I just hit merge, but I got your flex and gettext!

PR #1247, opened 2024-08-16
global-workflow-env: py-wxflow when +python

AlexanderRichert-NOAA:
### Summary

This PR makes global-workflow-env's dependency on py-wxflow dependent on +python.

### Testing

none

### Applications affected

Global Workflow

### Systems affected

all

### Dependencies

none

### Issue(s) addressed

none

### Checklist


climbfuji (approved):
That's fine, wondering if the dependency should go into ufs-pyenv in the future instead of ufs-weather-model-env.

PR #1258, opened 2024-08-20
Add py-fortranformat to neptune-env

### LINKED ISSUE #1229: [INSTALL]: fortranformat Python package
### Package name

fortranformat (py-fortranformat)

### Package version/tag

tbd

### Build options

tbd

### Installation timeframe

spack-stack-1.8.0

### Other information

See https://github.com/JCSDA-internal/pyiri-jedi/issues/22 if you have access. In short, this package is required for pyiri-jedi (high altitude physics module for NEPTUNE with JEDI DA).

eap:
Add the package py-fortranformat to neptune-env. Note that neptune-env as loaded by the unified environment template is the ~python variant so this package had to be modified manually.

### Testing

This was tested un a Ubuntu system. Testing is ongoing and I'll update this PR from a draft to a full PR when testing is done.

### Issue(s) addressed

Fixes #1229
Note the container update will be made in a separate PR.

### Checklist

### DIFF
--- a/spack-ext/repos/spack-stack/packages/neptune-env/package.py
+++ b/spack-ext/repos/spack-stack/packages/neptune-env/package.py
@@ -55,5 +55,6 @@ class NeptuneEnv(BundlePackage):
         depends_on("py-xarray", type="run")
         depends_on("py-xnrl", type="run")
         depends_on("py-pytest", type="run")
+        depends_on("py-fortranformat", type="run")
 
     # There is no need for install() since there is no code.


climbfuji (commented):
This looks ok to me. Just want to confirm that the python variant in the unified environment stays off for now (because of version conflicts that we've been getting on several systems).

eap:
@climbfuji - Yes I confirmed in my build that the python variant is not built unless I manually edit the spack yaml before running concretize and install. That means that even after this change, fortranformat will be unavailble on any platform using our unified-dev environment.

How are we going to service users dependent on this package? Are there platforms where we build the `+python` variant neptune environment? Alternatively, should we be putting this into a different meta package?

Adding @huishao-r for her thoughts on this ↑

climbfuji:
> @climbfuji - Yes I confirmed in my build that the python variant is not built unless I manually edit the spack yaml before running concretize and install. That means that even after this change, fortranformat will be unavailble on any platform using our unified-dev environment.
> 
> How are we going to service users dependent on this package? Are there platforms where we build the `+python` varient neptune environment? Alternatively, should we be putting this into a different meta package?
> 
> Adding @huishao-r for her thoughts on this ↑

We can't build the Python variant of neptune-env as is in the unified environment for two reasons:
1. py-xnrl is a code with restricted access; the code lives in the NRL enterprise github
2. Also, some of the dependencies of py-xnrl listed explicitly in the variant caused version conflicts in the past with the other complex python environments from JEDI and the UFS

We build neptune-env+python as part of the neptune-dev (NEPTUNE standalone) template on Navy machines (Narwhal, Nautilus, Atlantis, Sandy) regularly. Can't build anywhere else because of the xnrl business mentioned above.

Suggested way forward: split the python variant in neptune-env into two parts, one with basic utilities (incl. the new fortranformat) that concretize fine within the unified environment, and one with the more complicated utilities incl xnrl. If this sounds like a good idea, I will take care of it.

eap:
@climbfuji I agree with the approach you mentioned, also that explains why I have variant/version duplicates in my concretize.

```
py-cryptography@38.0.1
py-cryptography@41.0.3
py-pandas@2.0.3%gcc@11.4.0+excel~performance
py-pandas@1.5.3%gcc@11.4.0+excel+performance
py-poetry-core@1.8.1
py-poetry-core@1.7.0
``

climbfuji:
Let's merge this then and let me work on a follow-up PR.

climbfuji (approved):

PR #1265, opened 2024-08-21
Redefine variants in neptune-env virtual package

climbfuji:
### Summary

This PR defines three variants in the `neptune-env` virtual package: `espc`, `python`, `xnrl`. In the unified environment and in the skylab environment, the variants `espc` and `xnrl` are off, the `python` variant is on. This is so that `py-fortranformat` is available in those environments (required by `pyiri-jedi`). In the neptune standalone environment, all three variants are on. Note that the `xnrl` variant requires the `python` variant.

We can debate if we should move some of the Python packages from the `python` variant to the `xnrl` variant, but since these are in the unified environment anyway (because some other package needs them), it doesn't really matter.

### Testing

I concretize the neptune standalone environment and the unified environment on my laptop and verified that there are no duplicate packages or conflicts. I also tested that specifying `+xnrl ~python` throws an error. Further testing will be done in CI (see below).

### Applications affected

There are no changes to the neptune standalone environment, only to the unified environment and the skylab environment (if that is still used).

### Systems affected

n/a

### Dependencies

n/a

### Issue(s) addressed

Need `py-fortranformat` in unified environment.

### Checklist

### DIFF
--- a/spack-ext/repos/spack-stack/packages/neptune-env/package.py
+++ b/spack-ext/repos/spack-stack/packages/neptune-env/package.py
@@ -19,7 +19,9 @@ class NeptuneEnv(BundlePackage):
 
     version("1.4.0")
 
-    variant("python", default=True, description="Build Python libraries")
+    variant("python", default=True, description="Build Python dependencies")
+    variant("espc", default=True, description="Build ESPC dependencies")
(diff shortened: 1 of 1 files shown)


eap (approved):
Thanks for this. Good idea to define this as a variant rather than splitting the package.

PR #1270, opened 2024-08-23
Add github CLI to global-workflow-env

eap:
Add the github-cli to global-workflow-env

### Testing

* Using an unsubmitted comparison shell tool (coming soon) I compared the spack lock files of the unified-dev environment before and after addition of the gh command line tool to the template. No dependences are added or changed (other than `gh`) in order to include this CLI tool
* I built and ran this code on an ubuntu 22.04 host in the unified-dev environment.

### Issue(s) addressed

Fixes: #1105

### Checklist
- [x] This PR addresses one issue/problem/enhancement, or has a very good reason for not doing so.
- [x] These changes have been tested on the affected systems and applications.
- [x] All dependency PRs/issues have been resolved and this PR can be merged.

PR #1273, opened 2024-08-23
CI workflow ensures directories are writable before environment cleanup.

eap:
A change in https://github.com/JCSDA/spack-stack/pull/1270 uncovered a behavior where `go mod` derived package install modules cannot be removed from our scratch directory due to undesirable folder permissions. This has been observed before and golang maintainers marked the [behavior as WAI](https://github.com/golang/go/issues/35615) suggesting that people maintain these directory trees with go-language tools. In our case this is not highly feasible due to spack's intervention in the project structure meaning we must find a workaround.

The workaround here uses `find` to update directory permissions to ensure that write/execute is set on all directories in the tree. This has been tested locally but due to environment carryover we cannot ensure it has the desired effect for go packages in our CI system until it is submitted. (Ie; testing would require re-running an affected workflow and all other pull requests would still be subject to the issue and would break).

## Testing

CI tests here will exercise the code and ensure it doesn't break our existing environment.

## Issues addressed.

Bugfix for regression noted in #1270

### DIFF
--- a/.github/workflows/macos-ci-aarch64.yaml
+++ b/.github/workflows/macos-ci-aarch64.yaml
@@ -20,6 +20,7 @@ jobs:
         run: |
           pwd
           ls -lart
+          find ./* -type d -exec chmod u+xw {} \;
           rm -fr *
 
       - name: checkout
--- a/.github/workflows/ubuntu-ci-x86_64-gnu.yaml
+++ b/.github/workflows/ubuntu-ci-x86_64-gnu.yaml
@@ -20,6 +20,7 @@ jobs:
         run: |
           pwd
           ls -lart
+          find ./* -type d -exec chmod u+xw {} \;
           rm -fr *
 
       - name: checkout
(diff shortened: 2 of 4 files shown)


climbfuji:
Not just directories, but also files. Can you add
```
find ./* -type f -exec chmod u+w {} \;
```
please?

eap:
@climbfuji  - In testing, fixing the directory permission was adequate to allow the files to be deleted even with the read-only `444` permission on files. I also tested the full approach using a local `go mod` build tree and was able to clear the contents.

Works in osX and linux.

```
# Create directory and file both with `-r--r--r--` permission.
$ cd /tmp
$ mkdir not_deletable
$ echo "deletethis" > ./not_deletable/deleteme
$ chmod 444 not_deletable/deleteme
$ chmod 444 ./not_deletable

# Cannot delete this directory or contents.
$ rm -rf ./not_deletable
rm: cannot remove './not_deletable/deleteme': Permission denied

# Change directory only, it can be deleted.
$ chmod u+wx ./not_deletable
$ rm -rf ./not_deletable
```

climbfuji:
Alright, if that's the case then let's go ahead and merge w/o further CI testing. We can always come back if we need to.

climbfuji (approved):

Issue #1278, opened 2024-08-29
spack-stack-1.8.0 release tasks


climbfuji:
Tasks for the spack-stack-1.8.0 release:
- [x] Create release branches
- [x] Install release candidate on a few selected systems for testing. Finalize testing/report back/fix issues by September 6. Suggestion:
    - [x] Hera (@ratkovasic-noaa)
    - [x] Nautilus (@climbfuji): `/p/work2/heinzell/spack-stack-1.8.0-rc1/envs/*` with `*` being one of:
        - ue-intel-2021.5.0
        - ue-gcc-12.2.1
        - ue-aocc-4.0.0 (incomplete, because many packages don't build with `flang`)
        - ue-oneapi-2024.1.2
        - ne-intel-2021.5.0
        - ne-gcc-12.2.1
        - ne-oneapi-2024.1.2
        - ne-aocc-4.0.0 (incomplete, because many packages don't build with `flang`)
    - [x] Discover SCU16 (@srherbener, @ashley314)
    - [x] Acorn (@AlexanderRichert-NOAA)
    - [x] Blackpearl (tier-2; @climbfuji)
        - ue-oneapi@2024.2.0 using `ifx` (incomplete, see #1279; complete with `ifort`)
- [x] Update documentation / prepare changes of documentation to move documentation specific for a specific release to the Wiki: https://github.com/JCSDA/spack-stack/pull/1280
- [x] Roll out release starting September 9, to be completed by September 27. Update site configs and site-specific documentation until September 27.
    - [x] Acorn (EMC)
    - [x] ~~Atlantis (NRL)~~ defer to after system maintenance
    - [x] AWS-Pcluster (JCSDA, @eap)
    - [x] ~~Casper~~ (dropped to tier2)
    - [x] Derecho ~~(EPIC, @ratkovasic-noaa)~~ (NRL, @climbfuji)
        - [x] ue-intel-2021.10.0
        - [x] ue-gcc-12.2.0 (failing)
    - [x] Discover SCU16 (JCSDA, @ashley314)
        - [x] ue-intel-2021.6.0
        - [x] ue-gcc-12.1.0
    - [x] Discover SCU17 (JCSDA, @ashley314)
        - [x] ue-intel-2021.10.0
        - [x] ue-gcc-12.3.0
    - [x] Gaea C5 (EPIC) (@ratkovasic-noaa): /ncrc/proj/epic/spack-stack/spack-stack-1.8.0/envs/
[...]

PR #1285, opened 2024-09-04
Acorn updates for 1.8.0

AlexanderRichert-NOAA:
### Summary

This PR provides updates for the Acorn site config for the 1.8.0 release, which should eventually get merged into develop.

### Testing

Used this config to deploy 1.8.0 on Acorn.

### Applications affected

On Acorn I'm only building a subset of the Unified Env, namely ufs-weather-model-env, global-workflow-env, and a few other things.

### Systems affected

Acorn

### Dependencies

none

### Issue(s) addressed

Part of #1278

### Checklist


configs/sites/tier1/acorn/packages_intel.yaml
```diff
@@ -31,7 +31,10 @@
       - '@1.2.1 ~mkl'
     py-numpy:
       require::
-      - '@:1.23 ^openblas'
+      - '^[virtuals=lapack,blas] openblas'
```
climbfuji, line 34: We should probably use this syntax for all sites when we merge this back into develop?

configs/sites/tier1/acorn/packages_intel.yaml
```diff
@@ -31,7 +31,10 @@
...
+      - '@:1.25'
```
climbfuji, line 35: Can you plot with matplotlib and this version with intel@2022?

configs/sites/tier1/acorn/compilers.yaml
```diff
@@ -6,18 +6,17 @@ compilers:
       cxx: CC
       f77: ftn
       fc: ftn
-    flags: {}
+    flags:
+      cflags: -static-libgcc -static-libstdc++ -Bstatic -lstdc++ -Bdynamic -lm -lpthread
```
climbfuji, line 10: For my curiosity, why is this needed?

configs/sites/tier1/acorn/packages_intel.yaml
DavidHuber-NOAA, line 35: Does this `any_of` add to the `@:1.25` or negate it?

configs/sites/tier1/acorn/packages_intel.yaml
```diff
@@ -41,11 +44,16 @@
       require:
       - any_of: ["@2.27.0"]
         when: "%intel@19.1.3.304"
+        message: "2.25.0 is the last version to use C++11 (as opposed to C++17)"
```
DavidHuber-NOAA: Is 2.25.0 not required now?  Was the C++ standard updated to 17 with this update?

configs/sites/tier1/acorn/packages_intel.yaml
AlexanderRichert-NOAA: Yeah I think this is the best practices syntax

configs/sites/tier1/acorn/packages_intel.yaml
AlexanderRichert-NOAA: It basically makes it more restrictive for Intel 19 (technically both requirements are being applied).

climbfuji: But the opposite is needed. We need it restricted to @:1.23 with any newer Intel, and @:1.24 with the older Intel; for GNU, it can be @:1.25.

configs/sites/tier1/acorn/compilers.yaml
AlexanderRichert-NOAA: There have been issues on Acorn in the past with libstdc++ changing or being inconsistent between login and compute nodes in ways that cause dynamic library loading problems, so basically it's a small risk reducer (and so far everything I've tested compiles fine with it). The 'redundant' -static-libstdc++ and `-Bstatic -lstdc++` are because of some very strange quirk with the Cray wrappers and this was the only way to actually force it to use the static libstdc++ :upside_down_face:

configs/sites/tier1/acorn/packages_intel.yaml
AlexanderRichert-NOAA: It's working with intel@2022 (a.k.a. icc version 2021.6.0) using py-numpy@1.25.2, and intel@19 with py-numpy@1.24.4. So I think I can either leave this line as it is or add a `when: intel@2022`.

AlexanderRichert-NOAA: Yes

configs/sites/tier1/acorn/packages_intel.yaml
AlexanderRichert-NOAA: 2.27.0 is no longer in the recipe, so I'm leaving this as 2.25.0 (which is what's now installed for 1.8.0)

Hang-Lei-NOAA (approved):

configs/sites/tier1/acorn/packages_intel.yaml
climbfuji: Wow. Doesn't work on Hera or S4!

climbfuji (approved):

DavidHuber-NOAA (approved):

PR #1294, opened 2024-09-11
Allow specifying compiler version for 'spack stack create env'

AlexanderRichert-NOAA:
### Summary

This PR borrows the logic from #1213 that allows a compiler version to be specified in `spack stack create env`. It's a couple of tweaks to stack_env.py. This isn't strictly urgent, but I'm hoping to put this in the 1.8.0 release in anticipation of supporting installations by NCO (plus maybe others will find it useful; I know it will be useful for Acorn, and maybe also for NRL?).

### Testing

Tested on Acorn and personal machine.
 - `spack stack create env --compiler gcc` correctly yields `prefer: ['%gcc']` in spack.yaml
 - `spack stack create env --compiler gcc@8.5.0` correctly yields `prefer: ['%gcc@8.5.0']` in spack.yaml
 - `spack stack create env --compiler gcc` with unspecified name correctly yields env directory with name `empty.linux.default.gcc`
 - `spack stack create env --compiler gcc@8.5.0` with unspecified name correctly yields env directory with name `empty.linux.default.gcc-8.5.0`

### Applications affected

n/a

### Systems affected

All

### Dependencies

none

### Issue(s) addressed

none

### Checklist

### DIFF
--- a/spack-ext/lib/jcsda-emc/spack-stack/stack/stack_env.py
+++ b/spack-ext/lib/jcsda-emc/spack-stack/stack/stack_env.py
@@ -100,7 +100,7 @@ class StackEnv(object):
 
         if not self.name:
             # site = self.site if self.site else 'default'
-            self.name = "{}.{}.{}".format(self.template, self.site, self.compiler)
+            self.name = "{}.{}.{}".format(self.template, self.site, self.compiler.replace("@", "-"))
 
     def env_dir(self):
         """env_dir is <dir>/<name>"""
@@ -170,7 +170,7 @@ class StackEnv(object):
         self._copy_or_merge_includes("modules", modules_yaml_path, modules_yaml_modulesys_path, destination)
         # Merge or copy common package config(s)
         packages_yaml_path = os.path.join(common_path, "packages.yaml")
-        packages_compiler_yaml_path = os.path.join(common_path, f"packages_{self.compiler}.yaml")
+        packages_compiler_yaml_path = os.path.join(common_path, f"packages_{self.compiler.split('@')[0]}.yaml")
         destination = os.path.join(env_common_dir, "packages.yaml")
         self._copy_or_merge_includes("packages", packages_yaml_path, packages_compiler_yaml_path, destination)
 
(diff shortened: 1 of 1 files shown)


spack-ext/lib/jcsda-emc/spack-stack/stack/stack_env.py
```diff
@@ -263,11 +263,7 @@ def write(self):
             target_compiler = f"%{self.compiler}"
             for i in range(len(definitions)):
                 if "compilers" in definitions[i]:
-                    j = len(definitions[i]["compilers"])-1
-                    while j>=0:
-                        if not definitions[i]["compilers"][j] == target_compiler:
-                            definitions[i]["compilers"].pop(j)
-                        j -= 1
+                    definitions[i] = {"compilers": [target_compiler]}
```
climbfuji, line 266: This is so much easier; do you remember why we put that complicated logic in place originally (I don't)?

climbfuji (approved):
I am fine with merging this into the release branch. I haven't tested this myself, but I will when I move to the next platform to roll out 1.8.0.

Note that there's no need to go back to existing 1.8.0 installs and redo them. Thanks for making these changes.

spack-ext/lib/jcsda-emc/spack-stack/stack/stack_env.py
AlexanderRichert-NOAA: No I don't remember... Maybe something to do with supporting multiple compilers in an environment..?

climbfuji: Oh right. This is something we need to test before we merge. The Intel or oneAPI environments require some packages to be built with gcc, therefore the `compilers` line in the spack config must contain both compilers. See https://github.com/JCSDA/spack-stack/blob/a8bb4fb77b42364bcf9ca06c91a8e77e3b224ff3/configs/sites/tier1/nautilus/packages_intel.yaml#L3

We need to make sure that this still works (i.e. a spack module [tcl|lmod] refresh followed by spack stack setup-meta-modules still works as expected. I think it does, but better to check.

AlexanderRichert-NOAA: Sure. I think we would expect it to work though, on account of there being only one value for target_compiler (based on self.compiler)? Unless users can specify multiple...

climbfuji: It works. See "test env" step in https://github.com/JCSDA/spack-stack/actions/runs/10816758746/job/30008609091?pr=1294 (look for ecflow)

PR #1296, opened 2024-09-12
Documentation only updates specifically for building spack-stack on MacOS Sonoma

srherbener:
### Summary

These instruction updates are needed for MacOS Sonoma to avoid various pitfalls that were discovered during testing the spack-stack-1.8.0 release candidate.

### Testing

Testing build on both my personal Mac and my work Mac, which are both updated to the latest Sonoma (14.6.1) and Command Line Tools (15.3) versions.

### Applications affected

JEDI, UFS - all unified-dev environments on the Mac

### Systems affected

MacOS

### Dependencies

None

### Issue(s) addressed

None

### Checklist


doc/source/NewSiteConfigs.rst
```diff
@@ -248,7 +258,7 @@ Remember to activate the ``lua`` module environment and have MacTeX in your sear
    PATH="$HOMEBREW_ROOT/opt/curl/bin:$PATH" \
         spack external find --scope system curl
 
-   PATH="$HOMEBREW_ROOT/opt/qt@5/bin:$PATH" \
+   PATH="$HOMEBREW_ROOT/opt/qt5/bin:$PATH" \
```
srherbener, line 261: This change was made so that the documentation rendering can be copy-pasted. homebrew sets up both paths using "qt@5" and "qt5" and the '@' character was interference with the copy-paste.

doc/source/NewSiteConfigs.rst
```diff
@@ -152,22 +152,31 @@ These instructions are meant to be a reference that users can follow to set up t
    brew install libiconv
    # Note - need to pin to version 5
    brew install qt@5
-   # Note - For Sonoma only, install m4 and pkg-config
    brew install m4
+   # Note - For Sonoma only: install pkg-config, diffutils and clingo
    brew install pkg-config
+   brew install diffutils
+   brew install clingo
```
srherbener, line 159: I discovered that the spack-mirror repo for bootstrapping spack only goes up to MacOS Ventura, so when using Sonoma the spack bootstrap process would fail intermittently. This would happen during concretize, and the only bootstrap that is needed for concretize is clingo. Therefore, installing clingo in homebrew fixes this issue.

doc/source/NewSiteConfigs.rst
```diff
@@ -232,7 +241,8 @@ Remember to activate the ``lua`` module environment and have MacTeX in your sear
 
    spack external find --scope system \
        --exclude bison --exclude openssl \
-       --exclude python --exclude gettext
+       --exclude python --exclude gettext \
+       --exclude m4
```
srherbener, line 245: For a reason that I was not able to yet ascertain, if the exclude of m4 here isn't used, this command would find the system m4 (/usr/bin/m4) only (even though the homebrew m4 is in PATH) and then the find command later on that specifically targets m4 (by supplying the homebrew path to the homebrew m4) would also not find the homebrew m4. The issue is that the system m4 with Sonoma does not work with spack, whereas the homebrew m4 does work.

climbfuji (approved):
Please merge whenever you like, no need to wait for the tests

Issue #1342, opened 2024-10-14
Reduce number of modules being generated and loaded


climbfuji:
**Is your feature request related to a problem? Please describe.**

One complaint about spack-stack is that the environments define so many modules that are all loaded as dependencies. Not all of those are necessary (because some are build-time dependencies only but not declared as such; others are run-time dependencies but don't need to be loaded because of `RPATH` or dynamic linking).

**Describe the solution you'd like**

Reduce the number of modules being generated and loaded by extending the exclude list in `config/common/modules_*.yaml`.

**Additional context**

There are other alternatives that we can explore in the future, for example spack views (for all dependencies or the Python dependencies only). These solutions will require changes in the downstream applications (build system, runtime config). Maybe we can get away with excluding more modules and being more selective about what we need to load for a certain application.

AlexanderRichert-NOAA:
`modules:lmod:all:autoload:run`?

climbfuji:
I tried something like this in the past, and it failed horribly. Can't remember the details, probably worth revisiting.

climbfuji:
This isn't a huge issue at the moment, closing.

PR #1343, opened 2024-10-14
Move NEPTUNE Python dependencies into its their own package neptune-python-env, remove legacy environment variables for NEPTUNE

climbfuji:
### Summary

This PR moves the full NEPTUNE Python dependencies into their own package `neptune-python-dev`. Only the basic dependencies on `f90nml` and `python-dateutil` remain in `neptune-env` (needed by NUOPC).

The `espc` variant is turned off by default in `neptune-env`, but the `neptune-dev` template enables it. Likewise, the `xnrl` variant is turned off by default in `neptune-python-env`, but the `neptune-dev` template enables it.

Several legacy environment variables for `hdf5`, `libyaml`, `p4est` are removed from `configs/common/modules_*.yaml`.

### Testing

- [x] Manual testing on @climbfuji's development system blackpearl with `gcc` in neptune_atmos
- [x] CI testing in NEPTUNE on Nautilus with Intel classic, Intel llvm, GNU
- [x] Manual testing on @climbfuji's development system blackpearl with `gcc` in dsapi
- [x] Manual testing on @climbfuji's development system blackpearl with `gcc` in ioapi
- [x] spack-stack CI (see below)

### Applications affected

NEPTUNE, JEDI-NEPTUNE

### Systems affected

None

### Dependencies

n/a

### Issue(s) addressed

Working towards https://github.com/JCSDA/spack-stack/issues/1342 (for NEPTUNE only)

### Checklist

### DIFF
--- a/configs/templates/neptune-dev/spack.yaml
+++ b/configs/templates/neptune-dev/spack.yaml
@@ -8,8 +8,8 @@ spack:
   definitions:
   - compilers: ['%aocc', '%apple-clang', '%gcc', '%intel', '%oneapi']
   - packages:
-      - neptune-env +espc ^esmf@8.7.0b11 snapshot=b11
-      - neptune-python-env +xnrl
+      - neptune-env +espc        ^esmf@8.7.0b11 snapshot=b11
+      - neptune-python-env +xnrl ^esmf@8.7.0b11 snapshot=b11
 
   specs:
     - matrix:
--- a/configs/templates/neptune-dev/spack.yaml
+++ b/configs/templates/neptune-dev/spack.yaml
@@ -8,7 +8,8 @@ spack:
   definitions:
   - compilers: ['%aocc', '%apple-clang', '%gcc', '%intel', '%oneapi']
   - packages:
-      - neptune-env +espc +python +xnrl ^esmf@8.7.0b11 snapshot=b11
+      - neptune-env +espc ^esmf@8.7.0b11 snapshot=b11
+      - neptune-python-env +xnrl
 
   specs:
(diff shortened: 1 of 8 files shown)


configs/templates/neptune-dev/spack.yaml
```diff
@@ -8,7 +8,8 @@ spack:
   definitions:
   - compilers: ['%aocc', '%apple-clang', '%gcc', '%intel', '%oneapi']
   - packages:
-      - neptune-env +espc +python +xnrl ^esmf@8.7.0b11 snapshot=b11
+      - neptune-env +espc        ^esmf@8.7.0b11 snapshot=b11
```
areinecke, line 11: I think 8.7.0 was released, should we use that instead of the snapshot?

areinecke (approved):

configs/templates/neptune-dev/spack.yaml
climbfuji: I'll have to go back and update all the stacks, which is doable. But first I need to get esmf 8.7.0 from spack develop into spack-stack. This will add another several days. Do you want to wait for that? I am fine either way.

climbfuji: Thinking about this a bit more, I would like to have a follow-up PR that removes some of the auto-generated spack-stack modules to trim down the list of loaded modules even more (for both neptune-env and neptune-python-env). We'll hvae to roll out a new version of spack-stack on all platforms for that anyway, therefore I'll defer the official esmf@8.7.0 release to that rollout.

PR #1346, opened 2024-10-14
Remove LaTeX documentation from readthedocs

### LINKED ISSUE #1069: Turn off LaTeX doc generation in sphinx/readthedocs
We really don't need to generate PDFs using LaTeX for the spack-stack documentation - the html rendering from readthedocs is good enough.

Remove the unnecessary parts from `doc/source/conf.py` - this includes some entries that need to be updated every time manually for releases (for no good reason).

**Additional context**
n/a

climbfuji:
### Summary

Remove LaTeX documentation from the readthedocs configuration and fix two formatting errors in `doc/source/NewSiteConfigs.rst`. There are four warnings when I build the documentation manually, but no errors anymore. I checked the rendered HTML documentation, and the output looks ok (not sure what the warnings are about).

### Testing

See readthedocs build below (https://spack-stack--1346.org.readthedocs.build/en/1346/NewSiteConfigs.html)

### Applications affected

none

### Systems affected

none

### Dependencies

none

### Issue(s) addressed

Resolves #1069 

### Checklist

### DIFF
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -1,14 +1,8 @@
 cmake_minimum_required(VERSION 3.18 FATAL_ERROR)
 
-project(spack-stack VERSION 1.3.1)
+project(spack-stack VERSION 1.9.0)
 
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/CMakeModules/Modules")
-
-option(SPHINX_OUTPUT_HTML "Build HTML documentation" ON)
-option(SPHINX_OUTPUT_LATEX "Build LaTeX (PDF) documentation" OFF)
-
-message(STATUS "SPHINX_OUTPUT_HTML: ${SPHINX_OUTPUT_HTML}")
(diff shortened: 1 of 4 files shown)


doc/CMakeLists.txt
```diff
@@ -1,15 +1,9 @@
 cmake_minimum_required(VERSION 3.18 FATAL_ERROR)
 
-project(spack-stack VERSION 1.3.1)
```
AlexanderRichert-NOAA, line 3: Is there a way to get the version automatically? If we don't set the next version anywhere else then nevermind.

AlexanderRichert-NOAA (approved):
:)

doc/CMakeLists.txt
climbfuji: We do, in `doc/src/conf.py` we can use `dev` and `develop` for the short and long version in between releases, and then I set them on the release branch. `CMakeLists.txt` unfortunately doesn't allow that, only real versions.

Would be nice if we could set the version in one place and then inherit it from there. But probably a heavy lift, given that it's only three places.

PR #1352, opened 2024-10-21
Add NCO req's to setup.sh

AlexanderRichert-NOAA:
### Summary

This PR modifies setup.sh to make the installation more self contained by setting `SPACK_DISABLE_LOCAL_CONFIG=1` and placing the bootstrap directory under $SPACK_ROOT.

### Testing

Tested as part of acorn/wcoss2 work.

### Applications affected

n/a

### Systems affected

All

### Dependencies

none

### Issue(s) addressed

none

### Checklist


AlexanderRichert-NOAA:
@RatkoVasic-NOAA see if you think these tweaks might be useful for other sites/situations, but if there's any doubt we can keep it optional.

climbfuji:
This means that spack won't consider `~/.spack` at all?

AlexanderRichert-NOAA:
Yep:
```console
$ rm -rf ~/.spack
$ git clone --recurse-submodules https://github.com/JCSDA/spack-stack -b AlexanderRichert-NOAA-patch-1 spack-stack-test
$ cd spack-stack-test
$ . setup.sh --nco
$ spack bootstrap now
$ spack stack create env --name test --template empty --compiler gcc
$ spack env activate envs/test/
$ spack add gmake
$ spack concretize
$ spack install
$ spack module lmod refresh -y
$ spack stack setup-meta-modules
$ ls ~/.spack
ls: cannot access '/home/user/.spack': No such file or directory
```

climbfuji:
I like this. Maybe ask @srherbener or @ashley314 to test this on macOS? Can always use the existing guard for the `user_cache` to exclude macOS if needed.

RatkoVasic-NOAA:
I thought it was already removed using ~/.spack :-)
As I am concerned, I'd like to see bootstrap directory in $SPACK_ROOT, rather than in ~/.spack It is safer, too.

srherbener (commented):
I ran the test sequence given in this PR's comments on my Mac (Sonoma 14.6.1, CommandLineTools 15.3) and those worked. However, the ~/.spack directory got rebuilt containing some cache files. I think it was the `spack bootstrap now` command that rebuilt the  ~/.spack directory. Not sure why this is happening, but I'm also not overly concerned about it. If this works on the HPCs and doesn't on the Mac, I don't think it's worth the trouble to debug and fix it.

I think it would be good to add "bootstrap" to the spack repo .gitignore config file to help prevent accidentally checking the bootstrap directory into the repo.

AlexanderRichert-NOAA:
In that case if there are no objections, I'll take out the if statement...

climbfuji:
I think this is going to be problematic on macOS, since boostrapping spack there requires manual efforts outside an active environment. Maybe @srherbener can confirm or correct me?

AlexanderRichert-NOAA:
If there's any question we can put the bootstrap step under the `if [ "$(uname)" != "Darwin" ]` block...

climbfuji:
> If there's any question we can put the bootstrap step under the `if [ "$(uname)" != "Darwin" ]` block...

I think that would be safer to do. How about air-gapped systems? Just repeat the bootstrap-mirror process every time?

srherbener:
> I think this is going to be problematic on macOS, since boostrapping spack there requires manual efforts outside an active environment. Maybe @srherbener can confirm or correct me?

The mac has a prerequisite to install clingo which needs to be done before spack bootstrap will work. This is documented here: https://spack-stack.readthedocs.io/en/latest/NewSiteConfigs.html#prerequisites-one-off. Since there are prerequisites for other packages on the Mac is this okay for what you need?

setup.sh
```diff
@@ -7,9 +7,12 @@ echo "Setting environment variable SPACK_STACK_DIR to ${SPACK_STACK_DIR}"
 
 source ${SPACK_STACK_DIR:?}/spack/share/spack/setup-env.sh
 echo "Sourcing spack environment ${SPACK_STACK_DIR}/spack/share/spack/setup-env.sh"
+# Avoid using ~/.spack directory for caching, bootstrap, etc.
 if [ "$(uname)" != "Darwin" ]; then
   export SPACK_USER_CACHE_PATH=$SPACK_ROOT/user_cache
 fi
```
srherbener, line 14: I tried the test from the comment above (https://github.com/JCSDA/spack-stack/pull/1352#issuecomment-2427500838) and it completed on my Mac, but the `spack bootstrap now` command rebuilt my $HOME/.spack directory. I'm guessing that is because of this code here. Wouldn't it be okay to move the `.spack` directory on the Mac, or does that create other issues?

setup.sh
```diff
@@ -7,9 +7,12 @@ echo "Setting environment variable SPACK_STACK_DIR to ${SPACK_STACK_DIR}"
...
+export SPACK_DISABLE_LOCAL_CONFIG=1
+echo "Changing bootstrap path to $(spack bootstrap root '$spack/bootstrap')"
```
srherbener: After running the test (https://github.com/JCSDA/spack-stack/pull/1352#issuecomment-2427500838) I see this in my local clone of the spack repo:
```
[test] MacBook-Pro-5:spack steveherbener$ git status
HEAD detached at b6ef265ec4
Untracked files:
  (use "git add <file>..." to include in what will be committed)
	bootstrap/

nothing added to commit but untracked files present (use "git add" to track)
```

This is reflecting the desired outcome, and I think we should add "bootstrap" to the .gitignore file in the spack repo to help avoid accidentally adding the bootstrap directory to a commit.

AlexanderRichert-NOAA: Any issues after this with loading/concretizing environments? If not then it sounds like we're good to go, and I'm fine with adding bootstrap/ to .gitignore later.

AlexanderRichert-NOAA: Does this mean that the other issue with the location of the bootstrap dir is resolved?

setup.sh
```diff
@@ -7,9 +7,12 @@ echo "Setting environment variable SPACK_STACK_DIR to ${SPACK_STACK_DIR}"
 
 source ${SPACK_STACK_DIR:?}/spack/share/spack/setup-env.sh
 echo "Sourcing spack environment ${SPACK_STACK_DIR}/spack/share/spack/setup-env.sh"
+# Avoid using ~/.spack directory for caching, bootstrap, etc.
 if [ "$(uname)" != "Darwin" ]; then
   export SPACK_USER_CACHE_PATH=$SPACK_ROOT/user_cache
+  echo "Changing bootstrap path to $(spack bootstrap root '$spack/bootstrap')"
 fi
+export SPACK_DISABLE_LOCAL_CONFIG=1
```
climbfuji: You probably want to move this inside the if block?

AlexanderRichert-NOAA:
@climbfuji any idea why the CI's are failing? `spack external find: error: argument --scope: invalid choice: 'system' choose from:` ...

climbfuji:
> @climbfuji any idea why the CI's are failing? `spack external find: error: argument --scope: invalid choice: 'system' choose from:` ...

My guess is that the logic proposed here (DISABLE LOCAL CONFIG) disables the system scope - but I haven't tested that locally

AlexanderRichert-NOAA:
Yes, apparently it disables both system and user config. I'll test a little further but I think we're okay without it.

AlexanderRichert-NOAA:
I tested on personal computer and Acorn, and it works fine, i.e., still not creating ~/.spack after creating/concretizing/solving an environment with a fresh copy of spack-stack.

srherbener (approved):
I think this looks good now. Thanks!

setup.sh
```diff
@@ -7,8 +7,10 @@ echo "Setting environment variable SPACK_STACK_DIR to ${SPACK_STACK_DIR}"
 
 source ${SPACK_STACK_DIR:?}/spack/share/spack/setup-env.sh
 echo "Sourcing spack environment ${SPACK_STACK_DIR}/spack/share/spack/setup-env.sh"
+# Avoid using ~/.spack directory for caching, bootstrap, etc.
 if [ "$(uname)" != "Darwin" ]; then
   export SPACK_USER_CACHE_PATH=$SPACK_ROOT/user_cache
+  echo "Changing bootstrap path to $(spack bootstrap root '$spack/bootstrap')"
```
climbfuji, line 13: This comment isn't correct, is it? The corresponding code was removed.

AlexanderRichert-NOAA: I'm not sure what you mean? The comment could go under the if block, but otherwise I think this is accurate.

climbfuji: As far as I can see, this PR didn't change anything except adding two lines of comments. The bootstrap code still goes to `~/.spack`, only the `SPACK_USER_CACHE` path is set to `$SPACK_ROOT/user_cache` (which was already the case before this PR). Maybe I am not seeing it.

AlexanderRichert-NOAA: The `$(spack bootstrap root '$spack/bootstrap')` call relocates the bootstrap directory to $spack/bootstrap. It's just hidden in the 'echo' line.

climbfuji: Ahh! I really didn't see it :-) Thanks!

Issue #1329, opened 2024-10-07
S4 site config updates for spack-stack-1.9.0


climbfuji:
**Is your feature request related to a problem? Please describe.**

In https://github.com/JCSDA/spack-stack/pull/1328, we installed spack-stack-1.8.0 with intel@2021.10.0.

For spack-stack-1.9.0, we want the following updates:
- Addition of latest oneAPI compiler so that we can test with/switch to icx, icpx, ifx (or ifort)
- Addition of a GNU stack using the configured gcc@13.3.0 (missing: MPI installation for this compiler)
- Use of external qt@5 (in spack-stack-1.8.0, we build it and it works, but that makes for a lot of unnecessary packages to install each time)

**Describe the solution you'd like**

See above

**Additional context**

n/a

climbfuji:
Note. The `qt5` OS packages were just added, thus the next round of updates to the S4 site config should configure the external `qt5` package in S4's `packages.yaml`.

climbfuji:
Since the external `qt@5` was added to the S4 site config, and since we decided to not provide a GNU stack on S4 due to the problems with finding a suitable version for the MPI provider, the only remaining item is to check the version of Intel OneAPI currently used for spack-stack on S4, and if a newer one should be used instead (and is available). @InnocentSouopgui-NOAA

InnocentSouopgui-NOAA:
> Since the external `qt@5` was added to the S4 site config, and since we decided to not provide a GNU stack on S4 due to the problems with finding a suitable version for the MPI provider, the only remaining item is to check the version of Intel OneAPI currently used for spack-stack on S4, and if a newer one should be used instead (and is available). @InnocentSouopgui-NOAA

Intel OneAPI 2024.2 is available on S4.
I will work on that.

climbfuji:
This has all been done

Issue #1276, opened 2024-08-26
matplotlib 3.8.0 is unable to create a basic plot


DavidHuber-NOAA:
**Describe the bug**
The matplotlib python library is unable to generate a basic plot, i.e. https://matplotlib.org/stable/gallery/lines_bars_and_markers/simple_plot.html.  There may be a bug between the version of matplotlib and another python library.  Another user reported the same bug, though it was never resolved: https://github.com/matplotlib/matplotlib/issues/23923.

**To Reproduce**
```console
> module use /scratch1/NCEPDEV/nems/Alexander.Richert/spack-stack-py311-aug24/envs/test/install/modulefiles/Core
> module load stack-intel stack-python py-matplotlib
> wget https://matplotlib.org/3.8.4/_downloads/841352d8ea6065fce570abdf6225ef02/simple_plot.py
> python simple_plot.py
Traceback (most recent call last):
  File "/home/David.Huber/testplot.py", line 15, in <module>
    fig.savefig("test.png")
  File "/scratch1/NCEPDEV/nems/Alexander.Richert/spack-stack-py311-aug24/envs/test/install/intel/2021.5.0/py-matplotlib-3.8.0-fcxyi47/lib/python3.11/site-packages/matplotlib/figure.py", line 3390, in savefig
    self.canvas.print_figure(fname, **kwargs)
  File "/scratch1/NCEPDEV/nems/Alexander.Richert/spack-stack-py311-aug24/envs/test/install/intel/2021.5.0/py-matplotlib-3.8.0-fcxyi47/lib/python3.11/site-packages/matplotlib/backend_bases.py", line 2187, in print_figure
    result = print_method(
             ^^^^^^^^^^^^^
  File "/scratch1/NCEPDEV/nems/Alexander.Richert/spack-stack-py311-aug24/envs/test/install/intel/2021.5.0/py-matplotlib-3.8.0-fcxyi47/lib/python3.11/site-packages/matplotlib/backend_bases.py", line 2043, in <lambda>
    print_method = functools.wraps(meth)(lambda *args, **kwargs: meth(
                                                                 ^^^^^
  File "/scratch1/NCEPDEV/nems/Alexander.Richert/spack-stack-py311-aug24/envs/test/install/intel/2021.5.0/py-matplotlib-3.8.0-fcxyi47/lib/python3.11/site-packages/matplotlib/backends/backend_agg.py", line 497, in print_png
[...]

PR #1384, opened 2024-11-17
Unpin or update many packages (mostly Python) in `configs/common/packages.yaml`, fix S4 site config

### LINKED ISSUE #1065: Update shapely from 1.8.0 to latest version 2.x.y

We should update `py-shapely` to the latest version 2.x.y. It's currently pinned to 1.8.0 because of JEDI-EWOK. I created an issue in the JCSDA-internal skylab repo to update the plotting scripts to work with new versions of the package (https://github.com/JCSDA-internal/skylab/issues/439).

Let's try to target spack-stack-1.8.0 for this update. It will hopefully allow us to use newer versions of dependencies for `py-shapely`, too.

See above

**Additional context**
n/a

climbfuji:
### Summary

In preparation for spack-stack-1.9.0, this PR unpins or updates several packages in `configs/common/packages.yaml` (mostly Python packages). Most notably, `py-shapely` (@ericlingerfelt FYI) and `py-numpy` (@DavidHuber-NOAA FYI) are updated.

The `py-numpy` update may require bug fixes with the Intel classic compiler that @DavidHuber-NOAA worked on and that are currently under review in spack develop (see https://github.com/JCSDA/spack-stack/issues/1276).

Included is an update of the S4 site config, which had several flaws that prevented building and testing this PR.

### Testing

- [x]  CI
- [x] NEPTUNE tests with `neptune-dev` template (@climbfuji)
- [x] @AlexanderRichert-NOAA UFS
- [x] @srherbener to run JEDI/Skylab

### Applications affected

All.

### Systems affected

None directly.

### Dependencies

- [x] waiting on https://github.com/JCSDA/spack/pull/484

### Issue(s) addressed

Resolves https://github.com/JCSDA/spack-stack/issues/1065
Working towards https://github.com/JCSDA/spack-stack/issues/1329

### Checklist


configs/common/packages.yaml
```diff
@@ -143,9 +154,9 @@ packages:
     require: '@2.46.3 ~shared ~f2py'
     variants: '+pflogger'
   # To avoid duplicate packages being built
-  meson:
-    require:
-    - '@1.5.1:'
+  #meson:
```
climbfuji: Note to self. Remove all these commented-out packages once we are sure it all works as expected

climbfuji:
All, the NEPTUNE tests passed. Please start testing with UFS, JEDI, ... we need this PR in as soon as possible for spack-stack 1.9.0 (downstream PRs depend on it). Thanks!

srherbener:
@climbfuji I'll test JEDI/Skylab.

climbfuji:
Thanks very much @srherbener! You should expect problems with the shapely update (@ericlingerfelt may know more), but hopefully nothing else.

srherbener:
I'm getting concretize errors like this on several platforms:
```
[ue-intel] [sherbener@s4-submit ue-intel]$ spack concretize | tee log.concretize
==> Warning: Use of plain text `access_token` in mirror config is deprecated, use environment variables instead (access_token_variable)
==> Error: failed to concretize `jedi-geos-env%intel ^esmf@=8.6.1`, `ufs-srw-app-env%intel ^esmf@=8.6.1`, `global-workflow-env%intel ^esmf@=8.6.1`, `ewok-env%intel~cylc+ecflow`, ..., `gsi-env%intel` for the following reasons:
     1. cannot satisfy a requirement for package 'py-setuptools'.
==> Using cached archive: /data/users/sherbener/projects/spack-stack/cache/source_cache/blobs/sha256/3cc99d42a12e6c34bc187d5146cbdba71ca16d19163fe106f1a4cd3e59b01d2c
==> Using cached archive: /data/users/sherbener/projects/spack-stack/cache/source_cache/blobs/sha256/e8518de25baff7a74bdb42193e6e4b0496e7d0688434c42ce4bdc92fe4293a09
==> Installing "clingo-bootstrap@=spack%gcc@=10.2.1~docs+ipo+optimized+python+static_libstdcpp build_system=cmake build_type=Release generator=make patches=bebb819,ec99431 arch=linux-centos7-x86_64" from a buildcache
```
This particular message is from S4. Does this need another PR to be merged, or an update for the spack submodule commit hash?

Or perhaps pilot error. I did the following (on S4) before attempting to do the spack-stack build:
```
module purge
module load intel/2023.2 miniconda/3.8-s4
```
Is that the issue?

Thanks!

climbfuji:
Definitely no miniconda, those times are long gone

srherbener:
I'm still running into issues, but this time I think it's because without the miniconda I have a python3 that is too old:
```
# Still seeing this error: 1. cannot satisfy a requirement for package 'py-setuptools'.

# Here's the modules I have available:

[sherbener@s4-submit spack-stack]$ module list
No modules loaded
[sherbener@s4-submit spack-stack]$ module av

----------------------------------- /opt/apps/modulefiles/Core ------------------------------------
   gcc/10.2        intel/2023.2-newmpi        miniconda/3.8-s4    (D)    ncl/6.6.2
   gcc/13.3 (D)    intel/2023.2        (D)    miniconda/6.6.2-ncl        rocoto/1.3.7

  Where:
   D:  Default Module


# After running `source setup.sh`

[sherbener@s4-submit spack-stack]$ $SPACK_PYTHON --version
Python 3.6.8
[sherbener@s4-submit spack-stack]$ echo $SPACK_PYTHON
/usr/bin/python3

```
Is there another python module (besides miniconda) that I should be loading?

Thanks!

climbfuji:
There is no need to load any other Python. 3.6 is just fine to "drive" spack and to build the Python we want.

srherbener:
Okay, I'll forge ahead with python 3.6.8.

Note that the documentation here https://spack-stack.readthedocs.io/en/latest/PreConfiguredSites.html#create-local-environment says to make sure you have python 3.8+ available and is the default.

climbfuji:
> Okay, I'll forge ahead with python 3.6.8.
> 
> Note that the documentation here https://spack-stack.readthedocs.io/en/latest/PreConfiguredSites.html#create-local-environment says to make sure you have python 3.8+ available and is the default.

That is from the old days when we forced spack to use an external Python in the environment. We should remove that. All we need is Python 3.6+ to drive spack and to build the environments.

climbfuji:
> Okay, I'll forge ahead with python 3.6.8.
> 
> Note that the documentation here https://spack-stack.readthedocs.io/en/latest/PreConfiguredSites.html#create-local-environment says to make sure you have python 3.8+ available and is the default.

Please let me know if you are still having problems and I will try to hop onto S4 (I should still have access).

srherbener:
Reverting the py-setuptools version back to 63.4.3 got concretize to work. Concretize would not work with py-setuptools@69. I'm not sure what package had an issue with py-setuptools@69, but I would think there is a package config that we are using that doesn't allow for py-setuptools@69.

climbfuji:
> Reverting the py-setuptools version back to 63.4.3 got concretize to work. Concretize would not work with py-setuptools@69. I'm not sure what package had an issue with py-setuptools@69, but I would think there is a package config that we are using that doesn't allow for py-setuptools@69.

Reverting setuptools os a no-go, I think, it will break other environments. Did you try unified-dev or skylab-dev?

srherbener:
I was trying unified-dev.

climbfuji:
> I was trying unified-dev.

Hmm. It concretizes fine in CI on macOS and Ubuntu with all the different compilers.

srherbener:
@stiggy87 tried this PR in an AMI using Ubuntu 24.04 and gcc 12.3. I'm attempting S4 using intel@2021.10.0. In @stiggy87's case version 1.26.4 gets selected for py-numpy, and in my case version 1.23.5 gets selected for numpy. It just so happens that according to the depndencies set up in the py-numpy package.py script, py-numpy@1.23.5 is limited to py-setuptools@63 or older, whereas py-numpy@1.26.4 is okay with py-setuptools@69. I'm guessing a similar thing is happening with the CI tests.

I think this makes sense, but I'm not clear on why the py-numpy versions come out different in the two cases.

srherbener:
Could the different py-numpy versions be caused by the version of python3? In my case it's 3.6.8 and in @stiggy87's case its 3.11.7.

climbfuji:
Your Python is **not** 3.6.8. This is only used to run `spack` commands, not inside the environment. It's actually pretty simple, in the S4 site config, `py-numpy` is hardcoded to `1.23 ^openblas` - please try to change that to `1.26.4 ^openblas`.

climbfuji:
> Your Python is **not** 3.6.8. This is only used to run `spack` commands, not inside the environment. It's actually pretty simple, in the S4 site config, `py-numpy` is hardcoded to `1.23 ^openblas` - please try to change that to `1.26.4 ^openblas`.

Or just remove the `:1.23` part, leaving only `^openblas` (might be the better solution).

srherbener:
@climbfuji what you say makes sense, the concretization should be following what is specified in the  config not what SPACK_PYTHON is set to. 

Setting SPACK_PYTHON to a 3.11.7 version didn't help (as expected)

I'm tying again after removing the ':1.23.5' part of the py-numpy spec in the site/packages.yaml.

srherbener:
Concretize is working now! Thanks @climbfuji

climbfuji:
> Concretize is working now! Thanks @climbfuji

Yay. Hopefully the rest will be less bumpy. I will update all the site configs accordingly.

srherbener:
The install failed on building `antlr`. Here is the compile error:
```
                 from /data/users/sherbener/projects/spack-stack/cache/build_stage/spack-stage-antlr-2.7.7-gauwfia2hci6fltmfweq4d2ygty7o4nk/spack-src/lib/cpp/src/ANTLRUtil.cpp(9):
/usr/include/wchar.h(396): error: identifier "_Float32" is undefined
  extern _Float32 wcstof32 (const wchar_t *__restrict __nptr,
```

I noticed that there seems to be something suspicious with the antlr spec:
```
[ue-intel] [sherbener@s4-submit ue-intel]$ spack spec antlr
==> Warning: Use of plain text `access_token` in mirror config is deprecated, use environment variables instead (access_token_variable)
 -   antlr@2.7.7%intel@2021.10.0+cxx~java~pic~python build_system=autotools patches=33897ad arch=linux-rocky8-skylake_avx512
[e]      ^glibc@2.28%intel@2021.10.0 build_system=autotools arch=linux-rocky8-skylake
[e]      ^gmake@4.2.1%intel@2021.10.0~guile build_system=generic patches=ca60bd9,fe5b60d arch=linux-rocky8-skylake
```
The `arch` entry for the dependencies of antlr do not have the `_avx512` suffix., whereas the entry for `antlr` does have that suffix. I seem to recall a recent issue/PR about the AVX512 instruction set, but I'm not sure which is expected (with or without the `_avx512` suffix). But it doesn't seem right that there is a mix of with/without the suffix - seems like `arch` should be consistent for all the packages.

Any thoughts? Thanks!

climbfuji:
I have no idea, never seen this. The antlr version is the same as it has always been, no updates. Not sure if the `_avx512` has anything to do with it, but you might be able to force all packages to build for `arch=linux-rocky8-skylake` ?

climbfuji:
> I have no idea, never seen this. The antlr version is the same as it has always been, no updates. Not sure if the `_avx512` has anything to do with it, but you might be able to force all packages to build for `arch=linux-rocky8-skylake` ?

FYI, I am trying to build on Derecho (same compiler, intel@2021.10.0) - will let you know if I run into the same issue or not.

srherbener:
With the knowledge about the py-numpy version requirement in the site/packages.yaml, I went back to Orion and attempted to build there. I'm attempting to build with GNU/OpenMPI and getting this error from concretize:
```
[ue-gnu] orion-login-2[20] herbener$ spack concretize 2>&1 | tee log.concretize
==> Warning: Use of plain text `access_token` in mirror config is deprecated, use environment variables instead (access_token_variable)
==> Error: failed to concretize `jedi-ufs-env%gnu ^esmf@=8.6.1`, `gsi-env%gnu`, `ewok-env%gnu~cylc+ecflow`, `jedi-fv3-env%gnu`, ..., `esmf@=8.6.1%gnu snapshot=none` for the following reasons:
     1. Cannot set the required compiler: madis%gnu
```

After some poking around I discovered that madis seems to want to use the intel compiler, which I think led to the concretize error above. Here is what `spack spec madis` returns, in my unified-env, gnu environment.

```
[ue-gnu] orion-login-2[25] herbener$ spack spec madis
==> Warning: Use of plain text `access_token` in mirror config is deprecated, use environment variables instead (access_token_variable)
 -   madis@4.5%intel@2021.9.0+pic~pnetcdf build_system=makefile arch=linux-rocky9-skylake_avx512
[e]      ^glibc@2.34%intel@2021.9.0 build_system=autotools arch=linux-rocky9-skylake_avx512
 -       ^gmake@4.2.1%intel@2021.9.0~guile build_system=generic patches=ca60bd9,fe5b60d arch=linux-rocky9-skylake_avx512
 -       ^netcdf-fortran@4.6.1%intel@2021.9.0~doc+pic+shared build_system=autotools arch=linux-rocky9-skylake_avx512
 -           ^intel-oneapi-mpi@2021.14.0%intel@2021.9.0~classic-names+envmods~external-libfabric~generic-names~ilp64 build_system=generic arch=linux-rocky9-skylake_avx512
 -           ^netcdf-c@4.9.2%intel@2021.9.0~blosc~byterange+dap~fsync~hdf4~jna~logging+mpi~nczarr_zip+optimize~parallel-netcdf+pic~shared~szip~zstd build_system=autotools patches=0161eb8 arch=linux-rocky9-skylake_avx512
[e]              ^curl@7.76.1%intel@2021.9.0+gssapi+ldap~libidn2~librtmp~libssh~libssh2+nghttp2 build_system=autotools libs=shared,static tls=openssl arch=linux-rocky9-skylake_avx512
 -               ^hdf5@1.14.3%intel@2021.9.0~cxx+fortran+hl~ipo~java~map+mpi+shared~subfiling~szip+threadsafe+tools api=default build_system=cmake build_type=Release generator=make patches=82088c8,f42732a arch=linux-rocky9-skylake_avx512
 -                   ^cmake@3.27.9%intel@2021.9.0~doc~ncurses+ownlibs~qtgui build_system=generic build_type=Release patches=dbc3892 arch=linux-rocky9-skylake_avx512
 -                   ^pkg-config@0.29.2%intel@2021.9.0+internal_glib build_system=autotools arch=linux-rocky9-skylake_avx512
 -                   ^zlib-ng@2.2.1%intel@2021.9.0+compat~new_strategies+opt+pic+shared build_system=autotools arch=linux-rocky9-skylake_avx512
 -               ^libxml2@2.13.4%intel@2021.9.0+pic~python+shared build_system=autotools arch=linux-rocky9-skylake_avx512
 -                   ^xz@5.4.6%intel@2021.9.0~pic build_system=autotools libs=shared,static arch=linux-rocky9-skylake_avx512

```

Any ideas why madis wants to use the intel compiler instead of gcc?

Thanks!

climbfuji:
Unfortunately, I do not have the bandwidth to debug all tier-1 site configs. We need to share that work, and yes, sometimes it takes hours or days. Can you try to resolve the Orion issues yourself or with the help of EPIC, please?

As far as S4 is concerned: I think the problem is that the site config is bad/wrong. If you look at compilers.yaml, you see that Intel uses gcc@13 as the GCC backend. We've known for years (and there are several open issues in spack-stack) that Intel classic doesn't do well with a gcc@12 or newer backend. I hopped on S4, added gcc@10.2.0 as the backend for Intel classic, removed the py-numpy variant and also forced arch=linux-rocky8-skylake (the latter is probably not needed, but I don't like that some stuff is concretized with avx512 while some stuff isn't).

The gcc@10 backend change allowed me to compile antlr. I'll try to build the entire stack on S4 with Intel, and if it works I'll update the site config in the PR and point you to the stack for testing.

srherbener:
Sorry @climbfuji about all the questions. I'll spread out the questions elsewhere and I greatly appreciate the help and support that you provide. Unfortunately, I don't have the bandwidth for debugging all of these site configurations either. My real need is to just build somewhere so I can test this PR with skylab.

climbfuji:
> Sorry @climbfuji about all the questions. I'll spread out the questions elsewhere and I greatly appreciate the help and support that you provide. Unfortunately, I don't have the bandwidth for debugging all of these site configurations either. My real need is to just build somewhere so I can test this PR with skylab.

Hopefully S4 will be ready for you by the end of today so that you can test next week!

InnocentSouopgui-NOAA:
> > Sorry @climbfuji about all the questions. I'll spread out the questions elsewhere and I greatly appreciate the help and support that you provide. Unfortunately, I don't have the bandwidth for debugging all of these site configurations either. My real need is to just build somewhere so I can test this PR with skylab.
> 
> Hopefully S4 will be ready for you by the end of today so that you can test next week!

If the readiness of S4 refers to the network problem experienced last week, then it had been resolved since Tuesday. 
I am assisting users with extra libraries there before getting to the task I have on spack-stack for s4.

climbfuji:
> > > Sorry @climbfuji about all the questions. I'll spread out the questions elsewhere and I greatly appreciate the help and support that you provide. Unfortunately, I don't have the bandwidth for debugging all of these site configurations either. My real need is to just build somewhere so I can test this PR with skylab.
> > 
> > 
> > Hopefully S4 will be ready for you by the end of today so that you can test next week!
> 
> If the readiness of S4 refers to the network problem experienced last week, then it had been resolved since Tuesday. I am assisting users with extra libraries there before getting to the task I have on spack-stack for s4.

no, just bugs in the site config

srherbener:
> > Sorry @climbfuji about all the questions. I'll spread out the questions elsewhere and I greatly appreciate the help and support that you provide. Unfortunately, I don't have the bandwidth for debugging all of these site configurations either. My real need is to just build somewhere so I can test this PR with skylab.
> 
> Hopefully S4 will be ready for you by the end of today so that you can test next week!

Thanks @climbfuji - much appreciated!

climbfuji:
> > > Sorry @climbfuji about all the questions. I'll spread out the questions elsewhere and I greatly appreciate the help and support that you provide. Unfortunately, I don't have the bandwidth for debugging all of these site configurations either. My real need is to just build somewhere so I can test this PR with skylab.
> > 
> > 
> > Hopefully S4 will be ready for you by the end of today so that you can test next week!
> 
> Thanks @climbfuji - much appreciated!

@srherbener `/data/users/dheinzeller/spst-unpin-update/envs/ue-intel-2021.10.0/install/modulefiles/Core`

srherbener:
> 
> @srherbener `/data/users/dheinzeller/spst-unpin-update/envs/ue-intel-2021.10.0/install/modulefiles/Core`

Thanks @climbfuji! I am building jedi-bundle now, and will run ctests and skylab.

srherbener:
> > Okay, I'll forge ahead with python 3.6.8.
> > Note that the documentation here https://spack-stack.readthedocs.io/en/latest/PreConfiguredSites.html#create-local-environment says to make sure you have python 3.8+ available and is the default.
> 
> That is from the old days when we forced spack to use an external Python in the environment. We should remove that. All we need is Python 3.6+ to drive spack and to build the environments.

Note, I just submitted a PR (https://github.com/JCSDA/spack-stack/pull/1420) to correct the documentation.

srherbener:
jedi-bundle looks good, I'm running skylab now.

configs/common/packages.yaml
```diff
@@ -294,4 +305,5 @@ packages:
   wrf-io:
     require: '@1.2.0'
   zstd:
-    require: '@1.5.2 +programs'
+    #require: '@1.5.2 +programs'
```
AlexanderRichert-NOAA: Is there a reason not to pin the zstd version? It's not a huge issue, but I would tend to lean toward pinning versions for packages that are used by UWM, in other words I would generally put zstd in the same category as hdf5, netcdf, etc.

configs/sites/tier1/orion/packages_intel.yaml
```diff
@@ -30,4 +30,4 @@ packages:
     - '@1.2.1 ~mkl'
   py-numpy:
     require::
-    - '@:1.23 ^openblas'
+    - '^openblas'
```
AlexanderRichert-NOAA, line 33: If you want to be shmancy you could use an edge label here: `^[virtuals=blas,lapack] openblas`. But functionally I don't think it makes a difference.

Is there a reason to have this specified redundantly with the general blas/lapack provider settings?

climbfuji: This gets back to the openblas vs mkl debate. With Intel, the default provider is MKL (specificied in common/packages_intel.yaml). You said back then that you'd prefer the NOAA HPCs to keep using openblas until you have a chance of testing with MKL. Therefore, all the NOAA HPCs overwrite the virtual provider for blas/lapack with openblas. I found during concretizing that spack sometimes ignored this for numpy and used MKL directly (maybe because the virtual provider wasn't configured correctly in the package and/or our setup?).

AlexanderRichert-NOAA: Got it. I'm fine with this as is, and if we find a neater way to do the logic later then cool.

AlexanderRichert-NOAA (approved):
Concretizes fine on personal machine, concretizes and builds UWM deps fine on Hera.

climbfuji:
> Concretizes fine on personal machine, concretizes and builds UWM deps fine on Hera.

Yay! Thanks for testing :-)

configs/common/packages.yaml
climbfuji: Noted. I did set `zstd` to the version that got built in my and your tests (1.5.6) in my latest commit.

srherbener:
Testing skylab on S4 has been difficult. There is something about the flow that keeps inviting the login resource monitor to kill ecflow. We tried several changes to essentially run the flow in series, but ecflow still gets killed. I think we found an actual issue with a couple tasks based on matplotlib, but the matplotlib version didn't change (from 1.8.0). @fabiolrdiniz took a look and seemed to think that the matplotlib usage was not right, and was feeling that we need to look into that and address it with fixing the tasks.

It's not clear if shapely has anything to do with matplotlib tasks mentioned above, and seems unlikely, but we also saw a plotting task fail (which likely is affected by shapely). This is expected given the issue with the old shapely version.

It seems that the forecasting and variational tasks are intact and working. 

I think things are working well enough, that it makes sense to go ahead and merge this PR and then move forward fixing the issues we found. Do you agree @fabiolrdiniz?

fabiolrdiniz:
Thanks for tagging me, @srherbener. I'm a little concerned about the tasks being killed on S4. What we were speculating about exceeding the number of tasks running on the login node seems not to be the problem (since we set the workflow to run serially). I wonder if the same test works fine with another spack-stack version.

I agree that the matplotlib-related crashes may reveal an issue with how we plot figures. However, what is not clear to me is how confident we are with the functionality of that package. Is it capable of creating simple plots like the one below?

```python
#!/usr/bin/env python
import matplotlib.pyplot as plt
import numpy as np
x = np.array([1, 2, 3, 4])
y = x*2
plt.plot(x, y)
x1 = [2, 4, 6, 8]
y1 = [3, 5, 7, 9]
plt.plot(x1, y1, '-.')
plt.xlabel("X-axis data")
plt.ylabel("Y-axis data")
plt.title('multiple plots')
plt.tight_layout()
plt.savefig('test.png', format='png', dpi=300)
plt.close()
```

srherbener:
@fabiolrdiniz I just checked and the matplotlib version has been 3.7.4 since at least spack-stack-1.7.0. Perhaps something upstream from the plotting routine has changed (eg, messages from the forecast or variational apps) and the tasks with matplotlib need to be adjusted?

srherbener:
I'm building spack-stack with this PR feature branch, and corresponding spack submodule, on my Mac. I'll try testing jedi-bundle and skylab there and see what I get.

srherbener:
> Thanks for tagging me, @srherbener. I'm a little concerned about the tasks being killed on S4. What we were speculating about exceeding the number of tasks running on the login node seems not to be the problem (since we set the workflow to run serially). I wonder if the same test works fine with another spack-stack version.
> 
> I agree that the matplotlib-related crashes may reveal an issue with how we plot figures. However, what is not clear to me is how confident we are with the functionality of that package. Is it capable of creating simple plots like the one below?
> 
> ```python
> #!/usr/bin/env python
> import matplotlib.pyplot as plt
> import numpy as np
> x = np.array([1, 2, 3, 4])
> y = x*2
> plt.plot(x, y)
> x1 = [2, 4, 6, 8]
> y1 = [3, 5, 7, 9]
> plt.plot(x1, y1, '-.')
> plt.xlabel("X-axis data")
> plt.ylabel("Y-axis data")
> plt.title('multiple plots')
> plt.tight_layout()
> plt.savefig('test.png', format='png', dpi=300)
> plt.close()
> ```

@fabiolrdiniz, I tried this script in my spack-stack-1.8.0 environment and my newly built environment from this PR on my Mac. In both cases the script worked and I got this result:
test

srherbener:
With my new env based on this PR, jedi-bundle built successfully. I got a 39 ctest failures but that isn't too out of the ordinary for the Mac. Also, all of the fv3 and mpas tests pass and only two soca test failures: one due to a floating point exception the other due to a tolerance mismatch after the applicate completed successfully. All of the other test failures were ufo and saber tests.

I'm trying skylab now.

fabiolrdiniz:
Thanks for checking, @srherbener! By any chance, have you tried to run one of the experiments with your laptop build? If yes, did it fail similarly to S4? Thanks again!

fabiolrdiniz:
Sorry, only now I saw your previous message about SkyLab. We were writing messages at the same time. Thanks!

srherbener (approved):
Skylab on my Mac is working! It successfully completed the gfs-3dfgat-c12 experiment and made it through the first cycle successfully in the skylab-atm-land-small experiment.

climbfuji:
> Skylab on my Mac is working! It successfully completed the gfs-3dfgat-c12 experiment and made it through the first cycle successfully in the skylab-atm-land-small experiment.

Wohoo, this is great news. Thanks very much for your efforts testing this PR. I'll merge now. This is going to unlock a lot of updates for spack-stack-1.9.0.

Thanks again everyone for testing and reviewing.

Issue #1330, opened 2024-10-07
Remove sp library in spack-stack-1.9.0 --> spack-stack-2.1.0


climbfuji:
**Is your feature request related to a problem? Please describe.**

`sp` is deprecated and now provided by `ip@5` and later. We still provide `sp` in spack-stack-1.8.0, giving users enough time to transition to `ip`.

**Describe the solution you'd like**
No more `sp` in spack-stack-1.9.0.

**Additional context**

We need to make sure the applications are updated accordingly.

@climbfuji took care of NEPTUNE and GSIBEC. Also created https://github.com/NCAR/ccpp-physics/pull/1090 and https://github.com/NCAR/ccpp-scm/pull/524.

AlexanderRichert-NOAA:
The only thing using it is gsibec; I've asked for a release that can use ip (https://github.com/GEOS-ESM/GSIbec/pull/66).

climbfuji:
This is blocked by https://github.com/GEOS-ESM/GSIbec/issues/69

climbfuji:
Pushed out to spack-stack-2.0.0

climbfuji:
sp was removed from most, but not all, templates and configuration files in spack-stack-2.0.x. In spack-stack-2.1.x, we will remove the remaining bits and pieces (except the package in spack-packages, of course).

PR #1394, opened 2024-12-05
Build ip with gcc when using aocc as preferred compiler, bump ip to 5.1.0

climbfuji:
### Summary

Because `ip` doesn't build with the AMD `aocc` compilers, we switch to building it with `gcc` for that setup. While at it, bump `ip` to `5.1.0` and revert the workaround in the `neptune-env` template.

### Testing

Tested locally on my laptop: compile the neptune-env template (without Python packages) with `aocc@4.2.0` and verified that `ip` is built with `gcc` and the module loaded correctly.

### Applications affected

None (`aocc` is still experimental)

### Systems affected

None (`aocc` is still experimental)

### Dependencies

None

### Issue(s) addressed

Working towards https://github.com/JCSDA/spack-stack/issues/1330

### Checklist


configs/common/packages_aocc.yaml
```diff
@@ -13,3 +13,7 @@ packages:
   py-numpy:
     require:
     - '^openblas'
+  # ip@5 doesn't build with aocc@4 compilers
+  ip:
+    require:
+    - '%gcc'
```
climbfuji:
```suggestion
    - '%gcc'

```

AlexanderRichert-NOAA (approved):

PR #1410, opened 2024-12-11
Add new template cylc-dev (and test building in GitHub actions)

### LINKED ISSUE #1413: [QUESTION] Is JCSDA still using jedi-tools-env / do we need the +excel variant for py-pandas?

I am wondering if JCSDA is still using `jedi-tools-env`. Back in the days when I was at JCSDA, we used `jedi-tools-env` to run some of the GitHub admin jobs, for example managing labels across GitHub repositories, collecting GitHub statistics, ...

I don't recall exactly for which of these tasks, but I had to add a `+excel` variant to `py-pandas` for one of them. Now, the `+excel` variant is becoming problematic, since newer versions of `py-pandas` require `py-python-calamina -> py-maturin -> rust` as dependencies. This is the full `rust`, not just `rust-bootstrap`. And `py-maturin` doesn't compile with Intel Classic or Intel LLVM either.

If we don't need it anymore, I'd like to get rid of the `+excel` variant for `py-pandas`. Even better, if `jedi-tools-env` is no longer needed (maybe/hopefully @eap is handling all of these tasks differently with Terraform nowadays?), I'd like to get rid of it entirely.

If none of this is possible, then we should modify the common package definitions such that the `+excel` variant is only used with the GCC compiler (which is also the only compiler that triggers the build of `jedi-tools-env`).

**Additional context**

See https://github.com/JCSDA/spack-stack/pull/831#issue-1946218587 for the PR that added the `+excel` variant.

### LINKED ISSUE #1444: spack-stack environment build errors not caught in GitHub actions
For https://github.com/JCSDA/spack-stack/pull/1410/commits/ede071b8ef116f7ec9932a32de6e754e7af2648e, GitHub actions reports that the tests succeeded, but in fact the build of `cylc-dev` failed: https://github.com/JCSDA/spack-stack/actions/runs/12662753274/job/35288186039?pr=1410

_Originally posted by @climbfuji in https://github.com/JCSDA/spack-stack/issues/1410#issuecomment-2577956331_

climbfuji:
### Summary

This PR adds a new template `cylc-dev` to build a special environment for running `cylc`. Because of the way `cylc` works, this is best accomplished by creating an environment view and then a `cylc` wrapper that contains the minimum settings necessary to use `cylc` from the view's `bin` directory. See https://github.com/cylc/cylc-flow/issues/6532 for more information. Note that the creation of the `cylc` wrapper is not part of this PR.

Associated changes:
- To compile `cylc-dev` on the Ubuntu self-hosted runner, exclude `meson` from the `spack external find` call. Spack found a deprecated version `meson@0.61` and despite it being deprecated it tried to use it - with the result that the `harfbuzz` build failed (needed for `cylc-dev`). Letting spack build `meson` (it picks `meson@1.5.1`) solved the problem.
- Remove the `+excel` variant from `py-pandas`, which was something the `jedi-tools-env` used, but that is no longer needed and that caused build errors (and added several more dependencies).
- Fix a bug in the GitHub actions that would allow the tests to fail silently (replace `set +e` with `set -e`).
- To avoid duplicate packages, require `py-cython@3` and `py-numpy@1.26`
- With Intel Classic and LLVM, build `qt` with `gcc` (has no effect on pre-configured sites, where `qt` is configured as an external package)

### Testing

- [x] CI (also tests `cylc-dev` template with `gcc` on Ubuntu)
- [x] Atlantis (built `cylc-dev` with `gcc`, created `cylc` wrapper, tested flow, gui and NEPTUNE workflow)
- [x] Narwhal (built `cylc-dev` with `gcc`, created `cylc` wrapper, tested flow, gui)
- [x] Nautilus (built `cylc-dev` with `gcc`, created `cylc` wrapper, tested flow, gui)

### Applications affected

None

### Systems affected

None

### Dependencies

- [x] waiting on https://github.com/JCSDA/spack/pull/491
- [x] https://github.com/JCSDA/spack/pull/499

### Issue(s) addressed

Resolves https://github.com/JCSDA/spack-stack/issues/1413
Resolves https://github.com/JCSDA/spack-stack/issues/1444

### Checklist


climbfuji:
Note. For https://github.com/JCSDA/spack-stack/pull/1410/commits/ede071b8ef116f7ec9932a32de6e754e7af2648e, GitHub actions reports that the tests succeeded, but in fact the build of `cylc-dev` failed: https://github.com/JCSDA/spack-stack/actions/runs/12662753274/job/35288186039?pr=1410

.github/workflows/ubuntu-ci-x86_64-gnu.yaml
```diff
@@ -51,12 +51,15 @@ jobs:
           # Set up spack-stack
           source ./setup.sh
 
-          declare -a TEMPLATES=("unified-dev" "skylab-dev")
+          #declare -a TEMPLATES=("unified-dev" "skylab-dev" "cylc-dev")
```
climbfuji: Revert lines 54 and 55 before merging!

.github/workflows/ubuntu-ci-x86_64-gnu.yaml
```diff
@@ -113,20 +117,22 @@ jobs:
...
-            # jedi-base-env
-            echo "jedi-base-env ..."
-            spack install --fail-fast --source --no-check-signature jedi-base-env 2>&1 | tee log.install.gnu-11.4.0-buildcache.jedi-base-env
-            spack buildcache create -u /home/ubuntu/spack-stack/build-cache/ jedi-base-env
-
-            # jedi-ufs-env
-            echo "jedi-ufs-env ..."
-            spack install --fail-fast --source --no-check-signature jedi-ufs-env 2>&1 | tee log.install.gnu-11.4.0-buildcache.jedi-ufs-env
-            spack buildcache create -u /home/ubuntu/spack-stack/build-cache/ jedi-ufs-env
+            if [[ "${TEMPLATE}" == *"unified-dev"* || "${TEMPLATE}" == *"skylab-dev"* ]]; then
```
climbfuji, line 119: The new cylc-dev template doesn't depend on base-env, jedi-base-env, jedi-ufs-env

.github/workflows/macos-ci-aarch64.yaml
```diff
@@ -111,7 +111,7 @@ jobs:
             set +e
             spack mirror add local-binary file:///Users/ec2-user/spack-stack/build-cache/
             spack buildcache update-index local-binary || (echo "No valid binary cache found, proceed without" && spack mirror rm local-binary)
-            set +e
+            set -e
```
eap, line 114: Good catch here and elsewhere.

configs/common/packages.yaml
```diff
@@ -233,13 +232,15 @@ packages:
...
   py-matplotlib:
-     require: '@3.7.4'
+    require: '@3.7.4'
   # Variant ~mpi not working for latest py-netcdf4
   # https://github.com/spack/spack/issues/47652
   py-netcdf4:
     require: '@1.7.1 +mpi'
-  py-pandas:
-    require: '+excel'
+  # To avoid duplicate packages
```
eap, line 240: Wouldn't the concretizer `unify: true` prevent duplication?

eap (approved):
Thanks for this, based on description I'm still a little puzzled, I thought we were also intending to have a cylc meta module that could be loaded along side of other modules as in our unified dev environment.

I'm approving because this code looks fine and I assume you'll have a good response to my question above.

configs/common/packages.yaml
climbfuji: It will lead to concretization errors if `unify: true` is set. Unfortunately, the concretizer isn't perfect, it still has its bugs. But I am happy to move this (numpy and cython) into the cylc-dev template instead of applying it to all environments for now (but new packages that are coming down the road will want the same settings  for unified-env/neptune-env, too).

climbfuji:
> Thanks for this, based on description I'm still a little puzzled, I thought we were also intending to have a cylc meta module that could be loaded along side of other modules as in our unified dev environment.
> 
> I'm approving because this code looks fine and I assume you'll have a good response to my question above.

Initially, the idea was to add cylc to the unified-env and other larger environments. But, as we are expanding the use of Python in our environments (more and more packages with complicated dependencies), this is getting harder and harder.

Especially `cylc` has so many tight version requirements that it's impossible to build it in unified-env without duplicate packages ( (and is very hard to build with anything else but gcc). Then you also need the view+wrapper (wrapper is currently a manual process) steps to make it work with CYLC_PYTHONPATH and keep the environment to run `cylc` in and the one that `cylc` tasks use completely separate. This is a non-trivial setup to get right, and the link to the cylc issue/discussion is only the entry point. I've come to learn that cylc is designed to operate in its own, separate environment (which does make a lot of sense), and this can be done best with two separate spack-stack environments.

areinecke (approved):

stiggy87 (approved):

PR #1393, opened 2024-12-05
Add a tier2 site config for AWS AMIs

stiggy87:
### Summary

This is to add AWS as a tier2 supported platform. The two images are Rock8 and Ubuntu24.04. 

### Testing

I have custom built two running EC2 instances following these steps. However, Intel has had problems on both, and would like others in review to give it a test and provide feedback.

### Applications affected

None. These are just platforms that can be used for development.

### Systems affected

None. These are to support Rocky8 and Ubuntu24.04 EC2 instances.

### Dependencies

PR #1419

### Issue(s) addressed

https://github.com/JCSDA/spack-stack/issues/1379

### Checklist
- [x] This PR addresses one issue/problem/enhancement, or has a very good reason for not doing so.
- [x] These changes have been tested on the affected systems and applications.
- [x] All dependency PRs/issues have been resolved and this PR can be merged.

PR #1416, opened 2024-12-13
Add the GitHub CLI to the global workflow env

### LINKED ISSUE #1105: [INSTALL]: GitHub CLI
### Package name

gh

### Package version/tag

v2.28.0

### Build options

None

### Installation timeframe

Only needed on all supported RDHPCS systems.  Already in **lmod** on WCOSS2.

### Other information

_No response_

### WCOSS2

- [ ] Check this box if and only if your package should be installed on WCOSS2 Cactus and Dogwood (all spack-stack packages will be installed on Acorn). If not, you may disregard the rest of the items below and submit this request.

### WCOSS2: General questions

_No response_

### WCOSS2: Installation and testing

_No response_

### WCOSS2: Technical & security review list

- [ ] The code is mature, stable, and production ready
- [ ] The code is does not and cannot use the internet, and does not contain URLs (http, https, ftp, etc.) except in comments
- [ ] The package does not contain prebuilt binary files that have not been approved by NCO security review
- [ ] The code has no publicly disclosed cybersecurity vulnerabilities and exposures (search https://cve.mitre.org/cve/)
- [ ] The code is not prohibited by DHS, DOC, NOAA, or NWS
- [ ] The code comes from a trusted source. Trusted sources include other NWS, NOAA, or DOC, agencies, or other Federal agencies that operate at a FISMA high or equivalent level. Additionally, trusted sources could be third-party agencies through which there is an existing SLA on file (such as RedHat).
- [ ] The code is actively maintained and supported (it continues to get updates, patches, etc.)
- [ ] The code is not maintained by a private entity operating in a foreign country (if it is, make a note below)
- [ ] There is sufficient documentation to support maintenance
- [ ] There are no known security vulnerabilities or weaknesses
- [ ] Installing and running the code does not require privileged processes/users
- [ ] There are no software dependencies that are unapproved or have security concerns (if there are, make a note below)
- [ ] There are no concerns related to SA, SI, and SC NIST control families

### WCOSS2: Additional comments

_No response_

eap:
This change adds the GitHub CLI to the global workflow env. Go language based dependencies don't compile on the intel stack, so some intel-specific config has been added to force the use of an alternate GCC compiler. This change has been tested in GCC and Intel on an Ubuntu 24.04 site config.

### Testing

As above, tested in Ubuntu on GCC/Intel

### Applications affected

Global Workflow Env.

### Systems affected

N/A

### Dependencies

None

### Issue(s) addressed

Fixes: https://github.com/JCSDA/spack-stack/issues/1105

### Checklist


climbfuji:
The intel build fails because it can't build grep. I am pretty sure that we can assume grep is available as an external package on all systems.

eap:
@climbfuji I'm going to fix that anyways; we are observing the same grep issue on our Ubuntu site config (see [discussion here](https://github.com/JCSDA/spack-stack/pull/1393/files#r1883305992) - Steve has also verified that he is seeing it).

eap:
Ok, I've added the fix to this PR. No idea why this issue is showing up now but it seems to be affecting many intel builds and the fix is pretty easy.

.github/workflows/ubuntu-ci-x86_64-intel.yaml
```diff
@@ -115,6 +115,14 @@ jobs:
           echo "    - spec: qt@5.15.3" >> ${SPACK_SYSTEM_CONFIG_PATH}/packages.yaml
           echo "      prefix: /usr" >> ${SPACK_SYSTEM_CONFIG_PATH}/packages.yaml
 
+          # Add external grep.
+          echo "" >> ${SPACK_SYSTEM_CONFIG_PATH}/packages.yaml
+          echo "  grep:" >> ${SPACK_SYSTEM_CONFIG_PATH}/packages.yaml
+          echo "    buildable: False" >> ${SPACK_SYSTEM_CONFIG_PATH}/packages.yaml
+          echo "    externals:" >> ${SPACK_SYSTEM_CONFIG_PATH}/packages.yaml
+          echo "    - spec: grep@3.11" >> ${SPACK_SYSTEM_CONFIG_PATH}/packages.yaml
+          echo "      prefix: /usr" >> ${SPACK_SYSTEM_CONFIG_PATH}/packages.yaml
```
eap: This was added to fix the currently broken intel build. See the discussion linked in my response to Dom. This PR isn't the cause of the failure but I've included the fix here to unblock this PR and others.

climbfuji: I wonder if we should make this a "default" for all sites that we maintain:
- Check if spack has the necessary information to do `spack external find grep`
- Add this command to all GitHub workflows
- Update documentation (NewSiteConfig) to include the external find grep
- Add external grep spec for all containers and all preconfigured sites (tier-1; tier-2 is up to the individuals maintaining those sites)

eap: hmm, that's an interesting idea. For reasons I don't fully understand this only manifests when building with intel and gcc-11 which is the case in this system and in the ubuntu image I referenced. Intel plus gcc-12 or gcc-13 and grep is fine (but other things break). Good catch on 'exteral find' which definitely was the correct call here.

I think without more info on where/why this happens I'd be hesitant to make this default everywhere even though it's probably not harmful.

eap: Alright - the change I added using `external find grep` doesn't work and we have to manually modify the packages file. This is making the intel packages change more interesting since grep is definitely available everywhere

stiggy87 (approved):

climbfuji (changes requested):
This PR is blocked by https://github.com/JCSDA/spack-stack/pull/1419

.github/workflows/ubuntu-ci-x86_64-intel.yaml
```diff
@@ -62,6 +62,7 @@ jobs:
               --exclude curl --exclude python
           spack external find --scope system grep
           spack external find --scope system sed
+          spack external find --scope system grep
```
climbfuji:
```suggestion
```

climbfuji (approved):

PR #1418, opened 2024-12-15
Add grep to linux tier2 site config to fix intel build errors [proposed change and discussion]

eap:
This is a proposed change to address the failing grep build observed when building with the intel legacy compiler and gcc-11.x. A log of the failure is included [here: spack-stack-intel-grep-build-failure.txt](https://github.com/user-attachments/files/18142302/spack-stack-intel-grep-build-failure.txt). While this PR includes a specific fix by updating the `linux.default` site config, I'll outline several other options none of which seem much more satisfactory. I'm open to changing this PR if anyone can think of a compelling reason to use a different approach.

Fixes considered:
 * Update the linux.default site config to use an external build for grep [current proposal]
     * Pros: works for new site configs and integration tests
     * Cons: not compiler specific.
 * Update the `configs/common/packages_intel.yaml` to use an external package.
      * Pros: specific to intel,.
      * Cons: affects existing sites where this isn't an issue.
 * Update the integration test yaml to manually apply this fix. 
      * Pros: Fixes tests,.
      * Cons: leaves this issue to be resolved manually for site-config creators.
 * Use `spack external find grep`.
     * Cons: spack is unable to find this package.

### Testing

Manually tested on ubuntu and tested with CI.

### Applications affected

N/A

### Systems affected

New linux site configs.

### Dependencies

This change is blocking https://github.com/JCSDA/spack-stack/pull/1416 (and possibly more).

### Checklist

### DIFF
--- a/configs/sites/tier2/linux.default/packages.yaml
+++ b/configs/sites/tier2/linux.default/packages.yaml
@@ -1,0 +1,6 @@
+packages:
+  grep:
+    buildable: false
+    externals:
+    - spec: grep@3.11
+      prefix: /usr


climbfuji (changes requested):
This is unfortunately not the way to do it. We don't know the version of grep on the system (even though it may not matter so much). For new site configs, the change must go into the "New Site Config" section of the documentation (doc/source/...) after verifying that the command `spack external find --scope=system grep` works (or after making it work). We should do that for macOS, too, while we are at it.

Lastly, we should add a list to the PR that adds the external grep to all tier-1 and tier-2 site configs (yes, in this case to packages.yaml, because these are specific systems).

PR #1432, opened 2024-12-22
Update intel and gcc containers - fix builds and move to ubuntu 24.04

### LINKED ISSUE #1383: JEDI CI container build with Intel broken since update from spack develop as of 2024/10/31
**Describe the bug**
See https://github.com/JCSDA/spack-stack/pull/1381:
```
#18 ERROR: process "docker-shell cd /opt/spack-environment && spack env activate . && spack install --fail-fast && spack gc -y" did not complete successfully: exit code: 1
------
 > [builder 6/8] RUN cd /opt/spack-environment && spack env activate . && spack install --fail-fast && spack gc -y:
33.03 ==> Error: failed to concretize `fms@2024.02`, `g2@3.5.1`, `py-shapely@1.8.0`, `udunits@2.2.28`, ..., `py-pygithub@2.1.1` for the following reasons:
33.03      1. rust: Rust not compatible with Intel Classic compilers
33.03      2. rust: Rust not compatible with Intel Classic compilers
33.03         required because conflict is triggered when %intel 
33.03         required because conflict constraint  . You could consider setting `concretizer:unify` to `when_possible` or `false` to allow multiple versions of some packages.
```

**To Reproduce**
See https://github.com/JCSDA/spack-stack/pull/1381 or GitHub actions.

**Expected behavior**
No build errors

**System:**
JEDI CI container builds on Ubuntu self-hosted runner

**Additional context**

eap:
This change updates the Intel and GCC containers fixing the intel build and updating both environments to Ubuntu 24.04 to more closely align with our AWS machine images (AMIs). This update focuses on the Intel and GCC images since they are the core of our supported compiler environments,

__What about clang__? I will submit a similar fix for clang in an immediate followup change. Container images are especially difficult to update due to their longer builds and slightly different (often conflicting) configuration. I have not included clang in this PR in order to speed the delivery of these two containers (our core supported compiler platforms) and reduce the occurance of merge-related rewrites.

__What about oneapi-intel?__ The new llvm-based intel image will be submitted as a new image config outside the scope of this PR.

### Testing

This has been built and tested on an AWS intel instance.

### Applications / Systems affected

Presubmit CI and containers.

### Issue(s) addressed

Fixes: https://github.com/JCSDA/spack-stack/issues/1383

### Checklist


configs/containers/docker-ubuntu-gcc-openmpi.yaml
```diff
@@ -4,19 +4,19 @@ spack:
 
   config:
     checksum: false
-    build_jobs: 2
+    build_jobs: 5
```
eap: I'm realizing now that I aught to reduce build jobs back to `2` due to test automation.

climbfuji:
@eap Ignore the non-container build failures for the purpose of this PR (as long as you are not changing any files used by both, which at the moment you don't). I'll try to fix those tomorrow. I am surprised, though. The CI tests all passed last week for every PR we merged ...

configs/containers/specs/jedi-ci.yaml
```diff
@@ -38,7 +38,6 @@
     py-f90nml,
     py-gitpython,
     py-h5py,
-    py-numpy,
```
climbfuji, line 41: I assume you are removing this because it is implicitly included with py-pandas?

configs/containers/docker-ubuntu-gcc-openmpi.yaml
climbfuji: I think so, yes

stiggy87 (approved):

configs/containers/specs/jedi-ci.yaml
eap: correct - unnecessary inclusion and earlier when there was a version specified it was messing with concretize somehow due to the intel variant. I probably could have fixed it, but since it is implicitly included it isn't necessary in the spec.

configs/containers/docker-ubuntu-gcc-openmpi.yaml
eap: fixed.

climbfuji (approved):

PR #1434, opened 2024-12-24
Spack stack containers for intel-oneapi and clang14

eap:
This change updates the clang container to _Ubuntu24.04+Clang-14_ and this change creates the intel oneapi (LLVM-based) container. 

This is the first pass at creating the intel oneapi container and it functions to install spack-stack but has not been tested with JEDI. I will iterate on this base image if updates are needed to build and test JEDI/skylab.

### Testing

This has been built and tested on an Ubuntu 24.04 host with docker.

### Applications / Systems affected

Presubmit CI

### Checklist


configs/containers/docker-ubuntu-oneapi-impi.yaml
```diff
@@ -0,0 +1,320 @@
...
+      extra_rpaths: []
+  - compiler:
+      spec: oneapi@=2024.2.1
+      paths:
+        cc: /opt/intel/oneapi/compiler/2024.2/bin/icx
+        cxx: /opt/intel/oneapi/compiler/2024.2/bin/icpx
+        f77: /opt/intel/oneapi/compiler/2024.2/bin/ifx
+        fc: /opt/intel/oneapi/compiler/2024.2/bin/ifx
+        #f77: /opt/intel/oneapi/compiler/2024.2/bin/ifort
+        #fc: /opt/intel/oneapi/compiler/2024.2/bin/ifort
```
eap: I don't know which is the preferred compiler here - @climbfuji 's August 15 comment in #912 and the checked in code in `configs/sites/tier2/blackpearl` implies that ifort might work better, but I did build with `ifx` so maybe some of that info is out of date.

climbfuji: See my comment in slack (https://jcsda.slack.com/archives/C04D6ALT0SX/p1734999495177329). As a first pass and test container for JEDI, I would go with ifort. Especially if you are using 2024.2.

configs/containers/docker-ubuntu-oneapi-impi.yaml
```diff
@@ -0,0 +1,320 @@
...
+      flags: {}
+      operating_system: ubuntu24.04
+      target: x86_64
+      modules: []
+      environment:
+        # change to append_path once relevant PR is merged
```
climbfuji, line 38:
```suggestion
```

configs/containers/docker-ubuntu-oneapi-impi.yaml
```diff
@@ -0,0 +1,320 @@
...
+        append_path:
+          CPATH: /opt/intel/oneapi/compiler/2024.2/opt/compiler/include/intel64
```
climbfuji, line 40: FYI, I think this is only needed with ifort, so for now keep the CPATH line.

configs/containers/docker-ubuntu-oneapi-impi.yaml
```diff
@@ -0,0 +1,320 @@
...
+        # Needed by py-llvmlite
+        update-alternatives --install /usr/bin/llvm-config llvm-config /usr/bin/llvm-config-14 100 && \
+        rm -rf /var/lib/apt/lists/*
+      pre_final: |
+        # Set environment variables for installing tzdata
+        ENV DEBIAN_FRONTEND=noninteractive
+        ENV TZ=Etc/UTC
+        ENV CC=icx
+        ENV CXX=icpx
+        ENV FC=ifx
```
climbfuji: Change to ifort

configs/containers/docker-ubuntu-oneapi-impi.yaml
```diff
@@ -0,0 +1,320 @@
...
+        /opt/views/view/bin/pip3 install pynacl
+        # Copy spack find output from builder
+        COPY --from=builder /root/spack_find.out /root/spack_find.out
+        # Make a non-root user:nonroot / group:nonroot for running MPI
+        RUN useradd -U -k /etc/skel -s /bin/bash -d /home/nonroot -m nonroot --uid 43891 && \
+        echo "ulimit -s unlimited" >> /home/nonroot/.bashrc && \
+        echo "ulimit -v unlimited" >> /home/nonroot/.bashrc && \
+        echo "export CC=icx" >> /home/nonroot/.bashrc && \
+        echo "export CXX=icpx" >> /home/nonroot/.bashrc && \
+        echo "export FC=ifx" >> /home/nonroot/.bashrc && \
```
climbfuji: ifort

configs/containers/docker-ubuntu-oneapi-impi.yaml
```diff
@@ -0,0 +1,320 @@
...
+        echo "source /opt/intel/oneapi/setvars.sh" >> /etc/bash.bashrc && \
+        printf "[credential]\n    helper = cache --timeout=7200\n" >> /home/nonroot/.gitconfig && \
+        chown -R nonroot:nonroot /home/nonroot/.gitconfig
+        # Replicate settings for root user
+        RUN echo "ulimit -s unlimited" >> /root/.bashrc && \
+        echo "ulimit -v unlimited" >> /root/.bashrc && \
+        echo "export CC=icx" >> /root/.bashrc && \
+        echo "export CXX=icpx" >> /root/.bashrc && \
+        echo "export FC=ifx" >> /root/.bashrc && \
```
climbfuji: ifort

configs/containers/docker-ubuntu-oneapi-impi.yaml
eap: Thanks. That sounds like a good balance. I switched this back to ifort and it still compiles with no other changes.

stiggy87 (approved):

configs/containers/specs/jedi-ci.yaml
```diff
@@ -15,7 +15,7 @@
     eckit@1.28.3,
     ecmwf-atlas@0.40.0 +fckit +trans +tesselation +fftw,
     fiat@1.4.1,
-    ectrans@1.2.0 +fftw,
+    ectrans@1.2.0,
```
climbfuji, line 18: This must have slipped somehow in my first round of reviews. Why is this being turned off? Or is the default for variant `fftw` true?

eap: The intel oneapi definition requires the ectrans `+mkl ~fftw` variant and the container definition ended up with both variants listed causing the concretizer to fail because it could not unify.

We could potentially add the `+fftw` variants to all of the other containers if this is needed. I noticed that the common config for ectrans also had the fftw variant [disabled recently](https://github.com/JCSDA/spack-stack/commit/491b30fa8922bf2e0bbbb4774227095a6c1bf3b1)

climbfuji: I wonder how this works for Intel Classic (%intel)

eap: well, the easy answer is that it works because intel classic doesn't specify a variant therefore gets the +fftw variant. I take it that we should be using the `+mkl ~fftw` variant for both intel oneapi and classic?

climbfuji: I think that would be better. Although Intel Classic is a dying horse ...

eap: I've added the correct variant to the intel classic container. Its building now and I'll close this thread once I verify the change.

climbfuji (approved):
Happy with the changes if you are, too.

configs/containers/specs/jedi-ci.yaml
eap: The updated old intel build passed. Resolving this thread.

PR #1464, opened 2025-01-21
Add util/nrl/batch_install.sh, add patch to build py-netcdf4 without mpi

### LINKED ISSUE #1463: py-netcdf4 must be buildable with `~mpi`, even if netcdf is built with ~+mpi`
**Describe the bug**
Previous versions of `py-netcdf4` had a switch to turn on and off parallel features. Recent versions (1.7.1-post2, for example) auto-detect parallel features from the netCDF header. This is unfortunate, because it leads to errors like this with spack-stack develop as of today:
```
ValueError: mpi4py.MPI.Status size changed, may indicate binary incompatibility
```
In https://github.com/Unidata/netcdf4-python/issues/1389, @AlexanderRichert-NOAA and I are discussing the need for an explicit switch. We can apply a patch for existing versions `@1.7` to turn off parallel features when `~mpi~` is requested. We then also need to change the variant request in `configs/common/packages.yaml`.

**To Reproduce**
See above

**Expected behavior**
No error running `py-netcdf4`, and the ability to build without parallel features based on a variant, not based on auto-detect.

**System:**
Any

**Additional context**
The py-netcdf4 builds have diverged between spack-stack-dev and spack develop, because the new build in spack develop didn't work for us at all. I am talking to Chris Marsh, who maintains the py-netcdf4 package upstream about the need to toogle MPI explicitly. Once the package has been fixed upstream, @climbfuji will consolidate the two versions after the 1.9.0 release.

climbfuji:
### Summary

Add `util/nrl_batch_install.sh`. This includes a workaround for installing the `cycl` environment on Narwhal, for which we need to switch from the Cray compiler wrappers for `gcc@10` to the native compilers (lines 258-264 in `nrl_batch_install.sh`). I will work on a better solution for this in the next weeks.

During the installation, three problems were discovered that are fixed in this PR:
1. For `gcc@:10`, need to use older version of `cdo` in `configs/common/packages.yaml` that uses C++-17 instead of C++-20 (same as existing modifications for Intel Classic)
2. For Atlantis, remove external (old) bison to avoi duplicate packages
3. We need to be able to build `py-netcdf` without MPI (`~mpi`) even when `netcdf-c` was built with `+mpi` to avoid errors mentioned in https://github.com/JCSDA/spack-stack/issues/1463

Also update submodule pointer for spack for changes recently merges into spack-stack-dev (`py-numpy` bug fix for `%intel`).

### Testing

Used `util/nrl_batch_install.sh` to install all necessary environments on Atlantis, Narwhal, Nautilus.

### Applications affected

None

### Systems affected

Atlantis, Narwhal, Nautilus

### Dependencies

- [x] Waiting on https://github.com/JCSDA/spack/pull/502

### Issue(s) addressed

Resolves https://github.com/JCSDA/spack-stack/issues/1463

### Checklist


util/nrl_batch_install.sh
```diff
@@ -0,0 +1,404 @@
+#!/usr/bin/env bash
```
AlexanderRichert-NOAA: If it's not too inconvenient, could this go in a subdirectory, say, "nrl/"? I guess in part it depends whether you think there will be other NRL-specific utilities down the road...

climbfuji: That's ok with me. How about the site-specific subdirectories `acorn` and `narwhal`? Leave them where they are now?

AlexanderRichert-NOAA: I think so yes, and if more files/directories get added we can always reorganize later (like add top-level directories by org).

climbfuji: Done

AlexanderRichert-NOAA (approved):

areinecke (approved):

Issue #1485, opened 2025-01-30
spack-stack-1.9.0 --> 1.9.1 release tasks


climbfuji:
## List of tasks for the release

Project board: https://github.com/orgs/JCSDA/projects/9/views/1

- [x] Create release branches for spack and spack-stack, update `doc/source/conf.py` and `.gitmodules` to use `release/1.9.0` instead of `develop`
- [x] Clear project board (no open pull requests other than for specific site configs or documentation)
- [x] Prepare Wiki page for release information
- [x] Roll out **release 1.9.1** and update documentation (readthedocs), site config, and wiki page each time - **blocked by "Clear project board"**
    - [x] acorn (@AlexanderRichert-NOAA)
    - [x] atlantis (@climbfuji)
    - [x] ~~blueback~~ (@climbfuji; not yet available)
    - [x] derecho (@rickgrubin-noaa or @RatkoVasic-NOAA --> @climbfuji)
    - [x] ~~discover-scu16~~ no longer required
    - [x] discover-scu17 [@ashley314]
    - [x] gaea-c5 (@rickgrubin-noaa or @RatkoVasic-NOAA)
    - [x] gaea-c6 (@rickgrubin-noaa or @RatkoVasic-NOAA)
    - [x] hera (@RatkoVasic-NOAA)
    - [x] hercules (@rickgrubin-noaa)
    - [x] ~~jet~~ no longer maintained by EPIC
    - [x] narwhal (@climbfuji)
    - [x] nautilus (@climbfuji)
    - [x] ~~navy-aws~~ (@climbfuji; not yet available)
    - [x] noaa-aws (@RatkoVasic-NOAA)
    - [x] noaa-azure (@RatkoVasic-NOAA)
    - [x] noaa-gcloud (@RatkoVasic-NOAA)
    - [x] orion (@RatkoVasic-NOAA)
    - [x] s4 (@InnocentSouopgui-NOAA)
    - [x] ursa (@ulmononian )
    - [x] ~~JCSDA AWS Parallelcluster~~ [no release unless requested]
    - [x] JCSDA AWS AMI [@stiggy87 ]
    - [x] JCSDA CI containers [@eap] - [complete; updated aws and hub.docker.com]
- [x] Tag spack, update `doc/source/conf.py` and `.gitmodules` to use tag instead of release branch, tag spack-stack (@climbfuji)
- [x] Prepare and publish release notes (@climbfuji): https://github.com/JCSDA/spack-stack/releases
[...]

Issue #1116, opened 2024-05-16
Site config cleanup


AlexanderRichert-NOAA:
For the 1.8.0 release, let's do the following for each site config:
 - Verify OSes
 - Remove old/unused compilers
 - Check compiler versions
 - Check external package paths/versions
 - Create a test environment, concretize, and install a few packages

Site list:
 - [x] acorn (@AlexanderRichert-NOAA)
 - [x] atlantis (@climbfuji)
-  [x] blueback (@climbfuji)
 - [x] ~~aws-pcluster~~ currently not supported
 - [x] casper - move to tier2 (@climbfuji)
 - [x] derecho
 - [x] ~~discover-scu16~~ no longer supported
 - [x] discover-scu17
 - [x] ~~emc-rhel~~ currently not supported
 - [x] ~~frontera~~ no longer supported
 - [x] gaea-c5
 - [x] gaea-c6
 - [x] hera
 - [x] hercules
 - [x] ~~jet~~ no longer supported
 - [x] linux.default
 - [x] macos.default
 - [x] narwhal (@climbfuji)
 - [x] nautilus (@climbfuji)
 - [x] noaa-aws
 - [x] noaa-azure
 - [x] noaa-gcloud
 - [x] orion
 - [x] s4

climbfuji:
@AlexanderRichert-NOAA I did quite a bit of a cleanup in my last PR (the MKL PR #759). Given that, and the fact that we have to revisit all the sites when we remove the "revert to openblas" logic, are you ok with pushing this out to after the 1.8.0 release?

AlexanderRichert-NOAA:
Yep that's fine. Plus the automation work will be a good opportunity to review site configs.

climbfuji:
This is all covered by the site config updates for spack-stack-1.9.1. Will close as completed once 1.9.1 is available everywhere and the release code is tagged with all site config changes included.

PR #1499, opened 2025-02-07
1.9.0: update acorn site config

AlexanderRichert-NOAA:
### Summary

Site update for acorn. When 1.9.0 is merged into develop, '+python' should be in the esmf require line with the other variants in common/packages.yaml.

### Testing

tested on acorn

### Applications affected

UWM, GW, GSI

### Systems affected

Acorn

### Dependencies

none

### Issue(s) addressed

part of #1485, #1116

### Checklist

### DIFF
--- a/configs/sites/tier1/acorn/packages.yaml
+++ b/configs/sites/tier1/acorn/packages.yaml
@@ -41,3 +41,6 @@
       externals:
       - spec: grep@3.1
         prefix: /usr
+    esmf:
+      require:
+      - '+python'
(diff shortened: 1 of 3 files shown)


configs/sites/tier1/acorn/packages.yaml
```diff
@@ -41,3 +41,6 @@
       externals:
       - spec: grep@3.1
         prefix: /usr
+    esmf:
+      require:
+      - '+python'
```
climbfuji, line 46: I hopefully remember this!

climbfuji (approved):

Hang-Lei-NOAA (approved):

configs/sites/tier1/acorn/packages.yaml
climbfuji: And we need to check if that works all the time. We have environments (`neptune-env`) that don't require it (yet).

PR #1545, opened 2025-03-05
Update setup.sh: add export SPACK_DISABLE_LOCAL_CONFIG=true

### LINKED ISSUE #1544: spack-stack environments must not depend on existing user configurations

This is a comment from the operational folks at Fleet Numerical:
> If one is only dealing with the development of one application then the `.spack` directory can be whatever it is. However, for us we have to coordinate many applications and in the future other modules that use spack-spack that may have conflicts for the library user [who may use spack without spack-stack, too]. It means that we will have to remove the `.spack` directory with every build of an application.

Isolate spack-stack environments from any existing user configuration of spack, typically in `~/.spack/`

**Additional context**

We discussed this a while ago but decided against it, because we didn't have a good way and good documentation to create bootstrap mirrors for air-gapped systems. This was fixed in https://github.com/JCSDA/spack-stack/pull/1468 a while ago.

climbfuji:
### Summary

Update `setup.sh`: add `export SPACK_DISABLE_LOCAL_CONFIG=true` to completely isolate the spack-stack environments from user spack configurations. This requires additional steps to unset and restore the environment variable when creating new site configurations. Further, remove the exception for macOS (no longer needed, see https://github.com/JCSDA/spack-stack/pull/1545#issuecomment-2704598741 below for more information)

### Testing

- [x] Tested on NRL systems and @climbfuji's devbox
- [x] CI

### Applications affected

None

### Systems affected

"All" (not really), but at least air-gapped systems where the spack bootstrap mirror needs to be used every time from now on.

### Dependencies

n/a

### Issue(s) addressed

Closes https://github.com/JCSDA/spack-stack/issues/1544

### Checklist

### DIFF
--- a/.github/workflows/macos-ci-aarch64.yaml
+++ b/.github/workflows/macos-ci-aarch64.yaml
@@ -63,6 +63,8 @@ jobs:
           export ENVDIR=$PWD/envs/${ENVNAME}
           spack stack create env --site macos.default --template unified-dev --name ${ENVNAME} --compiler apple-clang
           spack env activate ${ENVDIR}
+
+          unset SPACK_DISABLE_LOCAL_CONFIG
           export SPACK_SYSTEM_CONFIG_PATH="${ENVDIR}/site"
 
           # Find external packages
@@ -83,7 +85,8 @@ jobs:
           # Find compilers
           spack compiler find --scope system
 
-          export -n SPACK_SYSTEM_CONFIG_PATH
+          export SPACK_DISABLE_LOCAL_CONFIG=true
+          unset SPACK_SYSTEM_CONFIG_PATH
 
           # For buildcaches
           spack config add config:install_tree:padded_length:200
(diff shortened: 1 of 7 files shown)


RatkoVasic-NOAA (approved):
LGTM

stiggy87 (approved):

climbfuji:
Ooops, I remember. The problem was this didn't work for new site configs, because we use `spack external find --scope=system`. I will look into other options (`scope=site`, for example).

climbfuji:
I fixed this in 2e91ad6, a17b244, and d42cf47 by adding one more step to the NewSiteConfig instructions (unset `SPACK_DISABLE_LOCAL_CONFIG` when setting `SPACK_SYSTEM_CONFIG_PATH`, and restoring `SPACK_DISABLE_LOCAL_CONFIG` when unsetting `SPACK_SYSTEM_CONFIG_PATH`). Same for GitHub actions CI runs.

I also removed the exception to skip avoiding ~/.spack for macOS. We needed this in the past when bootstrapping was a tedious and manual process, but fortunately this is no longer the case. I tested this on the macOS CI runner manally. Hopefully, it will work just as well in CI.

srherbener (approved):
I think this looks good. I did not test, rather I'm relying on the CI testing and others. I like that by default user configuration is ignored for the build, and only for customized (new site config) builds you can temporarily enable user configuration. Thanks!

PR #1567, opened 2025-03-14
release/1.9.0: Update .gitmodules for spack: release/1.9.0 --> spack-stack-1.9.1

climbfuji:
### Summary

All in the title.

### Testing

- [ ] CI **RUNNING**
- [x] Verified that the spack submodule hash and code is identical

### Applications affected

None

### Systems affected

None

### Dependencies

None

### Issue(s) addressed

Working towards #1485 

### Checklist
- [ ] ~~These changes have been tested on the affected systems and applications.~~

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,7 +1,7 @@
 [submodule "spack"]
   path = spack
   url = https://github.com/jcsda/spack
-  branch = release/1.9.0
+  branch = spack-stack-1.9.1
 [submodule "doc/CMakeModules"]
   path = doc/CMakeModules
   url = https://github.com/noaa-emc/cmakemodules


AlexanderRichert-NOAA:
Would it be better to point to a branch rather than a tag? That way it doesn't have to be retagged for updates.

climbfuji:
> Would it be better to point to a branch rather than a tag? That way it doesn't have to be retagged for updates.

We've always done it this way, and it makes it clear that this is the release where tags point to tags and not moving targets.

climbfuji:
Besides, operations at least at NRL requires that.

AlexanderRichert-NOAA (approved):

stiggy87 (approved):

PR #1571, opened 2025-03-19
Acorn updates, March 2025

### LINKED ISSUE #1570: 'spack stack create env' automatically filter unused compilers from site/compilers.yaml?
Spack does not provide a clean way to restrict the compilers used in a given environment. I keep running into situations where for example on Acorn, some Intel 2022 packages creep into my Intel 19 environment (this has been happening in the weekly builds, so the build cache is a mess of intermixed Intel versions). I was planning to add a utility (mainly for use in weekly builds) to modify site/compilers.yaml (or any compilers.yaml or spack.yaml) to filter compilers in an inclusive or exclusive way, for instance, "remove all intel versions except the specified one." It occurred to me that that might have some broader use, so if it would be a helpful default behavior in `spack stack create env`, I'm happy to add it. For instance, if a user does `spack stack create env --compiler intel@19.1.3.304`, then it could automatically modify site/compilers.yaml to remove all other Intel versions.

AlexanderRichert-NOAA:
### Summary

General spack-stack updates:
 - 'esmf+python' as 'require' rather than 'prefer'.
 - By default, `spack stack create env` will now filter site/compilers.yaml to remove compilers with the same name but different versions (there is an optional flag for _not_ doing this). There is also a new utility, filter_spack_compilers.py that allows users to filter compilers "manually" (inclusive, exclusive, and exclusive by compiler name/version).
 - In weekly builds, don't wipe out existing env dirs. I'm openminded as to whether this is actually a good idea.
 - Allow mapl duplicates in weekly builds.

Acorn-specific updates:
 - Enable MPI-based netcdf-{c,fortran} tests on Acorn.
 - Re-enable PBS Pro for weekly testing; build `gh`+deps on login nodes, everything else in PBS Pro.

### Testing

Tested on Acorn.

### Applications affected

Most (because of `esmf:require:+python`)

### Systems affected

Acorn; maintainers that don't want to filter site/compilers.yaml can use the `--keep-all-compilers` flag

### Dependencies

https://github.com/JCSDA/spack/pull/527

### Issue(s) addressed

Fixes #1570

### Checklist
- [x] This PR addresses one issue/problem/enhancement, or **has a very good reason for not doing so**.


rickgrubin-noaa:
Is the following a correct understanding?

- when creating an env as is currently done: `spack stack create env ... --compiler <name>`, it will now be necessary for the `--compiler` arg to specify a `<name>` **_and_** `<version>`, e.g. `oneapi@2024` if there is more than one entry for `<name>` in `compilers.yaml` ?
- when creating an env as is currently done: `spack stack create env ... --compiler <name>` without specifying `--keep-all-compilers`, and multiple versions of `<name>`, e.g. `oneapi@2024` and `oneapi@2025`, are both specified in `compilers.yaml`, then `<name>@<version>` **_not_** specified as an arg to `--compiler` is filtered ?
- if `--keep-all-compilers` is specified when creating an env, one can filter out a compiler after the fact with the `util/filter_spack_compilers.py` utility ?

AlexanderRichert-NOAA:
>     * when creating an env as is currently done: `spack stack create env ... --compiler <name>`, it will now be necessary for the `--compiler` arg to specify a `<name>` **_and_** `<version>`, e.g. `oneapi@2024` if there is more than one entry for `<name>` in `compilers.yaml` ?
No- If no version is specified by the --compiler flag, site/compilers.yaml will not be modified.
>     * when creating an env as is currently done: `spack stack create env ... --compiler <name>` without specifying `--keep-all-compilers`, and multiple versions of `<name>`, e.g. `oneapi@2024` and `oneapi@2025`, are both specified in `compilers.yaml`, then `<name>@<version>` **_not_** specified as an arg to `--compiler` is filtered ?
Correct (assuming that some version was provided with the --compiler flag)
>     * if `--keep-all-compilers` is specified when creating an env, one can filter out a compiler after the fact with the `util/filter_spack_compilers.py` utility ?
Correct

rickgrubin-noaa (approved):

spack-ext/lib/jcsda-emc/spack-stack/stack/stack_env.py
```diff
@@ -218,6 +219,22 @@ def get_upstream_realpaths(self, upstream_path):
                 upstream_paths += self.get_upstream_realpaths(entry[1]["install_tree"])
         return upstream_paths
 
+    def filter_site_compilers(self, compilers_yaml_path, compiler_to_keep):
+        compiler_to_keep = re.sub("@=", "@", compiler_to_keep)
+        config_compiler_name = re.sub("@.*", "", compiler_to_keep)
+        with open(compilers_yaml_path, "r") as f:
+            compilers_yaml = syaml.load_config(f)
+        print(compilers_yaml)
```
climbfuji: Remove this print statement?

util/weekly_build/03_SetupEnv.sh
```diff
@@ -15,14 +15,15 @@ for compiler in $COMPILERS; do
...
+    if [ ! -d $envdir ]; then
+      spack stack create env --name $envname --template $template --site $PLATFORM --compiler $compiler
+    fi
     cd $envdir
     spack env activate .
     spack config add "config:install_tree:padded_length:${PADDED_LENGTH:-200}"
     # Check for duplicates and fail before doing the "real" concretization:
     spack_wrapper log.concretize concretize --fresh
-    ${SPACK_STACK_DIR:?}/util/show_duplicate_packages.py log.concretize -d -i crtm-fix -i crtm -i esmf
+    ${SPACK_STACK_DIR:?}/util/show_duplicate_packages.py log.concretize -d -i crtm-fix -i crtm -i esmf -i mapl
```
climbfuji, line 26: This is already in release/1.9.0. Maybe we should merge the release branch back to develop first?

util/weekly_build/03_SetupEnv.sh
```diff
@@ -15,14 +15,15 @@ for compiler in $COMPILERS; do
     [[ ${compiler} == *"@="* ]] && envname=build-$template-${compiler/@=/-} || envname=build-$template-${compiler/@/-}
     envdir=$RUNDIR/$RUNID/envs/$envname
     echo "Setting up environment $envname in $envdir"
-    rm -rf $envdir
-    spack stack create env --name $envname --template $template --site $PLATFORM --compiler $compiler
+    if [ ! -d $envdir ]; then
```
climbfuji, line 18: I don't mind, either way works for my setup

climbfuji (changes requested):
Two questions/requests
1. Please update the documentation. We have documentation for utilities, for the spack-stack extension, ... - all in readthedocs. Given that this is a little more complicated than some of the other options, we should document it properly.
2. Do we need the extra utility `filter_spack_compilers.py`? I'd say it's enough to do this when setting up the environment, and if you forget then start over?

util/weekly_build/03_SetupEnv.sh
AlexanderRichert-NOAA: Either way is fine with me

AlexanderRichert-NOAA:
1. I added a couple bits of documentation, let me know if there's somewhere else that it should get mentioned or if you think it needs further elaboration.
2. Probably not? I intend to maintain a copy of it elsewhere for use outside of spack-stack, so I won't add it unless others would find it useful.

climbfuji (approved):
Happy to merge this and resolve the conflicts later when merging release/1.9.0 into develop.

PR #1580, opened 2025-03-25
Feature/update nvhpc config

### LINKED ISSUE #1563: Next steps for Nvidia compiler instructions
In PR #1542 (merged into `release/1.9.0`, _not_ `develop`) I updated and reformatted the nvidia-specific instructions.

There is now a template `configs/templates/jedi-mpas-nvidia-dev/spack.yaml` which describes a subset of the `jedi-mpas-env` virtual env, along with some notes about what was cut out and why.

There is also a site-config `configs/sites/tier2/ubuntu2404-nvhpc` which indicates what to install from `apt` and points to those site packages.

I believe next steps would be (but up for discussion)
- figure out why JEDI compiled with this environment fails to load `cublas` at runtime (@l90lpa will open an issue on this with details)
- [perhaps related to above] see if we can point spack to use `cublas` and `cufft` instead of installing `openblas` and `fftw`?
- make better use of gcc as a fallback compiler instead of resorting to the OS package manager; this would enable a more portable configuration
  - my attempts to set up gcc as fallback compiler led to many concretization issues, but presumably this is user error on my part
  - note that we _must_ use nvidia for _all_ fortran packages (technically, all packages that provide modules that will be imported by JEDI; this is because fortran has no "abi compatibility" thus module files are not portable across compilers), and probably all packages using MPI as well; this impacts certain components that may or may not build with nvidia (`gsibec`, ...); thus it may or may not be possible to build the entire `jedi-mpas-env` _even with ideal use of fallback compilers_.
- and surely any number of cleanups from a spack(-stack) expert point of view!

I'm opening this issue to give @stiggy87 some pointers on current state, and some ideas of where we could take this next.

stiggy87:
### Summary

Update to the site and template for the `jedi-mpas-nvidia-dev` feature.

### Testing

This has been done on an AWS GPU instance (g5.4xlarge) following instructions from @l90lpa. 

### Applications affected

This will only affect the JEDI MPAS Nvidia compiler work that @fmahebert completed.
 
### Dependencies

N/A

### Issue(s) addressed

Fixes #1563 

### Checklist


configs/sites/tier2/ubuntu2404-nvhpc/packages_nvhpc.yaml
```diff
@@ -0,0 +1,18 @@
...
+  mpi:
+    buildable: False
+  blas:
+    buildable: False
+  fckit:
+    require: '~openmp'
+  eckit:
+    require: '~openmp'
+  ecmwf-atlas:
+    require: '~openmp ~fftw'
```
climbfuji:
```suggestion
    require: '~openmp ~fftw'

```

configs/templates/jedi-mpas-nvidia-dev/spack.yaml
```diff
@@ -21,47 +21,4 @@ spack:
...
-  - udunits
-
-  # packages from jedi-mpas-env
-  # removed jasper
-  - metis
-
-  # added packages
-  - netcdf-cxx4
-  - zlib
+  - ectrans
```
climbfuji:
```suggestion
  - ectrans

```

climbfuji (changes requested):
It looks like you are only installing package `ectrans`?

stiggy87:
> It looks like you are only installing package `ectrans`?

@climbfuji Good question. I've asked, and @fmahebert told me majority of the work around this compiler is focused on `ectrans`. I am all for just adding it to the list and leaving the remaining packages there.

I've started a dialog with @fmahebert and  @l90lpa to get some more clarification.

stiggy87:
I pulled too much out for testing! I've added the original packages back into the spec and added `ectrans` to it.

configs/sites/tier2/ubuntu2404-nvhpc/README.md
```diff
@@ -80,19 +106,22 @@ cd envs/nvidia-env/
...
 
 ```bash
-module use ${SPACK_STACK_DIR}/envs/jedi-mpas-nvidia-env/install/modulefiles/Core
+module use /opt/nvidia/hpc_sdk/modulefiles
+module use ${SPACK_STACK_DIR}/envs/nvidia-env/install/modulefiles/Core
+
+module load nvhpc/25.1
+module load stack-nvhpc/25.1
+module load stack-openmpi/4.1.5
+module load ectrans/1.5.0
```
fmahebert: To keep ectrans on the same footing as all the other modules, perhaps remove it from this listing? No strong feelings on this from me.

configs/sites/tier2/ubuntu2404-nvhpc/packages_nvhpc.yaml
```diff
@@ -0,0 +1,18 @@
+packages:
+  all:
+    compiler:: [nvhpc@25.1, gcc@13.3.0]
+    providers:
+      mpi:: [openmpi@4.1.5]
+      blas:: [nvhpc@25.1]
+      zlib-api:: [zlib]
+      fftw:: [nvhpc@25.1]
```
fmahebert, line 8: For my education...
- I see you fixed the MPI version, thanks!
- You added `fftw` from `nvhpc` here but then specify `ecmwf-atlas ~fftw` below; how come? Does Nvidia's fft library not satisfy atlas' requirements?
- What changed here to allow `cublas` and `cufft` to be found at runtime? Was this about the drivers, or something else in the spack-stack setup?

configs/templates/jedi-mpas-nvidia-dev/spack.yaml
```diff
@@ -21,47 +21,4 @@ spack:
   include: []
 
   specs:
-
-  # packages from base-env
-  # removed python
-  # removed py-* packages
-  - hdf5
```
fmahebert: You say you added the packages back to the spec, but I still see a lot of deletions? Missing a commit perhaps?

configs/sites/tier2/ubuntu2404-nvhpc/packages_nvhpc.yaml
stiggy87: - Good catch on the `~fftw` I am going to remove that. It was caught when I was looking at the specs to determine what was causing concretization failures. 
- The `cublas` and the `cufft` I think was twofold, the driver and that we need to be explicit and setting the `fftw` and the `blas` to be provided by `nvhpc`.

fmahebert: Thanks for the explanations!

configs/sites/tier2/ubuntu2404-nvhpc/README.md
stiggy87: I can remove it, that way it's more agnostic to general flows.

configs/templates/jedi-mpas-nvidia-dev/spack.yaml
stiggy87: Forgot to push.

eap:
This is more of a question for @fmahebert on requirements; Should we consider adding in `jedi-fv3-env`, `soca-env` and `ewok-env`? It would be nice to be able to run skylab small experiments with this, although (1) I don't know if that's outside the scope of the needs here and (2) maybe that would be better added in a followup change?

Thoughts? At minimum it would be nice to get `ewok-env +ecflow ~cylc` in this change since we would immediately have ecflow and could run some simple experiments (probably?)

fmahebert:
@eap I (selfishly) view this PR's goal to be fixing issues in the current nvhpc spack-stack environment that are blocking our short-term code demonstration goals. Thus, I would recommend against the scope expansion of adding any new packages to the build as part of this PR.

To expand a bit — the demonstrations we're targeting will not involve skylab, workflows, or any of the soca/fv3 related code. Thus, adding the packages you've suggested is not immediately necessary, and risks introducing delays if we discover anything that doesn't build cleanly with nvhpc. In the longer term, it would certainly be nice to have more support for nvhpc across the system by working towards what you suggest. But I view that as a long-term project of building support (as and when warranted by JCSDA/partner needs), and beyond the scope of this PR.

Does that sound reasonable?

eap (approved):
@fmahebert  > Absolutely reasonable, I just wanted to ask the question. Adding in ewok and other skylab dependencies would be a good followup but it looks like this is working as intended for the demo.

climbfuji (approved):

fmahebert (approved):
Thanks for your help on this @stiggy87 !

PR #1601, opened 2025-04-08
show_duplicate_packages.py: use spack.lock instead of 'spack concretize' output

AlexanderRichert-NOAA:
### Summary

This PR updates show_duplicate_packages.py to use spack.lock to detect duplicates rather than `spack concretize` stdout. It is safer (especially if an environment has been concretized in multiple steps), it simplifies the code, and it's cleaner. It removes the -d option (it now only shows duplicates), and the -c option (which at this point we don't really have any use for because we're doing one-compiler environments).

### Testing

Tested on personal machine (+updating util tests).

### Applications affected

n/a

### Systems affected

all

### Dependencies

none

### Issue(s) addressed

none

### Checklist


AlexanderRichert-NOAA:
This is good to go assuming the remaining CIs succeed

doc/source/Utilities.rst
```diff
@@ -9,16 +9,14 @@ Miscellaneous utilities
...
-The utility located at util/show_duplicate_packages.py parses the output of ``spack concretize`` and detects duplicates. Usage is as follows:
+The utility located at util/show_duplicate_packages.py parses ``spack.lock`` and detects duplicates. Usage is as follows:
 
 .. code-block:: console
 
-   spack concretize | ${SPACK_STACK_DIR}/util/show_duplicate_packages.py
-   # - OR -
-   spack concretize |& tee log.concretize
-   ${SPACK_STACK_DIR}/util/show_duplicate_packages.py log.concretize
+   # In $SPACK_ENV, after concretization:
```
climbfuji: I think this is misleading - you don't need to be in the directory ${SPACK_ENV}, but you need to have an environment activated so that ${SPACK_ENV} is defined?

AlexanderRichert-NOAA: Good catch, thanks

climbfuji (approved):

Issue #1398, opened 2024-12-06
Fix up UWM testing logic in automated builds


AlexanderRichert-NOAA:
Right now, test_ufswm.sh requires using a fork of ufs-weather-model in order to point to the modules to be tested. We should be able to use some search-and-replace on the modulefiles so that we can use the head of develop in the authoritative repo.

climbfuji:
Fixed.

AlexanderRichert-NOAA:
@rickgrubin-noaa where does this happen in the weekly build scripts as far as pointing ufs-weather-model to the new modules?

rickgrubin-noaa:
> [@rickgrubin-noaa](https://github.com/rickgrubin-noaa) where does this happen in the weekly build scripts as far as pointing ufs-weather-model to the new modules?

The setting for [BRANCH](https://github.com/JCSDA/spack-stack/blob/develop/util/weekly_build/apptests/test_ufswm.sh#L9) handles differing modules without having to change automation scripts.

AlexanderRichert-NOAA:
The idea here is to be able to point test_ufswm.sh to the ufs-weather-model head of develop and have it run with the newly installed modules on any system. Re-opening the issue.
I think the approach here is to either have our weekly build logic build UWM modulefiles for a given system, or, probably more simply, do a search-and-replace on the path, and set appropriate environment variables to set each package version.

rickgrubin-noaa:
> The idea here is to be able to point test_ufswm.sh to the ufs-weather-model head of develop and have it run with the newly installed modules on any system. Re-opening the issue. I think the approach here is to either have our weekly build logic build UWM modulefiles for a given system, or, probably more simply, do a search-and-replace on the path, and set appropriate environment variables to set each package version.

`test_ufswm.sh` does point to `ufs-weather-model@develop` by default. Not setting any env variables would "run with the newly installed modules" on a system.

To me, it seems more work to modify scripts or CI/CD, but that said, no worries, as you see fit.

AlexanderRichert-NOAA:
I don't follow-- how would ufs-weather-model know to look at the newly installed modules for a given system? Unless I'm missing something then I would predict that running ./rt.sh would lead it to automatically infer the platform and load the modules present in develop from modulesfiles/*lua. Can you walk through how running test_ufswm.sh would run ufs-weather-model/rt.sh in a way that it knows to point to the packages being tested?

rickgrubin-noaa:
[...]

PR #1603, opened 2025-04-11
Stack automation / testing enhancements / bug fixes

### LINKED ISSUE #1510: malformed return code in util/weekly_build/apptests/test_ufswm.sh
**Describe the bug**
util/weekly_build/apptests/test_ufswm.sh (lines 20 and 21)

This bash script attempts to return command status as rc
but should return $rc
also the assignment of rc is malformed:

rc = $?
return rc

should be changed to:

rc=$?
return $rc

**To Reproduce**
1. running the script with valid args
2. an error is reported regarding the return code

**Expected behavior**
return code should be a standard bash return integer:  0, or non-zero

**System:**
Orion

**Additional context**
none

### LINKED ISSUE #1511: for automation tests, allow override of TEST_UFSWM for whether to execute tests
In util/weekly_build/06_AppTests.sh there is a check of TEST_UFSWM , = ON or OFF, which controls whether to execute apptests/test_ufswm.sh

if [ "$TEST_UFSWM" == ON ]; then
  . $(dirname $0)/apptests/test_ufswm.sh
fi

However, when sourcing the sites/my_machine.sh (e.g. orion.sh), this value is hardcoded as:

TEST_UFSWM=ON

Should be change to allow override:

TEST_UFSWM=${TEST_UFSWM:-"ON"}

**Additional context**
This allows for test execution options to be specified within CICD (Jenkins) jobs.

### LINKED ISSUE #1512: automation scripts' use of dirname $0 doesn't work in Jenkins jobs
**Describe the bug**
the bash scripts that use command  $(dirname $0) to derive a full path for execution does not work as desired within Jenkins job's execution frames.
In Jenkins jobs, an execution frame to run bash commands creates a temporary script in which to execute the commands. This script is unbound from the workspace, and lives on the host in /tmp/...

So $(dirname $0) always returns "/tmp", which is not the intended path wanted for these scripts:

* util/weekly_build/SpackStackBuildCache_AllSteps.sh (line 7)
* util/weekly_build/ShellSetup.sh (line 36)
* util/weekly_build/06_AppTests.sh (line 25)

**To Reproduce**
Run util/weekly_build/ShellSetup.sh within a Jenkins job

**Expected behavior**
executing any of the above scripts in a Jenkins job should find the full path of the target.

**System:**
Orion

**Additional context**
Perhaps add a top WORK_DIR global variable that can be referenced instead.

rickgrubin-noaa:
### Summary

This PR addresses issues that impact end-to-end automated build / test of the stack within CI/CD frameworks like Jenkins (used within EPIC).

### Testing

End-to-end run within a Jenkins execution frame on host orion were successful.

### Applications affected

n/a

### Systems affected

None affected.

Changes allow for removing in-place editing of files to achieve the results that this PR will add.

### Dependencies

n/a

These changes will lead to addressing issue #1398 as well as running multiple tests for `UFS-WM` (rather than the singular `control_c48` at the moment)

### Issue(s) addressed

Fixes #1510 
Fixes #1511 
Fixes #1512 

NOTE: the change to `configs/sites/tier1/orion/compilers.yaml` and `configs/sites/tier1/orion/compilers.yaml` are unrelated; they remove unnecessary env settings for compilers.

### Checklist


AlexanderRichert-NOAA:
This looks good- Does this mean that UWM RTs are part of the scheduled runs on Orion?

rickgrubin-noaa:
> This looks good- Does this mean that UWM RTs are part of the scheduled runs on Orion?

I've been playing around outside of Jenkins (current lack of allocated hours / have to use the windfall allocation) with how to best (read: most simply) run multiple UWM RTs, and then will integrate that into the orion site-specific script.

climbfuji:
Not sure why these GitHub action runs failed. This PR doesn't affect them at all

util/weekly_build/06_AppTests.sh
```diff
@@ -21,6 +21,6 @@ for compiler in $COMPILERS; do
   done
 done
 
-if [ "$TEST_UFSWM" == ON ]; then
-  . $(dirname $0)/apptests/test_ufswm.sh
+if [ "$TEST_UFSWM" == "ON" ]; then
+  . ${WORK_DIR:-$(dirname $0)}/apptests/test_ufswm.sh
```
climbfuji, line 25: Should we standardize this variable (i.e. document that this is an option that can be set)?

Since WORK_DIR or WORKDIR is a common name that some HPC centers or users may already  use, should we be more generic here and call it something else? Fine to merge as is and fix later when it becomes a problem.

AlexanderRichert-NOAA: Different question-- Is the idea of the `${WORK_DIR:-$(dirname $0)}` logic for if you want to run a copy of test_ufswm.sh that's not the same as the one that lives in the directory of the 06_AppTests.sh being run?

rickgrubin-noaa: `WORK_DIR` (or `WORKDIR`), as a named variable, can be named to whatever seems most appropriate for the purposes of automation / test; open to suggestions and happy to change it.

The purpose of it is primarily for separate CI/CD tools; in this particular case, for Jenkins (see [issue 1512](https://github.com/JCSDA/spack-stack/issues/1512)). Per your question, it could also be used in that manner (side effect I hadn't considered).

climbfuji (approved):
I am ok with this as is, too

AlexanderRichert-NOAA (approved):

Issue #1574, opened 2025-03-22
spack mirror create doesn't resolve spec constraints correctly


AlexanderRichert-NOAA:
In the weekly build scripts, the `spack mirror create` command in 04_SpackInstall.sh does not have the correct behavior when installing only a subset of the environment's root specs. In particular, even though the unified-dev template specs 'ufs-weather-model ^esmf@8.8.0', `spack mirror create`, at least on Acorn, is fetching esmf@8.6.1 instead. This is true even when I specify only `ufs-weather-model` or `ufs-weather-model ^esmf@8.8.0` for the arguments to `spack mirror create`. For now I intend to work around it by adding the option to disable fetching there, then I'll use `spack fetch` as part of Acorn's spack_install_wrapper(), but I'll leave this issue here for discussion in case anyone else is affected. For anyone using the default `--all` (maybe we should have an option for fetching all even if not all packages will be installed?), this shouldn't be a problem. I don't know if anyone else plans to only install a subset of unified-dev.

AlexanderRichert-NOAA:
@climbfuji FYI. I'll put in an Acorn-specific PR for this shortly.

AlexanderRichert-NOAA:
Fixed

PR #1618, opened 2025-04-25
Add scripts to download cargo dependencies and store in local cargo registry for offline installation

climbfuji:
### Summary

This PR adds scripts to fetch cargo dependencies into a local registry for cached or offline installation, and for downloading and installing `rust`/`cargo` if necessary. Credits to @AlexanderRichert-NOAA for the initial version of the fetch-cargo logic.

The new logic is also incorporated in the utility `utils/nrl/batch_install.sh` to demonstrate how they fit into the sequence of steps.

### Testing

Tested locally on my dev box and on Blueback with `utils/nrl/batch_install.sh`.

### Applications affected

None

### Systems affected

None

### Dependencies

None

### Issue(s) addressed

Working towards https://github.com/JCSDA/spack-stack/issues/1597

### Checklist
- [x] This PR addresses one issue/problem/enhancement, or has a very good reason for not doing so.
- [x] These changes have been tested on the affected systems and applications.
- [ ] ~~All dependency PRs/issues have been resolved and this PR can be merged.~~

PR #1623, opened 2025-04-29
Acorn updates; go fetcher

### LINKED ISSUE #1597: Packages reaching out to the WWW during spack install
This is more a question than a request for a fix.

`global-workflow-env` depends on `gh` which itself depends on `go`. The latter reaches out to the internet during installation. In the past, we made efforts to avoid this for installing on air-gapped systems.

Similar problems are found with recent Python stacks that depend on `rust` (or `rust-bootstrap`) and then download internal dependencies as part of their builds (e.g. `py-maturin`). As opposed to `global-workflow-env`, this is a more common problem because several newer versions of Python packages depend on `py-maturin` or use `rust`. Replacing `rust` with `rust-bootstrap` does not solve the problem (see https://github.com/spack/spack/discussions/45483 for more information).

### LINKED ISSUE #1615: Add OneAPI 2024.2.1 support on Acorn
OneAPI v2024.2.1 was recently installed on Acorn and the site configs should be updated to include support for this compiler suite.

The ability to build and use the unified-dev environment on Acorn.

**Additional context**
Part of a larger push to 1. move to oneapi on all platforms and 2. maintain spack-stack on WCOSS2.

### LINKED ISSUE #1616: Bring back platform-specific logic in setup.sh?
Is there a reason we got rid of the platform-specific logic (we had it for Orion and Acorn I believe) in setup.sh for loading modules, etc.? It facilitates consistency, and I just spent half an hour debugging a Python-related issue that would have been avoided if setup.sh had automatically set up env modules for me. It was an optional argument, so it doesn't impact anyone who doesn't want to use it. It would also be helpful on Acorn for setting TMPDIR in a way that accommodates login node and compute node installations.

AlexanderRichert-NOAA:
### Summary

This PR:
 - adds oneAPI 2024.2.1 on Acorn
 - adds util/fetch_go_deps.py
 - adds platform-specific logic for setup.sh (for loading modules, etc.)

### Testing

Successfully built unified env on Acorn.

### Applications affected

UWM, GW, GSI

### Systems affected

go fetcher can be used wherever, and site maintainers may choose to add site-specific logic in setup.sh; otherwise no expected impacts to platforms other than Acorn

### Dependencies

none

### Issue(s) addressed

Fixes #1615
Fixes #1616 
Fixes #1597 
Addresses #1574 by adding workaround for Acorn

### Checklist


AlexanderRichert-NOAA:
FYI @DavidHuber-NOAA

AlexanderRichert-NOAA:
@climbfuji I was toying with the idea of tweaking the compiler filtering logic for `spack stack create env` so that I can just give it a list of the compiler specs that I want to keep. The invocation would be something like:
```console
spack stack create env --template unified-dev --name test --site acorn --compiler oneapi@2024.2.1 --keep-compilers oneapi@2024.2.1,gcc@12.1.0
```
and the default would go back to the original, i.e., no compiler filtering. In any case, I want a better way to specify exactly what I want so that I don't have, say, `intel` creeping into my `oneapi` environment.

If that sounds good I can do either in this PR or in a separate one.

climbfuji:
> @climbfuji I was toying with the idea of tweaking the compiler filtering logic for `spack stack create env` so that I can just give it a list of the compiler specs that I want to keep. The invocation would be something like:
> 
> ```
> spack stack create env --template unified-dev --name test --site acorn --compiler oneapi@2024.2.1 --keep-compilers oneapi@2024.2.1,gcc@12.1.0
> ```
> 
> and the default would go back to the original, i.e., no compiler filtering. In any case, I want a better way to specify exactly what I want so that I don't have, say, `intel` creeping into my `oneapi` environment.
> 
> If that sounds good I can do either in this PR or in a separate one.

A separate PR is a good idea imo

climbfuji:
#1618 was merged, please resolve the conflicts and mark this PR as ready/add reviewers and I will look at it. Thanks!

doc/source/Utilities.rst
```diff
@@ -51,14 +51,13 @@ The util/parallel_install.sh utility runs parallel installations by launching mu
...
+The fetch_cargo_deps.py and fetch_go_deps.py utilities prefetch Rust/Cargo and Go dependencies, respectively, storing them in a local directory for later use during ``spack install``. This is required for installing on systems that do no have access to the internet during the ``spack install`` phase and complements other mirrors such as the spack source mirror. The ``$CARGO_HOME`` and ``$GOMODCACHE`` variables must be set, and the utilities must be run in an active, concretized environment. They will fetch Spack packages of type CargoPackage (or PythonPackage with ``rust`` dependency) and GoPackage, respectively, and fetch all dependencies based on the dependency listings found in those packages (e.g., Cargo.toml). The utilities will attempt to use each package's ``rust`` or ``go`` dependency for the appropriate executable (``cargo``, ``go``), in which case these utilities must be run after ``rust``/``go`` are installed. They will revert to using system-installed ``rust``/``go`` if available. For ``fetch_cargo_deps.py``, if a ``cargo`` executable cannot be found, it will run ``install_rust.sh`` to install it.
```
climbfuji: Tiny correction, for cargo I had to remove using each packages' `cargo` because (a) there is little use for this case (normally you do this cargo fetching before you install) and (b) I got errors for some packages with that logic.

setup.sh
```diff
@@ -2,6 +2,13 @@
 # Portable way to get current directory
 SPACK_STACK_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 
+hostname=$(hostname)
+case $hostname in
+  alogin*)
```
climbfuji, line 7: That hostname is very unfortunate. I can totally see other platforms using something like this, too ... but for now we don't have any other one.

AlexanderRichert-NOAA: We can use `hostname --fqdn`. I'll be out the next couple days but I'll update this when I get the chance. With that it can be alogin*ncep.noaa.gov or something like that.

climbfuji: We can add that when we need it - for now we're ok.

doc/source/Utilities.rst
climbfuji: I think we should revert this and split it into two sections. I'll provide a suggestion shortly.

climbfuji: https://github.com/AlexanderRichert-NOAA/spack-stack/pull/6

climbfuji: Took the liberty to merge this since you are out for a few days.

climbfuji (approved):

PR #1639, opened 2025-05-13
Add util/check_libirc.sh to check for libirc.so used in spack-stack environments

climbfuji:
### Summary

This PR adds a utility `util/check_libirc.so` that checks shared libraries and executables in a spack-stack environment for links to `libirc.so`. The documentation adds more details on why this is needed and when/how to use it, and I added the tool to the existing GitHub actions tests. The purpose of this utility is to catch issues with spack-stack environments only. 

I also created a wiki page that will provide details on how to avoid linking to `libirc.so` in downstream applications: https://github.com/JCSDA/spack-stack/wiki/Intel-oneAPI-compilers-and-libirc.so

The newly added check found `libirc.so` in several `met` binaries. I fixed that and improved the existing fix for `parallel-netcdf` in https://github.com/JCSDA/spack/pull/543.

Sneaking in an unrelated, tiny bug fix for a missing shared library path in `configs/sites/tier1/atlantis/compilers.yaml`.

### Testing

- [x] CI

### Applications affected

Unified environment.

### Systems affected

None

### Dependencies

- [x] waiting on https://github.com/JCSDA/spack/pull/543

### Issue(s) addressed

Related to https://github.com/JCSDA/spack-stack/issues/1436

### Checklist


AlexanderRichert-NOAA:
When you say adds the capability-- what is it that triggers using these templates/config files?

climbfuji:
> When you say adds the capability-- what is it that triggers using these templates/config files?

the meta module for the compiler

climbfuji:
> > When you say adds the capability-- what is it that triggers using these templates/config files?
> 
> the meta module for the compiler

To be clear, this only affects the spack-stack environments when USING them. I confirmed that with the bug fixes we put in place in our spack fork (parallel-netcdf), there is no single package that links to `libirc.so` anymore.

But we could add `-shared-intel` to the SPACK_*_ALWAYS_FLAGS in `spack/lib/...` to be on the safe side there, too.

AlexanderRichert-NOAA:
My main concern is making sure it doesn't conflict with existing configurations on each system. I don't see any flags configured in {icx,icpx,ifx}.cfg on Hera or Gaea (confirmed with `icx -v |& grep -oP "Configuration file: \K.+" | xargs cat`), but on Hercules and Orion, the ones that live in the bin/ dirs of each installation are not empty. They appear to be customized by Spack, which sets a bunch of rpath's as well as the gcc toolchain. Would it make sense to incorporate those into our custom {icx,icpx,ifx.cfg}?

stiggy87 (approved):

climbfuji:
> My main concern is making sure it doesn't conflict with existing configurations on each system. I don't see any flags configured in {icx,icpx,ifx}.cfg on Hera or Gaea (confirmed with `icx -v |& grep -oP "Configuration file: \K.+" | xargs cat`), but on Hercules and Orion, the ones that live in the bin/ dirs of each installation are not empty. They appear to be customized by Spack, which sets a bunch of rpath's as well as the gcc toolchain. Would it make sense to incorporate those into our custom {icx,icpx,ifx.cfg}?

Oh. I've never come across a system that used a customized cfg file. I guess that's a spack thing.

How do you want to include that? Automatically or manually/hardcoded for each system?

climbfuji:
> > My main concern is making sure it doesn't conflict with existing configurations on each system. I don't see any flags configured in {icx,icpx,ifx}.cfg on Hera or Gaea (confirmed with `icx -v |& grep -oP "Configuration file: \K.+" | xargs cat`), but on Hercules and Orion, the ones that live in the bin/ dirs of each installation are not empty. They appear to be customized by Spack, which sets a bunch of rpath's as well as the gcc toolchain. Would it make sense to incorporate those into our custom {icx,icpx,ifx.cfg}?
> 
> Oh. I've never come across a system that used a customized cfg file. I guess that's a spack thing.
> 
> How do you want to include that? Automatically or manually/hardcoded for each system?

@AlexanderRichert-NOAA I found another way that takes into account existing settings on each platform. I also updated the PR description. Can you take another look, please?

AlexanderRichert-NOAA:
I'm not familiar with those __INTEL variables but I don't immediately see any issue there.

As far as handling the existing *.cfg files, I'm not aversing to ignoring it for now, but if it caused issues then maybe we would want to have it automatically get the path to the existing one and copy that into the spack-stack one.

climbfuji:
I discovered the `__INTEL_PRE_*` flags while reading through the manpages. This approach is safer than potentially overwriting the existing config files. I did try merging config files using the command you posted in a previous comment, but `ifx` doesn't show the location of the config file when using `-v` (only `icx` and `icpx` do).

AlexanderRichert-NOAA:
Ah okay, now I see that's instead of not in addition. Yeah if those variables work then that's great.

AlexanderRichert-NOAA (approved):

climbfuji:
Sigh. We hit a snag here. The C compiler `icx` doesn't understand `-shared-intel`, but the C++ compiler `icpx` does. This is very inconvenient, since the C and C++ compilers share the same environment variable `__INTEL_PRE_CFLAGS`.

The previous approach that used `ICXCFG` etc wasn't great either, because there is no way to extract the currently set `IFXCFG` from `ifx` (works only for `icx` and `icpx`).

I wonder if it would be better to simply provide instructions on how to solve this at the application level (when and where to add `-shared-intel`), and add a check to spack-stack that parses all lib*.so* files in an installed environment to make sure there's no linking to `libirc.so`?

Thoughts @AlexanderRichert-NOAA?

AlexanderRichert-NOAA:
Hmm, darn... What about `$(dirname $(which ifx))/ifx.cfg`?

climbfuji:
> Hmm, darn... What about `$(dirname $(which ifx))/ifx.cfg`?

But we don't know if everybody is modifying that config or using a customized config elsewhere. But I am ok with this approach `$(dirname $(which ifx))/ifx.cfg` as long as you are and as long as we describe this in the documentation.

climbfuji:
> > Hmm, darn... What about `$(dirname $(which ifx))/ifx.cfg`?
> 
> But we don't know if everybody is modifying that config or using a customized config elsewhere. But I am ok with this approach `$(dirname $(which ifx))/ifx.cfg` as long as you are and as long as we describe this in the documentation.

Also, that's going to be difficult to implement in `spack stack setup-meta-modules`, especially for Cray systems.

stiggy87 (approved):

climbfuji:
@RatkoVasic-NOAA @AlexanderRichert-NOAA @stiggy87 This is finally ready and working as expected. It needs https://github.com/JCSDA/spack/pull/543.

AlexanderRichert-NOAA (approved):

PR #1640, opened 2025-05-14
Update tier2 site blackpearl: oneapi@2024.1.2 --> oneapi@2025.1.0

climbfuji:
### Summary

Update tier2 site blackpearl: oneapi@2024.1.2 --> oneapi@2025.1.0. The newest version of `py-torch` gets confused on my laptop about which SIMD instructions are available; pin to the second-most recent version (the one we used before updating from spack develop about four weeks ago).

### Testing

Ran NRL batch install script for NEPTUNE environment on the machine.

### Applications affected

None

### Systems affected

Blackpearl

### Dependencies

If there are PRs that need to be merged before or along with this one, please add "- [ ] waiting on LINK_TO_PR" for each of them

### Issue(s) addressed

None

### Checklist


AlexanderRichert-NOAA (approved):
Just out of curiosity why is the CPATH setting not needed for the new version?

climbfuji:
> Just out of curiosity why is the CPATH setting not needed for the new version?

Looks like at some point in the recent past, this bug fix was no longer needed. Early versions of oneapi had this bug that they didn't set the include paths correctly.

PR #1798, opened 2025-10-23
[feature/update_to_spack_v1] Update NRL batch install script and NRL site configs; add flag to 'spack stack create env' to promote warnings to errors; updates from develop and release/1.9.0 branches

climbfuji:
## Description

1. Add argument `--treat-warnings-as-errors` for `spack stack create env` so that warnings for non-existent `packages_*.yaml` are promoted to errors. The default remains unchanged - they remain warnings.
2. Update NRL batch install script for spack v1 updates (mostly due to site config updates).
3. By default, do not create spack compiler modules (not to be confused with meta modules, we definitely do want those by default), use the native compiler modules on HPCs. For user site configs and CI, explicitly add compiler modules to module includelist in `configs/sites/**/modules.yaml`.
4. Bring in updates from spack-stack develop (which are updates from the latest release 1.9.3): site configs and wgrib2 variants. All site config updates except NRL sites are from this step.
5. Remove explicit variants for `ewok-env` in `configs/templates/skylab-dev/spack.yaml` and `configs/templates/unified-dev/spack.yaml`, since they are the default values. This allows us to turn off the `ecflow` variant in the site configs (c.f. Atlantis).
6. Add missing export of environment variables from MPI packages in MPI meta modules `setup-meta-modules`).

### Dependencies

- [x] Waiting on https://github.com/JCSDA/spack-packages/pull/12

### Issues addressed

Working towards #1604 

### Applications affected

None

### Systems affected

NRL systems

## Testing

- CI: _Note whether the automatic tests (GitHub actions tests that run automatically for every commit) pass or not_
    - [x] GitHub actions CI tests pass
    - [ ] GitHub actions CI tests do not pass (_provide explanation_)
    - [ ] GitHub actions CI tests skipped (_provide explanation if necessary_)
- New tests added: _List and describe any new tests added to GitHub actions_
    - [ ] ...
- Additional testing: _Add information on any additional tests conducted_
    - [ ] ...

## Checklist

- [x] This PR addresses one issue/problem/enhancement or has a very good reason for not doing so.
- [x] These changes have been tested on the affected systems and applications.
- [x] All dependency PRs/issues have been resolved and this PR can be merged.
- [ ] ~~All necessary updates to the documentation on readthedocs are included in this PR~~ deferred
    - For site config updates, check in particular `doc/source/PreConfiguredSites.rst` and `doc/source/MaintainersSection.rst`
- [ ] ~~All necessary updates to the spack-stack wiki will be made when this PR is merged~~ deferred

PR #1804, opened 2025-10-29
fetch_cargo_deps.py: ensure CARGO_HOME is used

AlexanderRichert-NOAA:
## Description

This PR ensures that CARGO_HOME is used (rather than the default of ~/.cargo) when fetching cargo deps, and also gets rid of the llnl.* deprecation warning for fetch_cargo_deps.py.

### Dependencies

none

### Issues addressed

none

### Applications affected

most

### Systems affected

all

## Testing

    - [ ] GitHub actions CI tests pass
    - [x] Acorn (manual test)

## Checklist

- [x] All necessary updates to the documentation on readthedocs are included in this PR
- [x] All necessary updates to the spack-stack wiki will be made when this PR is merged

### DIFF
--- a/util/fetch_cargo_deps.py
+++ b/util/fetch_cargo_deps.py
@@ -25,7 +25,7 @@ from spack_repo.builtin.build_systems.cargo import CargoPackage
 from spack_repo.builtin.build_systems.python import PythonPackage
 from spack.package_base import PackageBase
 from spack.util.executable import Executable, which
-from llnl.util.filesystem import working_dir
+from spack.llnl.util.filesystem import working_dir
 from spack.store import find
 from spack.util.environment import EnvironmentModifications
 from spack.error import SpackError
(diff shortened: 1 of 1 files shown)


util/fetch_cargo_deps.py
```diff
@@ -117,6 +117,7 @@
         cargo_exe = Executable(cargo_wrapper)
 
     env = EnvironmentModifications()
+    env.set("CARGO_HOME", cargo_home)
```
climbfuji, line 120: I am surprised this is needed, because of lines 44-47 above
```
cargo_home = os.getenv("CARGO_HOME")
if not cargo_home:
    raise SpackError("CARGO_HOME must be set")
```

AlexanderRichert-NOAA: Indeed. It seems that maybe there's something about the Executable logic that's not automatically propagating it? :shrug:

climbfuji: I noticed something similar with the wrapper script, `#!/usr/bin/env bash` doesn't work anymore. I had to change it to just `#!/bin/bash` in PR #1798

climbfuji (approved):

Issue #1772, opened 2025-09-09
Push-button deployment automation


AlexanderRichert-NOAA:
## Is your feature request related to a problem?

Per our 9/9/25 discussion, one key way we can help ourselves reduce spack-stack maintenance workload is to automate deployments as much as possible.

This issue is a place to further discuss needs and wants, especially from the spack-stack maintainers who do the lion's share of the deployments. Below I'll describe one possible general approach, but the key here is to, if possible, find an approach that's suitable for every partner organization that wants to get in on it (for example we could use an approach that makes more use of the existing weekly build-and-test scripting).

## A possible solution

Imagine running `spack stack deploy`, and it will automatically detect which system it's running on, retrieve which installations (template+compiler combinations) should be present on that system (based on, say, config/sites/tier?/*/deployments.yaml), and create and install those environments not yet created. Add-on environments could be deployed _either_ by adding templates (including any associated custom recipes) to the appropriate branch (necessitating a git pull for each deployment), _or_ by storing templates for add-on environments in a separate repository to be accessed as needed by `spack stack deploy` (therefore no need to git pull into an existing spack-stack directory). Both approaches would give us the agility and flexibility needed to avoid re-release he~~ll~~ck in most cases.

`spack stack deploy` options would include specifying/excluding certain template+compiler combinations, force re-installing of selected environments, --configure-only to set up but not deploy each environment, and maybe an option for job scheduler-based installations.

A deployments.yaml file for a given site could look like:
```yaml
core_deployments:
  - template: "unified-dev"
    compilers: ["gcc/12.4.0", "oneapi/2024.2.1"]

addon_deployments:
  - template: "some-pkg-fix-dev"
    compilers: ["oneapi/2024.2.1"]
    upstreams: ["unified-dev"]
  - template: "some-pkg-second-fix-dev"
    compilers: ["oneapi/2024.2.1"]
    upstreams: ["unified-dev", "some-pkg-fix-dev"]
```

climbfuji:
[...]

PR #1821, opened 2025-11-15
More Acorn+WCOSS2/NCO updates

AlexanderRichert-NOAA:
## Description

This PR makes the following **general** changes:
 - blacklist compiler-wrapper and intel-oneapi-compilers-classic from modules
 - adds a `uwtools` variant to global-workflow-env; on by default but disable-able for stacks/sites where it's not needed
 - use 'require' to set virtual package providers in common/packages*.yaml so that we don't have to configure/override these separately for various dependencies (and to avoid possible concretization quirks because of "providers" being a soft preference).
 - move show_duplicate_packages output logic outside the show_duplicate_packages() function; also cleaned up some defunct logic left over from the previous iteration of this script

and the following **Acorn/WCOSS2/NCO-specific** changes:
 - overhaul acorn and wcoss2 site configs, including reconciling them so they're basically the same apart from some NCO-specific package configuration
 - overhaul of deploy.py: better logging; better integration with Spack Python API; move deployments.yaml under site config dir; validate env's by ensuring only user-specified packages can be built with GCC; add a "modules_flat/" directory that tests having all modulefiles in a flat structure (no compiler or mpi subdirectories)
 - resurrect 'nco' template and add it to acorn deployments
 - no more intel@2022 (per NCO & EMC agreement)

### Dependencies

none

### Issues addressed

Relates to #1772

### Applications affected

oneapi provider config changes (potentially) affect all applications
global-workflow-env

### Systems affected

show_duplicate_packages.py change (potentially) affects all deployments

## Testing

    - [ ] ...
    - [ ] tested all four env's defined in acorn's deployments.yaml

## Checklist

- [x] All necessary updates to the documentation on readthedocs are included in this PR
- [x] All necessary updates to the spack-stack wiki will be made when this PR is merged


configs/common/packages_oneapi.yaml
```diff
@@ -2,10 +2,12 @@ packages:
     all:
       prefer:
       - '%oneapi'
-      providers:
-        blas: [intel-oneapi-mkl]
-        fftw-api: [intel-oneapi-mkl]
-        lapack: [intel-oneapi-mkl]
+    blas:
```
climbfuji, line 6: Should we make similar changes for every supported compiler in packages_${COMPILER}.yaml?

We want openblas and fftw with GNU and LLVM.

climbfuji (commented):
Looks great, thanks for the cleanup.

configs/common/packages_oneapi.yaml
AlexanderRichert-NOAA: Sounds good to me, I can update those this afternoon

AlexanderRichert-NOAA: I updated common/{packages,packages_clang,packages_gcc}.yaml in addition to the oneapi one. See what you think.

climbfuji: Looks good to me

climbfuji (approved):
CI tests all passed. I checked that the GNU and Intel builds use the correct linalg providers and variants for the packages.

PR #1891, opened 2026-01-27
Add gittools as varant to neptune-python-env/neptune-dev, add rank-run to base-env

### LINKED ISSUE #1727: [INSTALL]: py-pygithub 2.7
### Package name

py-pygithub

### Package version/tag

2.7

### Build options

current

### Installation timeframe

spack-stack-2.*.*

### Other information

py-github@2.7 is out - https://github.com/PyGithub/PyGithub/releases/tag/v2.7.0

This contains updates from @climbfuji to monitor self-hosted runners defined at the organization level.

### LINKED ISSUE #1805: [INSTALL]: request to include `rank_run` in spack-stack
### Package name

rank_run

### Package version/tag

v1.0.0

### Build options

Needs MPI support

### Installation timeframe

Please include it in the next release

### Other information

`rank_run` is a light-weight MPI tool for running commands/scripts in different MPI ranks so that we can execute serial jobs in parallel. For example, we may run 24 wgrib2 commands in 24 ranks in one HPC job to process 24 hours of forecasts at the same time, etc. This tool is needed by rrfs-workflow and can be utilized by other applications as well.

The `rank_run` repository: https://github.com/RRFSx/rank_run

For more information, check https://github.com/RRFSx/rank_run/wiki

climbfuji:
## Description

1. Add `gittools` variant (`gh`, `py-pygithub`) to `neptune-python-env` and enable in `neptune-dev` template. I couldn't enable it in the `unified-dev` template for `oneapi`, because the concretizer ended up in an infinite loop, heating my room overnight but not concretizing the environment.
2. Add `rank-run` to `base-env`.
3. Add external `bash` to all DOD HPCMP systems (required by `gh` --> `go`, don't want to build `bash`!), also find in GitHub workflows (I checked that it is being found and used).
4. Fix bug in `.gitmodules`: unintentional leftover from previous PR that didn't revert the `spack` submodule pointer information.

### Dependencies

- [x] Waiting on https://github.com/JCSDA/spack-packages/pull/36

### Issues addressed

Closes #1805 
Closes #1727

### Applications affected

NEPTUNE

### Systems affected

NRL systems

## Testing

    - [ ] ...
    - [ ] ...

## Checklist

### DIFF
--- a/configs/sites/tier2/jean/packages.yaml
+++ b/configs/sites/tier2/jean/packages.yaml
@@ -21,6 +21,10 @@ packages:
     externals:
     - spec: automake@1.16.1
       prefix: /usr
+  bash:
+    externals:
+    - spec: bash@4.4.20
+      prefix: /usr
   binutils:
     externals:
     - spec: binutils@2.30.125~gold~headers
--- a/.github/workflows/ubuntu-ci-x86_64-gnu.yaml
+++ b/.github/workflows/ubuntu-ci-x86_64-gnu.yaml
@@ -101,6 +101,7 @@ jobs:
             spack external find --scope system wget
             spack external find --scope system texlive
             spack external find --scope system mysql
+            spack external find --scope system bash
 
             # Find compilers
             spack compiler find --scope system
--- a/.github/workflows/ubuntu-ci-x86_64-oneapi-ifx.yaml
+++ b/.github/workflows/ubuntu-ci-x86_64-oneapi-ifx.yaml
@@ -101,6 +101,7 @@ jobs:
             spack external find --scope system wget
             spack external find --scope system texlive
             spack external find --scope system mysql
+            spack external find --scope system bash
 
             # Find compilers
             spack compiler find --scope system
--- a/.github/workflows/ubuntu-ci-x86_64-oneapi.yaml
+++ b/.github/workflows/ubuntu-ci-x86_64-oneapi.yaml
@@ -101,6 +101,7 @@ jobs:
             spack external find --scope system wget
             spack external find --scope system texlive
             spack external find --scope system mysql
+            spack external find --scope system bash
 
             # Find compilers
             spack compiler find --scope system
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,9 +1,7 @@
 [submodule "spack"]
   path = spack
-  #url = https://github.com/jcsda/spack
-  #branch = spack-stack-dev
-  url = https://github.com/climbfuji/spack
-  branch = feature/update_spack-stack-dev_from_develop
+  url = https://github.com/jcsda/spack
+  branch = spack-stack-dev
 [submodule "repos/builtin"]
   path = repos/builtin
   #url = https://github.com/jcsda/spack-packages
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,7 +4,5 @@
   branch = spack-stack-dev
 [submodule "repos/builtin"]
   path = repos/builtin
-  #url = https://github.com/jcsda/spack-packages
-  #branch = spack-stack-dev
-  url = https://github.com/climbfuji/spack-packages
-  branch = feature/pygithub_rank_run
+  url = https://github.com/jcsda/spack-packages
+  branch = spack-stack-dev
--- a/.gitmodules
+++ b/.gitmodules
@@ -6,5 +6,7 @@
   branch = feature/update_spack-stack-dev_from_develop
 [submodule "repos/builtin"]
   path = repos/builtin
-  url = https://github.com/jcsda/spack-packages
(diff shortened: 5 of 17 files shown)


configs/sites/tier2/jean/packages.yaml
climbfuji, line 1: I only added the `bash` external. The other, non-visible changes are because the line endings changed from Windows to Unix.

AlexanderRichert-NOAA:
Which applications use rank-run? Is it needed for UFS Weather Model?

climbfuji:
> Which applications use rank-run? Is it needed for UFS Weather Model?

According to https://github.com/JCSDA/spack-stack/issues/1805, the rrfs-workflow uses it. It's such a trivial tool to install (dependencies, only one executable) that I don't have a problem including it in the base environment.

mathomp4 (approved):

AlexanderRichert-NOAA:
Separately from this PR I think it would pay to discuss base-env, what's in it, and whether we need it (i.e., as opposed to just separately enumerating dependencies for the apps that use it). My main concern is that at least currently, our main installations on WCOSS2 will be based on a trimmed down unified-dev template, but we're still constrained by what software is allowed on WCOSS2, so ideally I'd like the ufs-weather-model-env spec to correspond with just what UWM needs. I can create an issue and/or raise it in our next meeting.

climbfuji:
> Separately from this PR I think it would pay to discuss base-env, what's in it, and whether we need it (i.e., as opposed to just separately enumerating dependencies for the apps that use it). My main concern is that at least currently, our main installations on WCOSS2 will be based on a trimmed down unified-dev template, but we're still constrained by what software is allowed on WCOSS2, so ideally I'd like the ufs-weather-model-env spec to correspond with just what UWM needs. I can create an issue and/or raise it in our next meeting.

I thought about your question last night after I responded. That makes sense. We could create a variant with optional tools or define an env for rrfs-workflow and other workflows that don't have one yet, and then stick it in there. What do you think?

AlexanderRichert-NOAA:
I'm okay with either solution, though I prefer the latter solution of separate metapackages as that feels more logical to me in terms of having metapackages that correspond with applications (though if we need a 'misc-env' catch-all metapackage I'm fine with that too). How does that sound?

climbfuji:
> I'm okay with either solution, though I prefer the latter solution of separate metapackages as that feels more logical to me in terms of having metapackages that correspond with applications (though if we need a 'misc-env' catch-all metapackage I'm fine with that too). How does that sound?

If I were to create `misc-env`, are there any other packages that should be moved from `base-env` to `misc-env`? I assume we then add `misc-env` to `unified-dev`, `skylab-dev`, and `neptune-dev`?

AlexanderRichert-NOAA:
The packages that jump out at me are cloc, plus some of the Python stuff like pytest and setuptools. For the Python ones I could see either moving those to misc-env or adding a 'python-extras' variant. In my ideal world, spec'ing ufs-weather-model-env would provide strictly what's needed to build UFS Weather Model and run RTs (@rickgrubin @RatkoVasic-NOAA @natalie-perlin please chime in if you have another perspective).

climbfuji:
> The packages that jump out at me are cloc, plus some of the Python stuff like pytest and setuptools. For the Python ones I could see either moving those to misc-env or adding a 'python-extras' variant. In my ideal world, spec'ing ufs-weather-model-env would provide strictly what's needed to build UFS Weather Model and run RTs (@rickgrubin @RatkoVasic-NOAA @natalie-perlin please chime in if you have another perspective).

UFS is not unique here, same situation for NEPTUNE. We want a stripped-down environment to support exactly what is needed for  NEPTUNE in operations - hence the `neptune-ops` template and all the variants and additional logic in `neptune-env` etc.

I wonder if a variant `dev` or `devtools`, shared between all our `-env` packages that need them, would be a possibility?

AlexanderRichert-NOAA:
That sounds reasonable. I could live with setting '~devtools' or whatever in packages.yaml for Acorn/WCOSS2 for example.

climbfuji:
> That sounds reasonable. I could live with setting '~devtools' or whatever in packages.yaml for Acorn/WCOSS2 for example.

Ok. In this case, I suggest we move ahead with this PR as is (add rank-run to base-env) and discuss at the next spack-stack meeting the reorganization of the environments - which packages are part of the devtools variants for the different `-envs` that we share.

AlexanderRichert-NOAA:
Sounds good :+1:

AlexanderRichert-NOAA (approved):

climbfuji:
> Sounds good 👍

Thanks, could you also take a look at the submodule PR, please? https://github.com/JCSDA/spack-packages/pull/36

PR #1954, opened 2026-03-12
Containers and ec2 changes for 2.1 release

eap:
This is a batch of changes to the container environment in 2.1. Additionally this includes an update to the aws-ec2 environment which allows both numpy and aws tools to coexist with the older python setuptools.

### Systems affected

CI, containers, ec2.

## Testing

Local builds and testing.

## Checklist

- [ ] All necessary updates to the documentation ([spack-stack wiki](https://github.com/jcsda/spack-stack/wiki)) will be made when this PR is merged [__pending__]

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,19 @@
+# .dockerignore
+
+# Log files. 
+log.*
+*.log
+
+# spack-stack environments and builds.
+envs/
+.env/
+build/
+cache/
+
+# Misc.
+.vscode
+.Trashes
+.github/
+venv/
+.venv/
+
--- a/configs/sites/tier1/container/packages.yaml
+++ b/configs/sites/tier1/container/packages.yaml
@@ -32,26 +32,22 @@ packages:
     - "@1.7.2"
 
   # System packages installed during the container build.
-  grep:
+  autoconf:
     externals:
-    - spec: grep@3.11
-      prefix: /usr
-  findutils:
-    externals:
-    - spec: findutils@4.9.0
+    - spec: autoconf@2.71
       prefix: /usr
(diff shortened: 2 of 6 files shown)


configs/sites/tier2/aws-ubuntu2404/packages.yaml
```diff
@@ -1,66 +1,87 @@
...
+    require:
+    - "@1.34.162"
+  py-boto3:
+    require:
+    - "@1.34.162"
+  py-netcdf4:
+    require:
+    - "@1.7.2"
+
+  # System packages.
```
eap, line 22: The set of packages below have been alphabetized. The trade is an ugly diff here in exchange for much better diffs in the future.

.dockerignore
```diff
@@ -0,0 +1,19 @@
+# .dockerignore
```
eap, line 1: Adding a dockerignore because otherwise docker will load the entire build context, including spack-stack installs, even if the install directories are manually excluded from the `COPY` directive. This results in rapid consumption of disk resources during builds even when resources are ignored.

configs/sites/tier1/container/packages.yaml
```diff
@@ -1,8 +1,36 @@
...
+    - +pnetcdf
+  pixman:
+    variants: +pic
+  py-awscrt:
+    require:
+    - "@0.19.19"
+  py-botocore:
+    require:
+    - "@1.34.162"
+  py-boto3:
```
climbfuji, line 27: You figured it out, great!

climbfuji (approved):
This looks good to me, but I am not that familiar with the docker builds anymore.

PR #1973, opened 2026-04-01
Acorn: Remove $env template dir

AlexanderRichert-NOAA:
## Description

This PR removes the custom template dir under $env from the Acorn configuration; I only want to use this for WCOSS2-related deployments, otherwise regular module file usage becomes a mess for R&D models due to the prereq's. I retroactively fixed the module files for the 2.1.0 release on Acorn so the documentation is accurate.

### Dependencies

none

### Issues addressed

none

### Applications affected

n/a

### Systems affected

Acorn

## Testing

Tested on Acorn (verified correct behavior after removing config entry)

## Checklist

### DIFF
--- a/configs/sites/tier1/acorn/config.yaml
+++ b/configs/sites/tier1/acorn/config.yaml
@@ -2,5 +2,4 @@ config:
   build_jobs: 6
   build_stage: $tempdir/$user/spack-stage
   template_dirs:
-  - $env/site/templates
   - $spack/share/spack/templates


climbfuji:
Interestingly, I am not seeing any effect of the CODEOWNERS file, despite these settings:

image

climbfuji (approved):

climbfuji:
This is the reason:

image

I'll take care of the permissions.

climbfuji:
image

image

I could fix @Hang-Lei-NOAA's permissions (he should have more privileges than write, but we will clean up repository access in two weeks anyway, therefore leaving as is).

For @alysonstahl-noaa, she'll need to accept the invite for the CODEOWNERS file to become effective.

AlexanderRichert-NOAA:
Thanks for checking on that. It's a little disconcerting that it ends up discarding the settings entirely :shrug:

PR #2048, opened 2026-06-30
stack_env.py: 'include:' compiler- and lmod/tcl-specific files instead of convolving yaml's

AlexanderRichert-NOAA:
## Description

This PR proposes to refactor the way we include yaml configs with `spack stack create env`. Currently, the yaml configs get convolved into one packages.yaml for common/ and site/ (and similarly for modules.yaml). This PR modifies spack_env.py so that the files specific to compilers or to lmod vs. tcl module systems are pointed to explicitly in the newly created spack.yaml. For example:
instead of having the following where the common packages_oneapi.yaml and site-specific packages_oneapi-2024.2.1.yaml are convolved into common/packages.yaml and site/packages.yaml, respectively:
```yaml
  include:
  - site
  - common
```
`spack stack create env` would yield:
```yaml
  include:
  - site/packages_oneapi-2024.2.1.yaml
  - site
  - common/packages_oneapi.yaml
  - common/modules_lmod.yaml
  - common
```

Note that putting the file before their parent directories is done to retain the current precedence, where the compiler/module sys-specific customizations override the general ones.

Advantages of this approach:
- It makes it much easier to propagate changes back to $SPACK_STACK_DIR/config because the config files are no longer modified.
- `spack config blame` makes it very easy to know where a problematic setting came from without having to do a second layer of investigation into which exact file it was pulled from, what the precedence was, etc.
- It (arguably) keeps the configurations in a clearer structure/order, especially when it comes to comments, again because the individual files don't get modified. If a user is unclear on what the net result is of, say, having an entry for a given package in each of 3 or 4 config files, the net result is easily determined using `spack config get` or `spack config blame`

### Dependencies

none

### Issues addressed

none

### Applications affected

n/a

### Systems affected

all

## Testing

    - [x] Tested config validity on local machine (i.e., ran `spack config get/blame`)
    - [x] Tested on Acorn (concretization & validation; `spack config blame` to verify config file usage)

## Checklist

### DIFF
--- a/spack-ext/lib/jcsda-emc/spack-stack/stack/stack_env.py
+++ b/spack-ext/lib/jcsda-emc/spack-stack/stack/stack_env.py
@@ -144,6 +144,11 @@ class StackEnv(object):
         packages_compiler_yaml_path = os.path.join(common_path, packages_compiler_yaml)
         if os.path.exists(packages_compiler_yaml_path):
             self.includes.append(os.path.join("common", packages_compiler_yaml))
+        else:
+            if self.treatwarningsaserrors:
+                raise Exception(f"\n{packages_compiler_yaml_path} not found, please check if this is correct\n")
+            else:
+                logging.warning(f"\nWARNING: {packages_compiler_yaml_path} not found, please check if this is correct\n")
 
         lmod_or_tcl = self.get_lmod_or_tcl(self.site_configs_dir())
         modules_yaml_modulesys = f"modules_{lmod_or_tcl}.yaml"
@@ -177,6 +182,11 @@ class StackEnv(object):
         packages_compiler_yaml_path = os.path.join(site_path, packages_compiler_yaml)
         if os.path.exists(packages_compiler_yaml_path):
             self.includes.append(os.path.join(site_name, packages_compiler_yaml))
+        else:
+            if self.treatwarningsaserrors:
+                raise Exception(f"\n{packages_compiler_yaml_path} not found, please check if this is correct\n")
+            else:
+                logging.warning(f"\nWARNING: {packages_compiler_yaml_path} not found, please check if this is correct\n")
 
         lmod_or_tcl = self.get_lmod_or_tcl(self.site_configs_dir())
         modules_yaml_modulesys = f"modules_{lmod_or_tcl}.yaml"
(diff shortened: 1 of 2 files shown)


spack-ext/lib/jcsda-emc/spack-stack/stack/stack_env.py
```diff
@@ -329,7 +292,7 @@ def write(self):
                     logging.warning(
                         "WARNING: Upstream path '%s' does not appear to exist!" % upstream_path
                     )
-                re_pattern = ".+/(?P<spack_stack_ver>spack-stack-[^/]+)/envs/(?P<env_name>[^/]+)"
+                re_pattern = ".+/(?P<spack_stack_ver>spack-stack-[^/]+)/.+/(?P<env_name>[^/]+)/install?$"
```
climbfuji, line 295: Out of curiosity, why is this needed?

climbfuji (approved):

spack-ext/lib/jcsda-emc/spack-stack/stack/stack_env.py
AlexanderRichert-NOAA: I added this to help get the env name a little more robustly in a case where my spack-stack dir name didn't follow the 'spack-stack-x.y.z' pattern.

mathomp4 (approved):
Trust in Alex!

spack-ext/lib/jcsda-emc/spack-stack/stack/stack_env.py
climbfuji: Ok ... The `.+` will match `envs` and other names (e.g. `install`). I am not seeing a change to the `spack-stack-x.y.z` pattern, but that's fine, I was just curious.

climbfuji (approved):
This is a great improvement, thanks for working on this.

spack-ext/lib/jcsda-emc/spack-stack/stack/stack_env.py
AlexanderRichert-NOAA: Yeah good point, I wonder if I merged only part of my modifications. I think I'll leave this for now since it still works as best I can tell, but I'll refine in a separate PR for other cases if/when needed.

PR #2068, opened 2026-07-20
Update the global workflow environment

### LINKED ISSUE #1994: Update global_workflow_env with the most recent required libraries
## Is your feature request related to a problem?

The global workflow environment is out of date and needs to be updated to include the required libraries for current development work.

## Describe the solution you'd like

The libraries required should be updated in https://github.com/JCSDA/spack-stack/blob/develop/repos/spack_stack/spack_repo/spack_stack/packages/global_workflow_env/package.py to reflect the current libraries used by the package. Alternatively (perhaps better), the environment should be updated to be modular and inherit its needs from the associated submodules (UFS-WM, GSI, GDASApp, UPP, verif-global, UFS_Utils). These packages' environments should also be updated.

## Additional context
Originally discussed in #1993

AlexanderHrabski-NOAA:
## Description

This PR updates the `global-workflow-env` with the latest dependencies of the global workflow. The python variant is removed because it now must always be on, and some additional variants were added to support user installs for specific applications per the comments in #1994. Also removes the `ip2` dependence from `ufs-utils-envs`, which was leading to a concretization error. `ip2` was depreciated, and no longer exists in the `spack-packages` upstream. `ip` provides the lost functionality.

### Dependencies

#2063 is required to build the environment for Ubuntu WSL.

### Issues addressed

Resolves #1994 

### Applications affected

Affects the global workflow. Also modifies the unified environment, which includes the global workflow environment.

### Systems affected

All, via changing an environment that is part of the unified environment.

## Testing

    - [ ] GitHub actions CI tests pass

    - [x] Concretized and installed the `unified-env` on Ubuntu WSL with GNU compilers
    - [x] Concretized and installed `global-workflow-env` on Ubuntu WSL with GNU compilers
    - [x] Concretized and installed the `unified-env` on Ursa with Intel OneAPI compilers
    - [x] Concretized and installed `global-workflow-env` on Ursa with Intel OneAPI compilers

Keeping this a draft while I test. Please let me know if there are other systems you'd like me to test this on.

## Checklist


configs/templates/unified-dev/spack.yaml
```diff
@@ -11,7 +11,8 @@ spack:
   - dev-utils-env
   - ewok-env
   - geos-gcm-env          ^esmf@=9.0.0b11
-  - global-workflow-env   ^esmf@=8.6.1 ^crtm@=3.1.3
+  - global-workflow-env   ^esmf@=8.8.0 ^crtm@=v2.4.1-jedi.2
+
   # Is GMAO ready to move to crtm@3.1.3?
   - gmao-swell-env                     ^crtm@=v2.4.1-jedi.2
   - gsi-env                            ^crtm@=3.1.3
```
DavidHuber-NOAA, line 18: Now that the global workflow environment includes the GSI environment, should this be removed from the unified environment?

configs/common/modules_lmod.yaml
```diff
@@ -132,6 +132,16 @@ modules:
...
+      # The unified environment concretizes these two envs twice: once against
+      # crtm@3.1.3 (their own root specs) and once against crtm@v2.4.1-jedi.2,
+      # as pulled in by global-workflow-env. Suffix the latter so the module
+      # files don't clash.
+      gsi-env:
+        suffixes:
+          ^crtm@v2.4.1-jedi.2: 'crtm-2.4.1-jedi.2'
+      ufs-weather-model-env:
+        suffixes:
+          ^crtm@v2.4.1-jedi.2: 'crtm-2.4.1-jedi.2'
```
DavidHuber-NOAA: I should have caught this earlier. If we need to keep CRTM 2 for the GSI, we would need the version 2.4.0.1, not 2.4.1-jedi. I'm not sure how feasible this option is, though, as that would make for 3 CRTM packages, and 2.4.0.1 has not been in spack-stack since 1.9.x, so I'm not sure how compatible it is.

All of that to say, I think it would be best if we can just move to CRTM 3.x, and my suggestion would be to remove crtm-2.4.1-jedi from the GSI and global workflow environments entirely.

But I should defer this line to DA personnel:

@russtreadon-noaa do you think it would be acceptable to move the GSI and GDASapp to CRTM 3 for the spack-stack 2.x upgrade or do you need the capability to run with CRTM 2?

RussTreadon-NOAA: @DavidHuber-NOAA :  We should move GSI and GDASApp to CRTM v3.   @ADCollard was looking at CRTM v3 in the GSI.  I'm not aware of anyone looking at CRTM v3 in GDASApp

DavidHuber-NOAA: Thanks @RussTreadon-NOAA.

In that case, I'll stick to my original suggestion that we remove references to CRTM 2.4.x from the GSI and GW environments.

climbfuji:
@AlexanderHrabski-NOAA Can you take a look at the reviewer comments and pull in develop, please?

configs/templates/unified-dev/spack.yaml
AlexanderHrabski-NOAA: If the `gsi-env` is included in the unified environment only because the global workflow needs it, then I'd say yes. I'd like to defer to @climbfuji and @AlexanderRichert-NOAA on this though!

configs/common/modules_lmod.yaml
AlexanderHrabski-NOAA: The `gsi-env` included explicitly in the unified environment already depends on `crtm@3.1.3`. Regardless of whether or not the `gsi-env` is retained in the unified environment explicitly, adding a `^crtm@3.1.3` to the global workflow environment would cover the global workflow + GSI `crtm` dependence. JEDI omits `crtm` dependence to force the user to provide it explicitly. In keeping with that pattern, I think `global-workflow-env+gdas` probably shouldn't specify the `crtm` version needed for JEDI. For the vast majority of users, they will get the correct `crtm` version from the unified environment anyway. Otherwise, the version needed for JEDI will be missing (unless it's `crtm@3.1.3`) and the user will have to decide, which is the intention.

AlexanderHrabski-NOAA:
I rebased on develop and reverted the changes related to duplicate packages because the duplicates were due to having additional `crtm` versions. I left `gsi-env` in the unified environment -- after thinking about it further I'm in favor of leaving it in. However, I'd be happy to remove it at your request @climbfuji and @AlexanderRichert-NOAA.

AlexanderRichert-NOAA:
I personally don't have a preference. The idea of the metamodules is to make sure all the various apps are represented in the unified env, so as long as all the GSI dependencies end up there my two cents is it's okay either way.

climbfuji:
I defer to @RussTreadon-NOAA and @DavidHuber-NOAA

RussTreadon-NOAA:
As long as GSI applications are able to build and run correctly, I am personally fine with whatever the library team decides.

configs/common/packages_oneapi.yaml
```diff
@@ -78,3 +78,7 @@ packages:
     rust:
       require:
       - '%c,cxx=gcc'
+    # glib@2.88 with the meson@1.11 build system fails to build for intel-oneapi-compilers@2025.3.1
```
climbfuji: Interesting. I didn't run into issues building glib with Intel oneAPI, using versions between 2025.0.3 and 2026.1.0.

AlexanderHrabski-NOAA: I'd be happy to try to recreate the error. It's probably worth re-verifying the `unified-env` on Ursa after these changes and the rebase anyway!

configs/templates/unified-dev/spack.yaml
DavidHuber-NOAA: The GSI is also used by HAFS and RRFS, so perhaps it makes sense to keep it here with a tagged version of the CRTM. The global workflow will likely be the first of these apps to retire the GSI. When that happens, it may be worth revisiting.

configs/common/modules_lmod.yaml
DavidHuber-NOAA: Thanks @AlexanderHrabski-NOAA. The CRTM versions look good to me now.

DavidHuber-NOAA:
> I rebased on develop and reverted the changes related to duplicate packages because the duplicates were due to having additional `crtm` versions. I left `gsi-env` in the unified environment -- after thinking about it further I'm in favor of leaving it in. However, I'd be happy to remove it at your request @climbfuji and @AlexanderRichert-NOAA.

I'm in favor of this remaining in the unified environment. Both RRFS and HAFS use the GSI as well, so it makes sense to me to keep it in the unified environment. We can revisit this as these applications retire the GSI in favor of JEDI.

DavidHuber-NOAA (approved):

configs/common/packages_oneapi.yaml
AlexanderHrabski-NOAA: While testing, on Ursa, the `unified-env` now contains:
```
^glib@2.88.1+introspection~libmount~strip build_system=meson buildtype=release default_library:=shared tracing:=none platform=linux os=rocky9 target=zen4 %c,cxx=oneapi@2025.3.1
```
When I comment out this requirement of glib. During installation, I get the error:
```
ERROR: In override_options for atomic-c-23: None of values ['c23'] are supported by the C compiler. Possible values for option "c_std" are ['none', 'c89', 'c99', 'c11', 'c17', 'c18', 'c2x', 'gnu89', 'gnu99', 'gnu11', 'gnu17', 'gnu18', 'gnu2x']
```
Let me know if you'd like me to try to produce an MRE for this!

climbfuji: I built glib@2.88.1 successfully with Intel oneAPI 2026.0.0 and meson@1.11 on my laptop.

AlexanderHrabski-NOAA: I also was able to build `glib@2.88.1` with `oneAPI@2026.1.1` and `meson@1.11.1` on my Ubuntu 22 WSL. I'll try to reproduce the issue on my laptop.

AlexanderRichert-NOAA (approved):

configs/common/packages_oneapi.yaml
AlexanderHrabski-NOAA: Now a new issue #2093, and I've removed this change from the PR. The oneAPI CI test should serve as additional confirmation that building `glib` with `gcc` isn't needed here.

climbfuji (approved):
Thanks for removing the glib change

climbfuji:
I'll update this PR from develop after merging #2088 to kick off the final round of CI tests.

PR #2084, opened 2026-08-13
Fix meta-modules preferred compiler bug

eap:
The motivation for this PR comes from the the function `custom_sort_key` which was broken, leading to some bugs when generating meta-modules with compilers that were dependent on the preferred compiler.

When sorting, `Spec.__contains__`, checks the dependency tree of a spec, since llvm was built with gcc it got the same sort order as gcc in the list.

## Other fixes

The `for compiler in compilers` loop had a whole bunch of meta-module logic that only applied to the preferred compiler, this was gated with a "continue" clause. I just dropped that clause and moved everything into the function context. That saved some variable scope bookkeeping and (along with the sort fixes) guarantees that it runs every time.

### DIFF
--- a/spack-ext/lib/jcsda-emc/spack-stack/stack/meta_modules.py
+++ b/spack-ext/lib/jcsda-emc/spack-stack/stack/meta_modules.py
@@ -289,12 +289,11 @@ def setup_meta_modules():
             if any(core_compiler in x for x in compiler_list):
                 raise Exception("Not supported: spack-stack compilers in core compilers")
 
-    # Determine the preferred compiler. Exactly one of the compilers used for the
-    # stack must match it.
+    # Determine the preferred compiler.
     preferred_compiler = get_preferred_compiler(spack.config)
     logging.info("  ... preferred compiler: {}".format(preferred_compiler))
 
-    # Check that exactly one compiler matches the preferred compiler.
+    # Check that exactly one compiler spec matches the preferred compiler.
     preferred_compiler_specs = [c for c in compilers if c.name == preferred_compiler]
     if len(preferred_compiler_specs) != 1:
         raise Exception(f"Expected 1 preferred spec for '{preferred_compiler}'. Found "
--- a/spack-ext/lib/jcsda-emc/spack-stack/stack/meta_modules.py
+++ b/spack-ext/lib/jcsda-emc/spack-stack/stack/meta_modules.py
@@ -289,21 +289,22 @@ def setup_meta_modules():
             if any(core_compiler in x for x in compiler_list):
                 raise Exception("Not supported: spack-stack compilers in core compilers")
 
-    # Determine the preferred compiler and sort the list of compilers so that
-    # the preferred compiler comes last. This is so that all other compilers
-    # populate the MODULEPATHS_SAVE list before the preferred compiler
-    # takes it and adds it to the stack-COMPILER metamodule. Likewise, we need
-    # to save the list of compiler substitutions from the preferred compiler
-    # so that we have access to it when we build the MPI meta module.
+    # Determine the preferred compiler. Exactly one of the compilers used for the
+    # stack must match it.
     preferred_compiler = get_preferred_compiler(spack.config)
     logging.info("  ... preferred compiler: {}".format(preferred_compiler))
 
-    # Sort the list using a custom key
-    def custom_sort_key(entry):
-        # Return a tuple where the first element is 1 if the entry contains the word, else 0
-        # The second element is the entry itself for natural sorting within groups
-        return (1 if preferred_compiler in entry else 0, entry)
-    compilers = sorted(compilers, key=custom_sort_key)
+    preferred_compiler_specs = [x for x in compilers if x.name == preferred_compiler]
+    if not preferred_compiler_specs:
+        raise Exception(f"No compiler matching the preferred compiler '{preferred_compiler}' in {compiler_list}")
+    if len(preferred_compiler_specs) > 1:
+        raise Exception(f"Multiple compilers matching the preferred compiler "
+                        f"'{preferred_compiler}': {preferred_compiler_specs}")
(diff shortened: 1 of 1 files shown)


spack-ext/lib/jcsda-emc/spack-stack/stack/meta_modules.py
```diff
@@ -374,105 +371,102 @@ def custom_sort_key(entry):
                 module_choice = module_choice
             )
 
-        # The remainder of the loop is only needed for the preferred compiler
-        if not compiler.name in preferred_compiler:
-            continue
```
eap, line 379: Ditched this clause and dedented everything from here to `number_of_meta_modules_written += 1`

climbfuji:
Thanks for this cleanup. I'll need to give this a thorough test on a few systems.

spack-ext/lib/jcsda-emc/spack-stack/stack/meta_modules.py
```diff
@@ -502,6 +494,7 @@ def custom_sort_key(entry):
                 module_choice = module_choice
             )
 
+        # Collect the modulepaths for all compilers (preferred compiler last)
```
eap, line 499: Underneath the MPI provider loop we have a second compiler loop that operated much like the first compiler loop; it assuming that the last compiler would be the preferred compiler. Again here I dedent the secondary "preferred" block and remove the loop's "continue" statement causing it to execute after the loop using the previously saved `preferred_compiler_spec`

By my reading it is still important that the preferred compiler comes last since that influences path precedence.

spack-ext/lib/jcsda-emc/spack-stack/stack/meta_modules.py
```diff
@@ -289,21 +289,20 @@ def setup_meta_modules():
...
     preferred_compiler = get_preferred_compiler(spack.config)
     logging.info("  ... preferred compiler: {}".format(preferred_compiler))
 
-    # Sort the list using a custom key
-    def custom_sort_key(entry):
-        # Return a tuple where the first element is 1 if the entry contains the word, else 0
-        # The second element is the entry itself for natural sorting within groups
-        return (1 if preferred_compiler in entry else 0, entry)
-    compilers = sorted(compilers, key=custom_sort_key)
+    # Check that exactly one compiler spec matches the preferred compiler.
```
eap, line 296: This code is probably good but could break some sites. It should be possible (today) to have two compilers like "gcc@11" and "gcc@13" and have them both sort "last" and probably that would silently be successful but one of them would be put in the "real" last place according to whatver sort order was first... this feels like an uncaught error to me, but maybe not?

This logic will probably break any site with two coexisting versions of the same compiler in the same environment. Hopefully that isn't a case we need to support?

climbfuji:
@eap I am finally testing this PR - stay tuned

eap (commented):
review notes

climbfuji:
I think this PR is good the way it's now. I tested it on 2/3 platforms, just finishing the third.

climbfuji (approved):
Works on the three sites I tested (HPCs / devbox, lmod / tcl, intel / gcc / llvm)

spack-ext/lib/jcsda-emc/spack-stack/stack/meta_modules.py
climbfuji: We don't. one compiler version per vendor is sufficient.

spack-ext/lib/jcsda-emc/spack-stack/stack/meta_modules.py
```diff
@@ -289,21 +289,20 @@ def setup_meta_modules():
...
+    preferred_compiler_specs = [c for c in compilers if c.name == preferred_compiler]
+    if len(preferred_compiler_specs) != 1:
+        raise Exception(f"Expected 1 preferred spec for '{preferred_compiler}'. Found "
+                        f"{len(preferred_compiler_specs)} matching specs in:\n{compilers}")
+    preferred_compiler_spec = preferred_compiler_specs[0]
+
+    # Sort compilers so that the preferred compiler comes last; the MODULEPATHS
+    # are prepended giving the preferred compiler the highest precedence.
```
climbfuji, line 304:
```suggestion
    # are prepended giving the preferred compiler the highest precedence.
    # Python sorts tuples lexicographically, and booleans are ordered as integers:
    # False == 0, True == 1
```

climbfuji:
@AlexanderHrabski-NOAA @AlexanderRichert-NOAA I'd like to merge this soon, if possible. Any desire to review and/or test? Thanks!

rickgrubin-noaa (approved):
Successfully tested on two EPIC hosts.

climbfuji:
> Successfully tested on two EPIC hosts.

Thanks, Rick for testing! Appreciate it.

AlexanderHrabski-NOAA:
@climbfuji I'm not too familiar with the meta-module piece of spack-stack (though I am learning), and I'll need to shift my attention elsewhere for a couple days. I can try to find time on Friday to review this if it's still open!

climbfuji:
> @climbfuji I'm not too familiar with the meta-module piece of spack-stack (though I am learning), and I'll need to shift my attention elsewhere for a couple days. I can try to find time on Friday to review this if it's still open!

We can wait until Friday, no problem. We have plenty of other PRs in the pipeline :-)

AlexanderHrabski-NOAA:
I think this looks good. I tested the compiler order in an env constructed to match the case this addresses -- `gcc` is preferred, but the env also contains `llvm%gcc`. The old logic did not prefer `gcc`, but the new logic does as expected.

AlexanderHrabski-NOAA (approved):

climbfuji:
> I think this looks good. I tested the compiler order in an env constructed to match the case this addresses -- `gcc` is preferred, but the env also contains `llvm%gcc`. The old logic did not prefer `gcc`, but the new logic does as expected.

Thanks for your review and testing.

PR #2094, opened 2026-08-18
Add support for basic NEPTUNE env with nvhpc@26.5

climbfuji:
## Description

This PR adds support for building a basic NEPTUNE environment with `nvhpc@26.5`, similar to the `llvm@22` NEPTUNE environment. A large number of packages need to be built with GCC instead of NVHPC.

Because NVIDIA is updating their compiler suite from the legacy PGI backend to the modern LLVM IR in the near future, it doesn't make sense to try to fix the many build errors with NVHPC. Also, NVIDIA's support to fix bugs in the current compilers is limited - they defer users to the upcoming LLVM-based compilers.

### Dependencies

- [x] https://github.com/JCSDA/spack-packages/pull/79

### Issues addressed

NRL-internal issue

### Applications affected

None (this is an experimental setup)

### Systems affected

None

## Testing

    - [x] GitHub actions CI tests pass (CI tests not affected by this change)
    - [ ] ...
    - [ ] ...

## Checklist

- [x] These changes have been tested on the affected systems and applications. **to the extent possible**


configs/common/packages_nvhpc.yaml
```diff
@@ -0,0 +1,99 @@
...
+    prefer:
+    - nvhpc
+  # Virtual provider settings
+  blas:
+    require:
+    - nvhpc
+  fftw-api:
+    require:
+    - fftw
+    #- nvpl-fft
```
climbfuji, line 19: Future improvement ... could do the same for blas/lapack instead of taking the built-in one that ships with the compiler (requires performance testing).

AlexanderRichert-NOAA (approved):

PR #2099, opened 2026-08-21
Add compiler install scripts in util/compilers

climbfuji:
## Description

Add compiler install scripts in util/compilers. These can be - and are - hardcoded to specific versions; they are a pragmatic solution to a repetitive task. We can add install scripts for other compilers or newer versions of existing compilers and remove old ones as needed.

### Dependencies

None

### Issues addressed

None

### Applications affected

None

### Systems affected

None

## Testing

n/a

## Checklist

- [x] This PR addresses one issue/problem/enhancement or has a very good reason for not doing so.~~
- [ ] ~~These changes have been tested on the affected systems and applications.~~


util/compilers/intel_oneapi-2026.0.0_install.sh
```diff
@@ -0,0 +1,99 @@
...
+  exit 1
+fi
+if [ ! -d "${ONEAPI_INSTALL_DIR}/mkl/${MKL_VERSION}" ]; then
+  echo "Error, directory ${ONEAPI_INSTALL_DIR}/mkl/${MKL_VERSION} does not exist"
+  exit 1
+fi
+
+# Create modulefiles - special modules for Cray, Intel auto-generated modules elsewhere
+HOSTNAME=$(hostname)
+if [[ "${HOSTNAME}" == *"blueback"* || "${HOSTNAME}" == *"narwhal"* ]]; then
```
climbfuji, line 44: If your system is a cray, then add the hostname here.

rickgrubin-noaa (approved):

util/compilers/intel_oneapi-2026.0.0_install.sh
```diff
@@ -0,0 +1,99 @@
...
+setenv {MKLROOT} {${ONEAPI_INSTALL_DIR}/mkl/${MKL_VERSION}}
+prepend-path {PATH} {${ONEAPI_INSTALL_DIR}/mkl/${MKL_VERSION}/bin}
+prepend-path {PKG_CONFIG_PATH} {${ONEAPI_INSTALL_DIR}/mkl/${MKL_VERSION}/lib/pkgconfig}
+
+proc ModulesHelp { } {
+    global INTEL_LEVEL
+    global INTEL_CURPATH
+}
+
+# This module was produced with dom-gen 0.0.1
```
eap: what is this?

util/compilers/intel_oneapi-2026.0.0_install.sh
```diff
@@ -0,0 +1,99 @@
+#!/usr/bin/env bash
+
+# Utility script to install Intel oneAPI 2026.0.0, fix non-ascii characters
+# in the Intel auto-generated modulefiles (bug filed with Intel to correct
+# upstream) on non-Cray systems, or to create a single module on Cray systems.
```
eap, line 5: I haven't used the "intel-oneapi-toolkit" installer before; what is included with this installer? Just oneAPI c/c++/fortran/mkl or does it include other things like oneapi-mpi and tbb ? In any case documenting the targets up here would be helpful just so its clear what is included by this script.

util/compilers/intel_oneapi-2026.0.0_install.sh
```diff
@@ -0,0 +1,99 @@
...
+if [ ! -d "${ONEAPI_INSTALL_DIR}/compiler/${COMPILER_VERSION}" ]; then
+  echo "Error, directory ${ONEAPI_INSTALL_DIR}/compiler/${COMPILER_VERSION} does not exist"
+  exit 1
+fi
+if [ ! -d "${ONEAPI_INSTALL_DIR}/mkl/${MKL_VERSION}" ]; then
+  echo "Error, directory ${ONEAPI_INSTALL_DIR}/mkl/${MKL_VERSION} does not exist"
+  exit 1
+fi
+
+# Create modulefiles - special modules for Cray, Intel auto-generated modules elsewhere
```
eap, line 42: Intel bundles its own modules and module gen system; at least some of the installers do. Why not use that (or alternatively; is it the case that the intel-oneapi-toolkit installer doesn't include the modules?)

eap (commented):
Some questions.

util/compilers/intel_oneapi-2026.0.0_install.sh
rickgrubin-noaa: > I haven't used the "intel-oneapi-toolkit" installer before; what is included with this installer? Just oneAPI c/c++/fortran/mkl or does it include other things like oneapi-mpi and tbb ? In any case documenting the targets up here would be helpful just so its clear what is included by this script.

[Intel oneAPI Toolkit
](https://www.intel.com/content/www/us/en/developer/tools/oneapi/oneapi-toolkit-download.html?packages=oneapi-toolkit&oneapi-toolkit-os=linux&oneapi-lin=offline)

Compilers, MPI, MKL, TBB.

util/compilers/intel_oneapi-2026.0.0_install.sh
climbfuji: You cannot use the Intel-generated modules on a Cray. The PrgEnv-intel Cray environment requires exactly this type of module. I created it (hence dom-gen) from the template that Cray uses for their Intel oneAPI installations.

util/compilers/intel_oneapi-2026.0.0_install.sh
climbfuji: See below. I just needed to add something here ...

util/compilers/intel_oneapi-2026.0.0_install.sh
climbfuji: Additional, mandatory dependencies nowadays are umf and a few others.

mathomp4: It includes a lot. From my look at 2026.1 I see:

```
ID                                  Version       Installed Name
============================================================================================================
intel.oneapi.lin.dpcpp_dbg          2026.1.0+70   true      Intel® Distribution for GDB*
intel.oneapi.lin.dpl                2022.13.0+107 true      Intel® oneAPI DPC++ Library
intel.oneapi.lin.tbb.devel          2023.1.0+151  true      Intel® oneAPI Threading Building Blocks
intel.oneapi.lin.ccl.devel          2022.1.0+142  false     Intel® oneAPI Collective Communications Library
intel.oneapi.lin.dpcpp-cpp-compiler 2026.1.0+235  true      Intel® oneAPI DPC++/C++ Compiler
 └─intel.oneapi.lin.dpl             2022.13.0+107 true       └─Intel® oneAPI DPC++ Library
intel.oneapi.lin.ifort-compiler     2026.1.0+235  true      Intel® Fortran Compiler
intel.oneapi.lin.ipp.devel          2026.0.1+51   true      Intel® Integrated Performance Primitives
intel.oneapi.lin.ippcp.devel        2026.0.1+9    false     Intel® Cryptography Primitives Library
intel.oneapi.lin.mkl.devel          2026.1.0+236  true      Intel® oneAPI Math Kernel Library
intel.oneapi.lin.vtune              2026.2.0+178  false     Intel® VTune(TM) Profiler
intel.oneapi.lin.dnnl               2026.0.1+55   false     Intel® oneAPI Deep Neural Network Library
 └─intel.oneapi.lin.tbb.devel       2023.1.0+151  true       └─Intel® oneAPI Threading Building Blocks
intel.oneapi.lin.mpi.devel          2021.18.0+745 true      Intel® MPI Library
```

Normally when I install locally, I do:

- Intel oneAPI DPC++/C++ Compiler
- Intel Fortran Compiler
- Intel Distribution for GDB
- Intel oneAPI Math Kernel Library
- Intel Integrated Performance Primitives
- Intel MPI Library
- Intel oneAPI DPC++ Library
- Intel oneAPI Threading Building Blocks

Which means I exclude:

— Intel oneAPI Collective Communications Library
— Intel Cryptography Primitives Library
— Intel VTune Profiler
— Intel oneAPI Deep Neural Network Library

Some I suppose VTune might be nice to have but it is also pretty bulky. No one (yet) has asked for the other ones so I just ignore them.

climbfuji: Honestly, I don't really care about these. For one, if we install the compiler one time with this install script instead of every time a new environment is built with Spack that builds its own compilers, we are still saving a lot of disk space. Second, if someone comes around who is using spack-stack and then wants to use those dependencies for their work, I have to go back in. As I said, these are pragmatic scripts that are supposed to cover our needs with the minimum effort.

climbfuji: @eap These scripts are really only for us and for convenience. I would expect us to go and look at the Intel oneAPI toolkit website/documentation to see what's included - this changes with releases (they went through several iterations over the last years; some releases had separate base + hpc toolkits, now it is only one).

climbfuji: Maybe adding a link to the official website/documentation will do ... but running a google search is just as fast.

util/compilers/intel_oneapi-2026.0.0_install.sh
climbfuji: Let's rename it to "was produced with the spack-stack Intel oneAPI installer wrapper"

util/compilers/intel_oneapi-2026.0.0_install.sh
eap: > Maybe adding a link to the official website/documentation will do

That works for me; I just wasn't familiar with that; I've defaulted to using their apt install sources.

util/compilers/intel_oneapi-2026.0.0_install.sh
eap: got it.

util/compilers/intel_oneapi-2026.0.0_install.sh
```diff
@@ -0,0 +1,99 @@
+#!/usr/bin/env bash
+
+# Utility script to install Intel oneAPI 2026.0.0, fix non-ascii characters
+# in the Intel auto-generated modulefiles (bug filed with Intel to correct
+# upstream) on non-Cray systems, or to create a single module on Cray systems.
+
+# Missing: patching of libirc.so and libimf.so to add the missing symbolic
+# links to libc.so.6 and libm.so.6.
```
rickgrubin-noaa: For those interested: 

Depending on the location of `libm.so .6` and `libc.so.6`, should be:

`patchelf --add-needed /path/to/libm.so.6 ${ONEAPI_INSTALL_DIR}/compiler/${COMPILER_VERSION}/lib/libimf.so`

and

`patchelf --add-needed /path/to/libc.so.6 ${ONEAPI_INSTALL_DIR}/compiler/${COMPILER_VERSION}/lib/libirc.so`

and you'll need to install `patchelf` first.

util/compilers/intel_oneapi-2026.0.0_install.sh
```diff
@@ -0,0 +1,99 @@
...
+}
+
+# This module was produced with dom-gen 0.0.1
+
+module-whatis   "Intel oneAPI compiler"
+EOF
+
+else
+  cd ${ONEAPI_INSTALL_DIR}
+  ./modulefiles-setup.sh --output-dir=${ONEAPI_INSTALL_DIR}/modulefiles --ignore-latest 2>&1 | tee log.modulefiles
```
rickgrubin-noaa: I'll also add: I have successfully stacked multiple versions of the oneAPI toolkit into the same install directory.

If an existing `modulefiles` directory exists (e.g. you installed `2026.0.0` and are now installing `2026.1.0`) into the same root install directory, `modulefiles-setup.sh` will ask to overwrite the `modulefiles` directory. If running interactively, you can answer `'y'` or add the `--force` argument to the invocation of `modulefiles-setup.sh.` Any overwritten modulef iles are re-created.

climbfuji: That's correct. If `--force` works even if the `modulefiles` directory does not exist, then I can simply add it here and the user is free to install in an existing oneapi directory or keep them separate. I prefer the latter, because it's easier to remove an old version (`rm -rf`) than going through the (un)installer. But at least the user has the choice.

util/compilers/intel_oneapi-2026.0.0_install.sh
climbfuji: I am thinking to drop the comment for the next version (copy of this script), because we have the `+fix-rt-linkage` variant in Spack that covers that for our Spack environments. This way, if a (weird?) user depends on the old way these libraries were created - this is what Intel always uses as argument to not fix their libraries - they won't run into surprises.

util/compilers/intel_oneapi-2026.0.0_install.sh
climbfuji: @eap 7fac88e44cc5aecacb16ef56bb450fb8839e95cc

util/compilers/intel_oneapi-2026.0.0_install.sh
climbfuji: @eap e00762c0ac729684ee7c8f6abf881600c774ac22

util/compilers/intel_oneapi-2026.0.0_install.sh
climbfuji: e00762c0ac729684ee7c8f6abf881600c774ac22

util/compilers/intel_oneapi-2026.0.0_install.sh
climbfuji: e00762c0ac729684ee7c8f6abf881600c774ac22

util/compilers/intel_oneapi-2026.0.0_install.sh
rickgrubin-noaa: > I am thinking to drop the comment for the next version (copy of this script), because we have the `+fix-rt-linkage` variant in Spack that covers that for our Spack environments. This way, if a (weird?) user depends on the old way these libraries were created - this is what Intel always uses as argument to not fix their libraries - they won't run into surprises.

Agree -- in this context the spack variant is a better approach / solution.

eap (approved):
Thank you Dom.

